PR #35, opened 2025-04-02
WIP: lane debugging support

palves:
New lane debugging support from scratch, on top of "amd-staging with existing lane support stripped out" branch.

The users/palves/amd-staging-without-lane-support branch is amd-staging with rocgdb's current lane debugging support stripped out.

The users/palves/lane-debugging branch then reapplies all the lane debugging patches again, up until a diff against amd-staging is (almost) empty.  Then, it adds patches on top adjusting to the user interface changes that we will want to submit upstream.

Eventually, the new patches will be squashed into the older patches so that we don't have the intermediate behavior in between.  It should then be easier to rebase it all on upstream master.  Meanwhile, having this based on amd-staging instead helps with the prototyping and discussions because upstream does not have support the needed [DWARF extensions](https://llvm.org/docs/AMDGPUDwarfExtensionsForHeterogeneousDebugging.html) and other important bits, required for accessing variables, unwinding, etc.

### DIFF
--- a/gdb/ada-tasks.c
+++ b/gdb/ada-tasks.c
@@ -1583,8 +1583,8 @@ task_apply_all_command (const char *cmd, int from_tty)
 
   for (const auto &info : thr_list_cpy)
     if (switch_to_thread_if_alive (info.second.get ()))
-      thread_try_catch_cmd (info.second.get (), info.first, cmd,
-			    from_tty, flags);
+      thr_lane_try_catch_cmd (false, info.second.get (), 0, info.first, cmd,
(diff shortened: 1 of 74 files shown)


lumachad:
Given the PR date, Is this PR active still?

palves:
> Given the PR date, Is this PR active still?

Yes, this is going to take a while to be fully done due to dependency on DWARF, and the PR is helpful to share the progress with Intel.

lumachad:
Switching to draft then.

PR #42, opened 2025-07-11
gdb/testsuite: Test the "info dispatches" command

alexrosen45:
Adds tests for `info dispatches` with multiple dispatches, which brings ROCgdb test line coverage from 68% to 78% for
```
"bfd/elf64-amdgcn.c"
"gdb/amd-dbgapi-target.c"
"gdb/amd-dbgapi-target.h"
"gdb/amdgpu-tdep.c"
"gdb/amdgpu-tdep.h"
"gdb/solib-rocm.c"
```
The test uses two new kernels (`single_wave_kernel1` and `single_wave_kernel2`) in `gdb/testsuite/gdb.rocm/show-info.cpp` to test `info dispatches` with multiple dispatches. There is also a test for `info dispatches` with no dispatches.

### DIFF
--- a/gdb/testsuite/gdb.rocm/show-info.cpp
+++ b/gdb/testsuite/gdb.rocm/show-info.cpp
@@ -39,6 +39,22 @@ __global__ void bit_extract_kernel(uint32_t* C_d, const uint32_t* A_d, size_t N)
     }
 }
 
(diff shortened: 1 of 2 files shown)


alexrosen45:
use _amdeng account instead

PR #55, opened 2026-03-30
binutils/readelf: Remove unused static variables

amd-bfilipov:
Remove static variables do_ctf and do_sframe that are set but never read, causing build failures with LLVM's extended
-Wunused-but-set-variable warning.

These variables are dead code:
- Declared at lines 244-245 as static booleans
- Set to true when --ctf or --sframe options are parsed
- Never actually used or read anywhere in the code
- The actual dump functionality is triggered by request_dump() calls

Build error with -Werror enabled:
binutils/readelf.c:244:13: error: variable 'do_ctf' set but not used
  [-Werror,-Wunused-but-set-variable]
binutils/readelf.c:245:13: error: variable 'do_sframe' set but not used
  [-Werror,-Wunused-but-set-variable]

Fixes: AIROCGDB-553

There is a commit that supresses the issue:  3e011457055bd3868bd05cf80bc1a4482777809b but we should still fix this.
I will send the same patch upstream, this is just a stand in until it gets merged, so we don't block LLVM bumps.

### DIFF
--- a/binutils/readelf.c
+++ b/binutils/readelf.c
@@ -241,8 +241,6 @@ static bool do_dump = false;
 static bool do_version = false;
 static bool do_histogram = false;
 static bool do_debugging = false;
-static bool do_ctf = false;
-static bool do_sframe = false;
 static bool do_arch = false;
 static bool do_notes = false;
 static bool do_archive_index = false;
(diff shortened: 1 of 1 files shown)


lumachad:
Stale.

PR #56, opened 2026-03-31
gdb/amdgcn: Dynamically select which bits to use to hold CORE_ADDR

amd-shahab:
Relies on dbgapi's PR: [Add AMD_DBGAPI_PROCESS_INFO_SIGNIFICANT_ADDRESS_BITS](https://github.com/ROCm/rocm-systems/pull/3287)

To support the various address-spaces from AMDGPU, amdgpu-tdep implements gdbarch hooks to encode the address-space ID in the top bits of a CORE_ADDR.  Exactly which bits to use out of the CORE_ADDR is currently statically built into GDB.  However, for AMDGPU, the "flat" address-space (also known as generic) is configured by the driver, and it is not part of the ABI which bits out of the top byte are used or not to carry meaning.

To address this, librocm-dbgapi >= 0.80 provides a mask of bits which carry information for the current process.  GDB can use this mask to deduce which bits out of a CORE_ADDR can safely be used to encode address space IDs.

This patch provides an implementation for GDB to encode the address-space ID into the top bits reported as unused by dbgapi.  Note that dbgapi reports which bits can be used by any address space, including "global".  This means that the unused bits must be the top ones, and we should account for the fact that "normalized" addresses are sign extended in the top bits.  This means that the top bits of a valid address can be either all 0s or all 1s.  For this reason, we do not support encoding an address space ID whose binary representation on the available bits of a CORE_ADDR is all 1s, so the all 1s case can be reserved for sign-extended addresses.

Bug: AIROCGDB-26
Change-Id: I4cbb41a922443d64003ebc6415547c02ab43afcf

### DIFF
--- a/gdb/amdgpu-tdep.c
+++ b/gdb/amdgpu-tdep.c
@@ -1755,7 +1755,6 @@ amdgpu_address_scope (struct gdbarch *gdbarch, ptid_t ptid, CORE_ADDR address)
     error (_("amd_dbgapi_dwarf_address_space_to_address_space failed"));
 
   amd_dbgapi_status_t ret;
-#if AMD_DBGAPI_VERSION_MAJOR > 0 || AMD_DBGAPI_VERSION_MINOR >= 79
   /* Access to GPU globals can be made from host threads.  When we place a
      watchpoint on such global GPU variable, a matching watchpoint is also
      added on the CPU side for the same address, so it is possible to call here
@@ -1770,11 +1769,6 @@ amdgpu_address_scope (struct gdbarch *gdbarch, ptid_t ptid, CORE_ADDR address)
   ret = amd_dbgapi_address_dependency (process_id, wave_id,
 				       address_space_id, segment_address,
 				       &segment_address_dependency);
-#else
-  ret = amd_dbgapi_address_dependency (address_space_id,
-				       segment_address,
-				       &segment_address_dependency);
-#endif
   if (ret != AMD_DBGAPI_STATUS_SUCCESS)
     error (_("amd_dbgapi_address_dependency failed"));
 
--- a/gdb/amdgpu-tdep.c
+++ b/gdb/amdgpu-tdep.c
@@ -30,16 +30,62 @@
 #include "gdbtypes.h"
 #include "inferior.h"
 #include "language.h"
+#include "observable.h"
 #include "producer.h"
 #include "reggroups.h"
 #include "extract-store-integer.h"
+#include <optional>
 
-/* Bit mask of the address space information in the core address.  */
-constexpr CORE_ADDR AMDGPU_ADDRESS_SPACE_MASK = 0xff00000000000000;
+struct amdgpu_per_inferior
+{
+  amdgpu_per_inferior ()
+    : significant_bits {}
+  {}
 
-/* Bit offset from the start of the core address
-   that represent the address space information.  */
-constexpr unsigned int AMDGPU_ADDRESS_SPACE_BIT_OFFSET = 56;
+  std::optional<amd_dbgapi_segment_address_t> significant_bits = {};
+};
+
+static const registry<inferior>::key<amdgpu_per_inferior> amdgpu_per_inf;
+
+static void
+amdgpu_observer_inferior_appeared (inferior *inf)
+{
+  /* Invalidate the cached data by clearing it.  */
+  amdgpu_per_inf.clear(inf);
+}
+
+static amdgpu_per_inferior &
+get_amdgpu_per_inferior (inferior *inf)
+{
+  return amdgpu_per_inf.try_emplace (inf);
+}
+
+/* Return a bit-mask showing which bits of a segment address are significant
+   for at least one address space in at least one agent.  Any bits outside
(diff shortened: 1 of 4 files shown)


gdb/amdgpu-tdep.c
```diff
@@ -1435,8 +1481,42 @@ static CORE_ADDR
...
+
+  /* The address must not have bits set outside of the significant bits.  */
+  gdb_assert ((significant_bits & address) == address);
+
+  const amd_dbgapi_segment_address_t aspace_id_mask = ~significant_bits;
+  /* Encode the address space in the "free" bits of the segment address.
+     We expect that the free bits are somewhere close to the MSB (do not have
+     to be contiguous).  Skip all the LSBs which can't be touched using ctz
+     as an optimisation.  */
+  size_t shift = __builtin_ctzl (aspace_id_mask);
```
palves: __builtin_ctzl works with long, but long is 32-bit on 64-bit windows.   So this should be __builtin_ctzll instead.

Also looks a bit odd to assume that this works on all compilers without a fallback, though it works on at least gcc and clang, so maybe it's fine.

gdb/amdgpu-tdep.c
```diff
@@ -1454,15 +1534,49 @@ amdgpu_integer_to_address (struct gdbarch *gdbarch,
...
+    return 0;
+
+  /* The all 1s address space is not valid, so 0.  */
+  if ((addr & aspace_id_mask) == aspace_id_mask)
+    return 0;
+
+  arch_addr_space_id aspace_id = 0;
+  size_t aspace_id_shift = 0;
+  /* Skip the majority of LSBs in one go using ctz.  */
+  size_t shift = __builtin_ctzl (aspace_id_mask);
```
palves: __builtin_ctzll

gdb/amdgpu-tdep.c
```diff
@@ -1435,8 +1481,42 @@ static CORE_ADDR
...
+		      sizeof (amd_dbgapi_segment_address_t) * HOST_CHAR_BIT);
+	}
+
+      /* Store one bit of address space there.  */
+      address = ((address & ~(1ull << shift))
+		 | ((static_cast<amd_dbgapi_segment_address_t>
+		     (address_space_id) & 1) << shift));
+      shift++;
+      address_space_id >>= 1;
+    }
```
palves, line 1507: I think this be equivalent to the above.  It'd be more efficient, for not having a nested loop, and you wouldn't need ctz:

    uint64_t mask = aspace_id_mask;
    uint64_t id = address_space_id;
    uint64_t result = 0;

    for (; id != 0 && mask != 0; id >>= 1) 
      {
         /* Isolate the lowest set bit in the mask.  */
          uint64_t lowbit = mask & ~(mask - 1);
          if (id & 1)
             result |= lowbit;
          /* Clear bit from mask.   */
          mask ^= lowbit;
      }
    address |= result;

(disclaimer: completely untested, written in github comment directly.)

gdb/amdgpu-tdep.c
```diff
@@ -1454,15 +1534,49 @@ amdgpu_integer_to_address (struct gdbarch *gdbarch,
...
+  size_t shift = __builtin_ctzl (aspace_id_mask);
+  while ((addr >> shift) != 0)
+    {
+      if (((1ull << shift) & aspace_id_mask) != 0)
+	aspace_id
+	  |= (static_cast<arch_addr_space_id> (!!(addr & (1ull << shift)))
+	      << aspace_id_shift++);
+      shift++;
+    }
+
```
palves, line 1560: Here, I think we can do the same as in amdgpu_segment_address_to_core_address:

    arch_addr_space_id aspace_id = 0;
    uint64_t dest_bit = 1;    /* bit position in resulting ID */
    uint64_t mask = aspace_id_mask;

    /* Iterate for as many bits as are set in the mask.  */
    while (mask != 0)
      {
        /* Isolate lowest set bit in mask.  */
        uint64_t lowbit = mask & ~(mask - 1);
        
        if (addr & lowbit)
          aspace_id |= dest_bit;
  
        dest_bit <<= 1;
        /* Clear bit from mask.   *
        mask ^= lowbit;
      }
  
    return aspace_id;

gdb/amdgpu-tdep.c
amd-shahab: ignored in favour of the [suggested routine](https://github.com/ROCm/ROCgdb/pull/56#discussion_r3092745716).

gdb/amdgpu-tdep.c
amd-shahab: ignored in favour of the [other suggested routine](https://github.com/ROCm/ROCgdb/pull/56#discussion_r3092850688).

gdb/amdgpu-tdep.c
amd-shahab: brilliant!

gdb/amdgpu-tdep.c
amd-shahab: master piece.

amd-shahab:
- get rid of invocations of `ctz()` builtins in favour of better algorithms suggested by @palves

palves (approved):

Issue #12, opened 2022-04-14
Asic family name not present in the sysfs


tekelle:
I am running CentOS 7.9 and ROCm4.3.0.  When I try to run rocgdb I am getting "amd-dbgapi: fatal error: os_agent_id 42028: asic family name not present in the sysfs.
Could not attach process 4832 (rc=-2)

Configuration is:

configure --host=x86_64-pc-linux-gnu --target=x86_64-pc-linux-gnu
                --with-auto-load-dir=$debugdir:$datadir/auto-load
                --with-auto-load-safe-path=$debugdir:$datadir/auto-load
                --with-expat
                --with-gdb-datadir=/opt/rocm-4.3.0/share/rocgdb (relocatable)
                --with-jit-reader-dir=/opt/rocm-4.3.0/lib/gdb (relocatable)
                --without-libunwind-ia64
                --with-lzma
                --with-babeltrace
                --without-intel-pt
                --without-mpfr
                --without-xxhash
                --with-python=/usr
                --with-python-libdir=/usr/lib
                --without-debuginfod
                --without-guile
                --disable-source-highlight
                --with-separate-debug-dir=/opt/rocm-4.3.0/lib/debug (relocatable)

I have:
2 Vega 10 XL/XT [Radeon RX Vega 56/64] (rev c1) and a Vega 10 [Radeon Instinct MI25] (rev 01) cards.

How can this be resolved?  This is a closed system so everything I post will be hand-typed in.

ppanchad-amd:
@tekelle Apologies for the lack of response.  Vega 56/64 and Vega 10 cards are no longer supported in the latest ROCm 6.2. Closing ticket. Thanks!

Issue #11, opened 2022-03-21
Rocgdb failing to launch application with "cannot attach to process" rc=-2


drtpotter:
Hi there,

I'm trying to get **rocgdb** to debug an OpenCL application that is running on my Radeon VII. I'm using ROCM 5.0 with the ROCclr runtime on OpenSUSE tumbleweed.

Every time I try to run the application like this

`rocgdb ./mat_mult_badmem_gdb.exe`

and then type **run** I get this error

```
Starting program: 
mat_mult_badmem_gdb.exe 
Could not attach to process 107215 (rc=-2)
```

I have tried both the binary release of rocgdb (5.0.0) and have **compiled** rocgdb from Github sources according to the [instructions](https://github.com/ROCm-Developer-Tools/ROCgdb/blob/amd-master/README-ROCM.md). The outcome is the same with either binary or source releases and Google hasn't been my friend. Same thing happens while running rocgdb as **root.** The error does not occur if I use the Github source to compile and run vanilla gdb without the AMD-specific goodies, but then I can't access the GPU.

If you can provide any insight into what error code rc=-2 is and how I can work around this issue that would be great. It is preventing me from finishing a course section on how to debug OpenCL programs with rocgdb!

Kind regards,
Toby

drtpotter:
Bug is still present with the 5.1 release.

t-tye:
@drtpotter please could you get a log using the rocgdb command:

set debug amdgpu log-level verbose

drtpotter:
Hey @t-tye 

Thanks for looking into this. After setting the debug I get this error. I'm using Linux Kernel 5.17.

(gdb) run
Starting program: ~/OpenCL_Course/course_material/L4_Debugging/mat_mult_badmem.exe 
amd-dbgapi: amd_dbgapi_process_attach (client_process_id=0x22236f0, process_id=0x23c4fc8) {
amd-dbgapi:    callback: get_os_pid (pid=0x7ffec8f03e04) {
amd-dbgapi:    callback: } = STATUS_SUCCESS, *pid=12657
amd-dbgapi:    attaching process_1 to OS process 12657
amd-dbgapi:    detached process_1
amd-dbgapi:    linux_driver_t statistics (pid 12657): 0 reads (0), 0 writes (0)
amd-dbgapi: fatal error: KFD_IOC_DBG_TRAP_GET_VERSION failed
amd-dbgapi: } = STATUS_FATAL
[...]

Issue #10, opened 2022-01-05
Which version of LLVm is required by ROCgdb to support "DWARF Extensions for Optimized SIMT/SIMD (GPU) Debugging"


Chunming-Zhou:
Which version of LLVm is required by ROCgdb to support "DWARF Extensions for Optimized SIMT/SIMD (GPU) Debugging"?

Maetveis:
According to the release notes ROCM 5.1 supposedly enabled this, but I can't make it work.

I did the following steps:
1. Installed rocm-hip-runtime-dev and rocm-gdb 5.1.1 on ubuntu 20.04 (docker image)
2. Compiled a simple program with the following command `/opt/rocm/llvm/bin/clang++ -L /opt/rocm/lib -lhipamd64 -g -O0 -x hip main.cpp`
3. Run with rocgdb `LD_LIBRARY_PATH=/opt/rocm/lib rocgdb ./a.out`
4. Try to add a breakpoint to the device code: `break hello` or break `main.cpp:5`

Expected:
gdb adds the breakpoint, breaks when it is hit on the device. The values of variables etc can be inspected.

Actual result:
gdb fails to add the breakpoint with this message:
No compiled code for line 5 in file "main.cpp".


```c++
#include <hip/hip_runtime.h>

__global__
void hello(int a) {
    printf("Hello World: %d\n", a);
}

int main() {
    hipLaunchKernelGGL(hello, dim3(1), dim3(1), 0, 0, 10);
    hipDeviceSynchronize();
}
```

Note with previous versions (pre 5.1) adding the breakpoint worked, but no value where displayed in device code.

Can you help me what is needed to make this work?

ppanchad-amd:
@Chunming-Zhou @Maetveis Apologies for the lack of response.  Do you still need assistance with this issue? Thanks!

taylding-amd:
Hi @Chunming-Zhou, at this point, ROCgdb relies on LLVM changes that are not in upstream LLVM. So it requires the version of LLVM that is included with the same ROCm version as the ROCgdb being used.

taylding-amd:
Hi, this issue will be closed now due to inactivity. Please feel free to reopen it or ask follow-up questions should further assistance be required. Thanks!

Issue #9, opened 2021-10-08
Unknown CFA rule. A problem internal to GDB has been detected


zjin-lcf:
While working with DPCPP (https://github.com/intel/llvm) compiling SYCL kernels targeting ROCm (gfx908) I came across the gdb error. It is true that there are some issues with compiling the SYCL kernels. I am not familiar with gdb internals. Does the gdb error provide some hint/clue to the compiler issue ?

https://github.com/zjin-lcf/oneAPI-DirectProgramming/tree/master/babelstream-sycl   

```
rocgdb ./main
Precision: float
Array size: 134.2 MB (=0.1 GB)
Total size: 402.7 MB (=0.4 GB)
Memory access fault by GPU node-2 (Agent handle: 0x532080) on address 0x7ffeb0019000. Reason: Page not present or supervisor privilege.

Thread 6 "main" received signal SIGSEGV, Segmentation fault.
[Switching to AMDGPU Thread 2:3:1:1 (130944,0,0)/0]
copy<float>(cl::sycl::queue&, cl::sycl::buffer<float, 1, cl::sycl::detail::aligned_allocator<char>, void>&, cl::sycl::buffer<float, 1, cl::sycl::detail::aligned_allocator<char>, void>&)::{lambda(cl::sycl::handler&)#1}::operator()(cl::sycl::handler&) const::{lambda(cl::sycl::nd_item<1>)#1}::operator()(cl::sycl::nd_item<1>) const (
/home/release/git/aomp13/ROCgdb/gdb/dwarf2/frame.c:1029: internal-error: Unknown CFA rule.
A problem internal to GDB has been detected,
further debugging may prove unreliable.
```

npmiller:
> The ROCm LLVM based AMD GPU compiler is open source in github so it may be possible to use that to generate correct CFI information. There were some recent fixes which may not be available in the github util the next release is published.

I've just ran into this issue again, do you know if there are any plans to upstream CFI generation for AMD targets?

t-tye:
> > The ROCm LLVM based AMD GPU compiler is open source in github so it may be possible to use that to generate correct CFI information. There were some recent fixes which may not be available in the github util the next release is published.
> 
> I've just ran into this issue again, do you know if there are any plans to upstream CFI generation for AMD targets?

The debug extension support for AMD GPU has started to be upstreamed through the RFC at:

https://discourse.llvm.org/t/rfc-heterogeneous-debug-info/66872

[...]

PR #59, opened 2026-04-02
gdb/dwarf2/expr: Cache the register reads

lancesix:
A recent change in the ROCm compiler started to show some timeout in the
gdb.rocm/deep-stack.exp testcase.  In that test, we have a 12 frames
deep stack, and we print a stack trace.  With the new compiler output,
building this stacktrace would take about 190 seconds... Yes, about
3 minutes to unwind 12 frames.

Looking at the unwind information, we now see this new pattern for
unwinding vector registers:

    VGPR40=[
      DW_OP_regx VGPR40
      DW_OP_swap
      DW_OP_LLVM_user
      DW_OP_LLVM_offset_uconst 0x0
      DW_OP_LLVM_user DW_OP_LLVM_call_frame_entry_reg EXEC
      DW_OP_deref_size 0x8
      DW_OP_LLVM_user DW_OP_LLVM_select_bit_piece 0x20 0x40
    ]

This new construct uses DW_OP_LLVM_select_bit_piece, which allows us to
build a composite which has one piece per lane of the vector register,
each piece referencing either
- the register itself (not touched because the lane was not active, so
  no need to spill), or
- a saved copy of the register to memory (at CFA).

The issue turned out to be that wen building the composite, we ended up
creating up to 64 pieces (one per lane), each can be a dwarf_register
instance.  And even if all of those reference the same register, they
do not share anything more than the regnum.

When we evaluate something like dwarf_composite::is_optimized_out, we
end-up calling dwarf_register::is_optimized_out for each piece in
sequence.  Each one of those calls will call dwarf_register::read and
eventually frame_register_wind, i.e. we fully unwind the register once
for each lane.  If the next frame, and the next, and the next use the
same pattern, we end-up with an impossible complexity of
O(lanes ** depth), which explains how we can end-up with 3 minutes to do
a 12 frames unwinding.

This pattern does not only happen when calling value::optimized_out,
this also happens when calling value->entirely_available for example,
making things very inefficient..

To solve this, this patch makes two things:
- Make it so when we use DW_OP_LLVM_select_bit_piece, all the
  dwarf_register in the resulting composite share a storage,
- Make that storage cache the result of calling frame_register_unwind, so
  we do not have to repeat this for every piece.

The cache is cleared any time we call dwarf_register::write, or if we
detect that it might be outdated (target_changed or memory_changed got
called since we cached the data).

With this, doing the backtrace from the test now takes slightly more
than .5 seconds, which is back to acceptable territory.

    (gdb) bt
    2026-04-02 17:40:05.195 - command started
    #0  base_case () at /home/.../gdb.rocm/deep-stack.cpp:40
    https://github.com/ROCm/ROCgdb/issues/1  0x00007ffff637649c in deep<0u> () at /home/.../gdb.rocm/deep-stack.cpp:54
    https://github.com/ROCm/ROCgdb/issues/2  0x00007ffff637655c in deep<1u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    https://github.com/ROCm/ROCgdb/issues/3  0x00007ffff637661c in deep<2u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    https://github.com/ROCm/ROCgdb/issues/4  0x00007ffff63766e4 in deep<3u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    https://github.com/ROCm/ROCgdb/issues/5  0x00007ffff63767a4 in deep<4u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    https://github.com/ROCm/ROCgdb/issues/6  0x00007ffff6376864 in deep<5u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    https://github.com/ROCm/ROCgdb/issues/7  0x00007ffff6376924 in deep<6u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    https://github.com/ROCm/ROCgdb/issues/8  0x00007ffff63769e4 in deep<7u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    https://github.com/ROCm/ROCgdb/issues/9  0x00007ffff6376aa4 in deep<8u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    https://github.com/ROCm/ROCgdb/issues/10 0x00007ffff6376b64 in deep<9u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    https://github.com/ROCm/ROCgdb/issues/11 0x00007ffff6376c24 in deep<10u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    https://github.com/ROCm/ROCgdb/issues/12 0x00007ffff6376d40 in hip_deep () at /home/.../gdb.rocm/deep-stack.cpp:60
    2026-04-02 17:40:05.753 - command finished
    Command execution time: 0.558060 (cpu), 0.558099 (wall)

### DIFF
--- a/gdb/dwarf2/expr.c
+++ b/gdb/dwarf2/expr.c
@@ -32,12 +32,21 @@
 #include "dwarf2/read.h"
 #include "frame.h"
 #include "gdbsupport/underlying.h"
+#include "gdbsupport/refcounted-object.h"
 #include "gdbarch.h"
 #include "gdbthread.h"
 #include "inferior.h"
 #include "objfiles.h"
 #include "observable.h"
 #include "extract-store-integer.h"
+
+using context_generation_t = ULONGEST;
+
+/* A monotically increasing counter modified every time any inferior memory
+   is touched or any register in any target is modified.  This can be used to
+   know if state (register value) cached by the evaluator might be stale.  */
+
+static context_generation_t context_generation = 0;
 
 /* This holds gdbarch-specific types used by the DWARF expression
    evaluator.  See comments in execute_stack_op.  */
@@ -1244,13 +1253,12 @@ public:
 		  bool on_entry = false, LONGEST offset = 0,
 		  LONGEST bit_suboffset = 0)
     : dwarf_location (arch, offset, bit_suboffset),
-      m_regnum (regnum), m_on_entry (on_entry)
+      m_storage {register_storage_ptr::new_reference
+		   (new register_storage (regnum, on_entry))}
   {}
 
   dwarf_register (const dwarf_register &reg)
-    : dwarf_location (reg),
-      m_regnum (reg.m_regnum),
-      m_on_entry (reg.m_on_entry)
+    : dwarf_location (reg), m_storage (reg.m_storage)
   {}
 
   std::shared_ptr<dwarf_entry> clone () const override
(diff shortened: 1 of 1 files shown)


simark (commented):
Not a complete review, I had 10 minutes free so I made a few superficial comments.

palves (commented):
In the commit log: "that wen building"

gdb/dwarf2/expr.c
```diff
@@ -39,6 +39,14 @@
 #include "observable.h"
 #include "extract-store-integer.h"
 
+using programs_state_generation_t = size_t;
+
+/* A monotically increasing counter modified every time any inferior memory
+   is touched register in any target is modified.  This can be used to know
+   if state (register value) cached by the evaluator might be stall.  */
```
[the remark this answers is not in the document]
lancesix: Done.

gdb/dwarf2/expr.c
```diff
@@ -1277,12 +1283,53 @@ class dwarf_register : public dwarf_location
...
+    bool m_on_entry;
+
+    struct read_cache
+    {
+      read_cache (gdbarch *arch, const frame_info_ptr &frame, int regnum)
+	: m_frame { frame }, m_state_generation { programs_state_generation }
+	{
+	  m_data.resize (register_size (arch, regnum));
+	  read_from_register (frame, regnum, 0, m_data, &m_optimized_out,
+			      &m_unavailable);
```
palves: Since this is always reading the whole register at offset 0, do we need to call read_from_register?  Can we call frame_register_unwind directly?  That's save the local temp buffer (on the heap) plus the memcpy that read_from_register needs to do to handle partial registers.

gdb/dwarf2/expr.c
```diff
@@ -1277,12 +1283,53 @@ class dwarf_register : public dwarf_location
...
+    private:
+      frame_info_ptr m_frame;
+      programs_state_generation_t m_state_generation;
+      bool m_optimized_out;
+      bool m_unavailable;
+      gdb::byte_vector m_data;
+    };
+    std::optional<read_cache> m_read_cache;
+  };
+  std::shared_ptr<register_storage> m_storage;
```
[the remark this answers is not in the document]
lancesix: oh, indeed. I used shared_ptr here as a way to obtain refcounting behaviour, I forgot we had refcounted_object.

Changing this.

gdb/dwarf2/expr.c
lancesix: Indeed.

gdb/dwarf2/expr.c
```diff
@@ -4768,3 +4836,38 @@ dwarf2_evaluate (const gdb_byte *addr, size_t len, bool as_lval,
...
+     really, also applies to lval_memory values), target_changed ends-up being
+     called.  Any such assignment should invalidate any cached value we hold
+     in the expression evaluator.  */
+  gdb::observers::target_changed.attach (expr_observer_target_changed,
+					 "dwarf2_expr");
+
+  /* Note that we can implicitly cache register values which are not currently
+     stored in a register.  For example, if a register was spilled to memory
+     from the previous frame, a memory change could impact a register value
+     we cached.  It is possible to modify such memory without using triggering
```
palves: "without using triggering"

Seems like something off here.

palves (commented):
One comment nit I noticed now.

BTW, you still have "wen" in the commit log.

Otherwise LGTM.

gdb/dwarf2/expr.c
lancesix: Inded.  Fixed.

lancesix:
> BTW, you still have "wen" in the commit log.

Fixed.

palves (approved):

PR #62, opened 2026-04-08
gdb.rocm: replace hipcc with amdclang++ as the HIP compiler

spatrang:
## Motivation

The GDB testsuite uses `hipcc` for HIP compiler discovery, naming, and
documentation. `hipcc` is deprecated and will be removed at some point, so this
switches the testsuite over to `amdclang++`.

### Compiler discovery (`lib/future.exp`)
- Look for `amdclang++` under `$ROCM_PATH/lib/llvm/bin` instead of `hipcc`.
- Rename `HIPCC_FOR_TARGET` to `HIP_COMPILER_FOR_TARGET` (the environment
  override).
- Rename `gdb_find_hipcc` / `find_hipcc` to `gdb_find_hip_compiler` /
  `find_hip_compiler`.

### Compile and link flags (`lib/gdb.exp`, `boards/hip.exp`)
- `amdclang++` does not infer the input language from the file extension, so
  tag each input explicitly: `-x hip` before C/C++/HIP sources and `-x none`
  before object files/archives. A single invocation can then handle a mix of
  sources and objects (e.g. `set_unbuffered_mode.o` / `hip-driver.o` injected
  via `ldflags`).
- Pass `--hip-link` only when producing an executable.
- Pass `--rocm-path` only when `ROCM_PATH` is set; otherwise defer to
  `amdclang++`'s own HIP discovery.
- Drop the explicit `-O0` (`amdclang++` already defaults to `-O0`).
- `amdclang++` is clang-based and accepts `-fdiagnostics-color`, so stop
  skipping it for HIP; this keeps color escape sequences out of `gdb.log`.

### Testcases
- Rename `allow_hipcc_tests` to `allow_hip_tests` in `lib/rocm.exp`,
  `boards/hip.exp`, and all `gdb.rocm/` and `gdb.perf/` testcases.
- Replace `hipcc`'s `--genco` with `amdclang++`'s `--cuda-device-only` in the
  two code-object tests (`code-object-load-while-breakpoint-hit.exp`,
  `snapshot-objfile-on-load.exp`).

### Documentation (`gdb/doc/gdb.texinfo`)
- Update the example compile command to `amdclang++ -x hip --hip-link ...`,
  keeping the note about the equivalent `hipcc` invocation.

### DIFF
--- a/gdb/testsuite/boards/hip.exp
+++ b/gdb/testsuite/boards/hip.exp
@@ -302,6 +302,10 @@ proc gdb_compile {source dest type options} {
 	#    '-mllvm -amdgpu-spill-cfi-saved-regs'
 	#    [-Wunused-command-line-argument]
 	#
+	# Pass --rocm-path explicitly so amdclang++ can locate the HIP
+	# headers and device libraries when it's invoked from the board
+	# via PATH rather than an absolute path.  rocm_path comes from
+	# rocm.exp and defaults to /opt/rocm if ROCM_PATH is not set.
 	global rocm_path
 	set hip_options "early_flags=\
 			 --rocm-path=$rocm_path\
--- a/gdb/testsuite/boards/hip.exp
+++ b/gdb/testsuite/boards/hip.exp
@@ -302,7 +302,9 @@ proc gdb_compile {source dest type options} {
 	#    '-mllvm -amdgpu-spill-cfi-saved-regs'
 	#    [-Wunused-command-line-argument]
 	#
+	global rocm_path
 	set hip_options "early_flags=\
+			 --rocm-path=$rocm_path\
 			 -O0\
 			 -std=gnu++11\
 			 -fgpu-rdc\
@@ -377,6 +379,7 @@ proc gdb_compile {source dest type options} {
 		lappend options "ldflags=$gdb_saved_hip_driver_obj"
 
 		lappend options $hip_options
+		lappend options "early_flags=--hip-link"
 
 		# All compiled.  Now link, and we're done.
 		return [hip_org_gdb_compile $source $dest $type $options]
@@ -422,6 +425,7 @@ proc gdb_compile {source dest type options} {
 	    }
 
 	    lappend options $hip_options
+	    lappend options "early_flags=-x hip"
 	}
     }
     hip_org_gdb_compile $source $dest $type $options
--- a/gdb/testsuite/boards/hip.exp
+++ b/gdb/testsuite/boards/hip.exp
@@ -15,7 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 # This file is a dejagnu "board file" and is used to run the testsuite
-# against C/C++ device code compiled with hipcc.
+# against C/C++ device code compiled with amdclang++ (HIP).
 #
 # Example usage:
 #  bash$ make check RUNTESTFLAGS='--target_board=hip'
@@ -80,8 +80,8 @@ load_lib rocm.exp
 load_generic_config "unix"
 process_multilib_options ""
 
-set_board_info compiler [find_hipcc]
-set_board_info c++compiler [find_hipcc]
+set_board_info compiler [find_hip_compiler]
+set_board_info c++compiler [find_hip_compiler]
 
 set_board_info gdb,cannot_call_functions 1
 
@@ -149,8 +149,16 @@ proc skip_opencl_tests {} {
 # Yup, we don't want to run the gdb.rocm/ tests against this board.
 # That would be kind of recursive.  Those tests really expect that
 # "main()" is host code and start kernels themselves.
-proc allow_hipcc_tests {} {
+proc allow_hip_tests {} {
     return 0
+}
+
+# Backwards compatibility alias.  The testsuite previously used
+# allow_hipcc_tests when hipcc was the HIP compiler wrapper.  Keep
+# this alias so that any out-of-tree board files or scripts that
+# reference the old name continue to work.
+proc allow_hipcc_tests {} {
+    return [allow_hip_tests]
 }
 
 # Override this to fail faster.  If we let the default run, tests
@@ -271,10 +279,14 @@ proc gdb_compile {source dest type options} {
 
 	# We need:
 	#
-	# . explicit -O0, because hipcc optimizes by default.
+	# . explicit -O0, because the legacy hipcc wrapper optimized by
+	#   default.  amdclang++ does not, but we keep -O0 explicit for
+	#   clarity and to stay safe if defaults ever change.
 	#
-	# . -std=gnu++11 because hipcc defaults to -std=c++11, and
-	#   some testcases assume GNU extensions.
+	# . -std=gnu++11 because the legacy hipcc wrapper defaulted to
+	#   -std=c++11, and some testcases assume GNU extensions.
+	#   amdclang++ defaults to a later standard but we keep this
+	#   for consistency.
 	#
 	# . -fgpu-rdc enables separate compilation mode, so we can
 	#   compile the kernel from multiple translation units.
@@ -303,7 +315,7 @@ proc gdb_compile {source dest type options} {
 	} elseif {[info exists CC_FOR_TARGET]} {
 	    set compiler $CC_FOR_TARGET
 	}
-	if {[string first "hipcc" $compiler] != -1} {
+	if {[string first "amdclang++" $compiler] != -1} {
 
 	    # Remove "c++" from the options when linking, otherwise we
 	    # would get:
--- a/gdb/testsuite/boards/hip.exp
+++ b/gdb/testsuite/boards/hip.exp
@@ -15,7 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 # This file is a dejagnu "board file" and is used to run the testsuite
-# against C/C++ device code compiled with hipcc.
+# against C/C++ device code compiled with amdclang++ (HIP).
 #
 # Example usage:
 #  bash$ make check RUNTESTFLAGS='--target_board=hip'
@@ -80,8 +80,8 @@ load_lib rocm.exp
 load_generic_config "unix"
 process_multilib_options ""
 
-set_board_info compiler [find_hipcc]
-set_board_info c++compiler [find_hipcc]
+set_board_info compiler [find_hip_compiler]
+set_board_info c++compiler [find_hip_compiler]
 
 set_board_info gdb,cannot_call_functions 1
 
@@ -149,7 +149,7 @@ proc skip_opencl_tests {} {
 # Yup, we don't want to run the gdb.rocm/ tests against this board.
 # That would be kind of recursive.  Those tests really expect that
 # "main()" is host code and start kernels themselves.
-proc allow_hipcc_tests {} {
+proc allow_hip_tests {} {
     return 0
 }
 
@@ -271,10 +271,7 @@ proc gdb_compile {source dest type options} {
 
 	# We need:
 	#
-	# . explicit -O0, because hipcc optimizes by default.
-	#
-	# . -std=gnu++11 because hipcc defaults to -std=c++11, and
-	#   some testcases assume GNU extensions.
+	# . -std=gnu++11 because some testcases rely on GNU extensions.
 	#
 	# . -fgpu-rdc enables separate compilation mode, so we can
 	#   compile the kernel from multiple translation units.
@@ -290,8 +287,15 @@ proc gdb_compile {source dest type options} {
 	#    '-mllvm -amdgpu-spill-cfi-saved-regs'
 	#    [-Wunused-command-line-argument]
 	#
-	set hip_options "early_flags=\
-			 -O0\
+	# When ROCM_PATH is set in the environment, pass --rocm-path
+	# so amdclang++ uses that specific ROCm install for HIP.
+	# When unset, let amdclang++ auto-discover (parent of its own
+	# LLVM dir, falling back on /opt/rocm).
+	set rocm_path_arg ""
+	if {[info exists ::env(ROCM_PATH)] && $::env(ROCM_PATH) ne ""} {
+	    set rocm_path_arg "--rocm-path=$::env(ROCM_PATH)"
+	}
+	set hip_options "early_flags=$rocm_path_arg\
 			 -std=gnu++11\
 			 -fgpu-rdc\
 			 -mllvm -amdgpu-spill-cfi-saved-regs\
@@ -303,7 +307,7 @@ proc gdb_compile {source dest type options} {
 	} elseif {[info exists CC_FOR_TARGET]} {
 	    set compiler $CC_FOR_TARGET
 	}
-	if {[string first "hipcc" $compiler] != -1} {
+	if {[string first "amdclang++" $compiler] != -1} {
 
 	    # Remove "c++" from the options when linking, otherwise we
 	    # would get:
@@ -362,9 +366,13 @@ proc gdb_compile {source dest type options} {
 		    # original may be automatically deleted.
 		    remote_download host $hip_driver_obj $gdb_saved_hip_driver_obj
 		}
-		lappend options "ldflags=$gdb_saved_hip_driver_obj"
+		# Prefix "-x none" so amdclang++ does not treat the
+		# driver .o as HIP source (the preceding inputs on
+		# the command line are .cpp tagged "-x hip").
+		lappend options "ldflags=-x none $gdb_saved_hip_driver_obj"
 
 		lappend options $hip_options
+		lappend options "early_flags=--hip-link"
 
 		# All compiled.  Now link, and we're done.
 		return [hip_org_gdb_compile $source $dest $type $options]
@@ -410,6 +418,7 @@ proc gdb_compile {source dest type options} {
 	    }
 
 	    lappend options $hip_options
+	    lappend options "early_flags=-x hip"
 	}
     }
     hip_org_gdb_compile $source $dest $type $options
--- a/gdb/testsuite/lib/future.exp
+++ b/gdb/testsuite/lib/future.exp
@@ -121,20 +121,28 @@ proc gdb_find_rustc {} {
     return $rustc
 }
 
-proc gdb_find_hipcc {} {
+proc gdb_find_hip_compiler {} {
     global tool_root_dir
     if {![is_remote host]} {
-	set hipcc [lookfor_file $tool_root_dir hipcc]
-	if {$hipcc eq "" && [info exists ::env(ROCM_PATH)]} {
-	    set hipcc [lookfor_file $::env(ROCM_PATH)/bin hipcc]
+	set compiler [lookfor_file $tool_root_dir amdclang++]
+	if {$compiler eq "" && [info exists ::env(ROCM_PATH)]} {
+	    set compiler [lookfor_file $::env(ROCM_PATH)/llvm/bin amdclang++]
 	}
-	if {$hipcc eq ""} {
-	    set hipcc hipcc
+	if {$compiler eq ""} {
+	    set compiler [lookfor_file /opt/rocm/llvm/bin amdclang++]
+	}
+	if {$compiler eq ""} {
+	    set compiler amdclang++
 	}
     } else {
-	set hipcc ""
+	set compiler ""
     }
-    return $hipcc
+    return $compiler
+}
+
+# Backwards compatibility alias.
+proc gdb_find_hipcc {} {
+    return [gdb_find_hip_compiler]
 }
 
 proc gdb_find_ldd {} {
@@ -343,7 +351,7 @@ proc gdb_default_target_compile_1 {source destfile type options} {
 	    if {[board_info $dest exists hipcompiler]} {
 		set compiler [target_info hipcompiler]
 	    } else {
-		set compiler [find_hipcc]
+		set compiler [find_hip_compiler]
 	    }
 	}
 
@@ -409,6 +417,7 @@ proc gdb_default_target_compile_1 {source destfile type options} {
     global GO_FOR_TARGET
     global GO_LD_FOR_TARGET
     global RUSTC_FOR_TARGET
+    global HIP_COMPILER_FOR_TARGET
     global HIPCC_FOR_TARGET
 
     if {[info exists GNATMAKE_FOR_TARGET]} {
@@ -456,7 +465,11 @@ proc gdb_default_target_compile_1 {source destfile type options} {
 	}
     }
 
-    if {[info exists HIPCC_FOR_TARGET]} {
+    if {[info exists HIP_COMPILER_FOR_TARGET]} {
+	if {$compiler_type == "hip"} {
+	    set compiler $HIP_COMPILER_FOR_TARGET
+	}
+    } elseif {[info exists HIPCC_FOR_TARGET]} {
 	if {$compiler_type == "hip"} {
 	    set compiler $HIPCC_FOR_TARGET
 	}
@@ -762,12 +775,20 @@ if {[info procs find_rustc] == ""} {
     rename gdb_find_rustc ""
 }
 
-if {[info procs find_hipcc] == ""} {
-    rename gdb_find_hipcc find_hipcc
+if {[info procs find_hip_compiler] == ""} {
+    rename gdb_find_hip_compiler find_hip_compiler
     set use_gdb_compile(hip) 1
     gdb_note [join [list $note_prefix "HIP" $note_suffix] ""]
 } else {
-    rename gdb_find_hipcc ""
+    rename gdb_find_hip_compiler ""
+}
+
+# Backwards compatibility: provide find_hipcc as an alias that always
+# delegates to find_hip_compiler (which is guaranteed to exist at this
+# point, either from the rename above or from an external definition).
+catch {rename gdb_find_hipcc ""}
+if {[info procs find_hipcc] == ""} {
+    proc find_hipcc {} { return [find_hip_compiler] }
 }
 
 # If dejagnu's default_target_compile is missing support for any language,
--- a/gdb/testsuite/lib/future.exp
+++ b/gdb/testsuite/lib/future.exp
@@ -121,20 +121,20 @@ proc gdb_find_rustc {} {
     return $rustc
 }
 
-proc gdb_find_hipcc {} {
+proc gdb_find_hip_compiler {} {
     global tool_root_dir
     if {![is_remote host]} {
-	set hipcc [lookfor_file $tool_root_dir hipcc]
-	if {$hipcc eq "" && [info exists ::env(ROCM_PATH)]} {
-	    set hipcc [lookfor_file $::env(ROCM_PATH)/bin hipcc]
+	set compiler [lookfor_file $tool_root_dir amdclang++]
+	if {$compiler eq "" && [info exists ::env(ROCM_PATH)]} {
+	    set compiler [lookfor_file $::env(ROCM_PATH)/lib/llvm/bin amdclang++]
 	}
-	if {$hipcc eq ""} {
-	    set hipcc hipcc
+	if {$compiler eq ""} {
+	    set compiler amdclang++
 	}
     } else {
-	set hipcc ""
+	set compiler ""
     }
-    return $hipcc
+    return $compiler
 }
 
 proc gdb_find_ldd {} {
@@ -343,7 +343,7 @@ proc gdb_default_target_compile_1 {source destfile type options} {
 	    if {[board_info $dest exists hipcompiler]} {
 		set compiler [target_info hipcompiler]
 	    } else {
-		set compiler [find_hipcc]
+		set compiler [find_hip_compiler]
 	    }
 	}
 
@@ -409,7 +409,7 @@ proc gdb_default_target_compile_1 {source destfile type options} {
     global GO_FOR_TARGET
     global GO_LD_FOR_TARGET
     global RUSTC_FOR_TARGET
-    global HIPCC_FOR_TARGET
+    global HIP_COMPILER_FOR_TARGET
 
     if {[info exists GNATMAKE_FOR_TARGET]} {
 	if { $compiler_type == "ada" } {
@@ -456,9 +456,9 @@ proc gdb_default_target_compile_1 {source destfile type options} {
 	}
     }
 
-    if {[info exists HIPCC_FOR_TARGET]} {
+    if {[info exists HIP_COMPILER_FOR_TARGET]} {
 	if {$compiler_type == "hip"} {
-	    set compiler $HIPCC_FOR_TARGET
+	    set compiler $HIP_COMPILER_FOR_TARGET
 	}
     }
 
@@ -762,12 +762,12 @@ if {[info procs find_rustc] == ""} {
     rename gdb_find_rustc ""
 }
 
-if {[info procs find_hipcc] == ""} {
-    rename gdb_find_hipcc find_hipcc
+if {[info procs find_hip_compiler] == ""} {
+    rename gdb_find_hip_compiler find_hip_compiler
     set use_gdb_compile(hip) 1
     gdb_note [join [list $note_prefix "HIP" $note_suffix] ""]
 } else {
-    rename gdb_find_hipcc ""
+    rename gdb_find_hip_compiler ""
 }
 
 # If dejagnu's default_target_compile is missing support for any language,
--- a/gdb/testsuite/lib/future.exp
+++ b/gdb/testsuite/lib/future.exp
@@ -126,10 +126,7 @@ proc gdb_find_hip_compiler {} {
     if {![is_remote host]} {
 	set compiler [lookfor_file $tool_root_dir amdclang++]
 	if {$compiler eq "" && [info exists ::env(ROCM_PATH)]} {
-	    set compiler [lookfor_file $::env(ROCM_PATH)/llvm/bin amdclang++]
-	}
-	if {$compiler eq ""} {
-	    set compiler [lookfor_file /opt/rocm/llvm/bin amdclang++]
+	    set compiler [lookfor_file $::env(ROCM_PATH)/lib/llvm/bin amdclang++]
 	}
 	if {$compiler eq ""} {
 	    set compiler amdclang++
@@ -138,11 +135,6 @@ proc gdb_find_hip_compiler {} {
 	set compiler ""
     }
     return $compiler
-}
-
-# Backwards compatibility alias.
-proc gdb_find_hipcc {} {
-    return [gdb_find_hip_compiler]
 }
 
 proc gdb_find_ldd {} {
@@ -776,14 +768,6 @@ if {[info procs find_hip_compiler] == ""} {
     gdb_note [join [list $note_prefix "HIP" $note_suffix] ""]
 } else {
     rename gdb_find_hip_compiler ""
-}
-
-# Backwards compatibility: provide find_hipcc as an alias that always
-# delegates to find_hip_compiler (which is guaranteed to exist at this
-# point, either from the rename above or from an external definition).
-catch {rename gdb_find_hipcc ""}
-if {[info procs find_hipcc] == ""} {
-    proc find_hipcc {} { return [find_hip_compiler] }
 }
 
 # If dejagnu's default_target_compile is missing support for any language,
--- a/gdb/testsuite/lib/future.exp
+++ b/gdb/testsuite/lib/future.exp
@@ -418,7 +418,6 @@ proc gdb_default_target_compile_1 {source destfile type options} {
     global GO_LD_FOR_TARGET
     global RUSTC_FOR_TARGET
     global HIP_COMPILER_FOR_TARGET
-    global HIPCC_FOR_TARGET
 
     if {[info exists GNATMAKE_FOR_TARGET]} {
 	if { $compiler_type == "ada" } {
@@ -468,10 +467,6 @@ proc gdb_default_target_compile_1 {source destfile type options} {
     if {[info exists HIP_COMPILER_FOR_TARGET]} {
 	if {$compiler_type == "hip"} {
 	    set compiler $HIP_COMPILER_FOR_TARGET
-	}
-    } elseif {[info exists HIPCC_FOR_TARGET]} {
-	if {$compiler_type == "hip"} {
-	    set compiler $HIPCC_FOR_TARGET
 	}
     }
 
--- a/gdb/testsuite/lib/gdb.exp
+++ b/gdb/testsuite/lib/gdb.exp
@@ -6248,6 +6248,52 @@ proc gdb_compile {source dest type options} {
 	}
     }
 
+    # For HIP executables, amdclang++ needs -x hip to treat .cpp/.c as
+    # HIP source, but -x hip applied to a .o file makes amdclang++ try
+    # to parse the object as HIP source, which fails.  Since gdb_compile
+    # may later inject additional .o inputs (e.g. set_unbuffered_mode.o),
+    # we cannot safely mix any source files with any .o inputs in a
+    # single amdclang++ invocation.  To avoid that, pre-compile each
+    # source file to its own .o first (with -x hip), then recurse with
+    # an all-.o source list.  The recursive call will not add -x hip
+    # and the final link will work correctly.
+    if {[lsearch -exact $options hip] != -1
+	&& $type eq "executable"
+	&& [lsearch -exact $options getting_compiler_info] == -1} {
+	set has_source 0
+	foreach src $source {
+	    set ext [file extension $src]
+	    if {$ext ne ".o" && $ext ne ".a"} {
+		set has_source 1
+		break
+	    }
+	}
+	if {$has_source} {
+	    set new_source {}
+	    set idx 0
+	    foreach src $source {
+		set ext [file extension $src]
+		if {$ext eq ".o" || $ext eq ".a"} {
+		    lappend new_source $src
+		} else {
+		    # Include an index in the temp filename so that
+		    # two source files with the same basename (e.g.
+		    # dir1/foo.cpp and dir2/foo.cpp) do not collide
+		    # to the same intermediate .o.
+		    set out \
+			[standard_temp_file \
+			     ${idx}-[file tail $src].hip.o]
+		    set res [gdb_compile $src $out object $options]
+		    if {$res ne ""} {
+			return $res
+		    }
+		    lappend new_source $out
+		}
+		incr idx
+	    }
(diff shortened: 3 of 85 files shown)


gdb/testsuite/lib/gdb.exp
```diff
@@ -6269,7 +6269,7 @@ proc gdb_compile {source dest type options} {
     if {[lsearch -exact $options rust] != -1} {
 	# -fdiagnostics-color is not a rustcc option.
     } elseif {[lsearch -exact $options hip] != -1} {
-	# -fdiagnostics-color is not a hipcc option.
+	# -fdiagnostics-color is not needed for HIP compilation.
```
lumachad: Is it not needed or simply not an option like the old comment says? Does it work with amdclang++ now?

lumachad (commented):
Thanks! Looks mostly good to me. Made some comments about small adjustments.

lancesix (commented):
I only skimmed thought the patch at this point.  Note that some of what you are touching comes from upstream, and we might want to land the changes there as well.  I'd probably advise for having this changed upstream first, then we can import it into amd-staging.

gdb/testsuite/lib/gdb.exp
```diff
@@ -6269,7 +6269,9 @@ proc gdb_compile {source dest type options} {
     if {[lsearch -exact $options rust] != -1} {
 	# -fdiagnostics-color is not a rustcc option.
     } elseif {[lsearch -exact $options hip] != -1} {
-	# -fdiagnostics-color is not a hipcc option.
+	# The legacy hipcc wrapper did not support -fdiagnostics-color.
+	# amdclang++ does support it, but we skip it for HIP compilation
+	# to keep the behavior consistent.
```
palves: I don't think we need to keep this.  If we can pass -fdiagnostics-color=never to amdclang++, then it's helpful to do it, to avoid having escape sequences in gdb.log.

gdb/testsuite/lib/gdb.exp
```diff
@@ -6741,9 +6760,9 @@ proc gdb_compile {source dest type options} {
 		set unbuf_obj_options {nowarnings}
 		if {[lsearch -exact $options hip] != -1} {
 		    lappend unbuf_obj_options hip
-		    # While HIPCC makes Clang treat .cc/.cpp files as
-		    # HIP files (via '-x hip'), it treats .c files as
-		    # C files (via '-x c').  Since we have a .c file,
+		    # While amdclang++ treats .cc/.cpp files as HIP
+		    # files (via '-x hip'), it treats .c files as C
+		    # files (via '-x c').  Since we have a .c file,
```
palves: This comment doesn't make much sense after the change.  Before, this:

      "# While HIPCC makes Clang treat .cc/.cpp files as HIP files (via '-x hip'), it treats .c files as
       # C files (via '-x c')"

was saying that is you do "hipcc foo.cc", hipcc invokes clang with "-x hip foo.cc" automatically.
OTOH, it doesn't pass "-x hip" to clang if you do "hipcc foo.c", it passes "-x c" instead.

With amdclang++ you always have to _explicitly_ add the "-x hip" for all of .cc/.cpp/.c files.

So just remove the whole comment.

gdb/testsuite/lib/gdb.exp
```diff
@@ -6796,19 +6815,19 @@ proc gdb_compile {source dest type options} {
...
-		    #  $ hipcc source.cc set_unbuffered_mode_obj.o
+		    #  $ amdclang++ source.cc set_unbuffered_mode_obj.o
 		    #
 		    # Unfortunately, if a foo.o file appears after a
-		    # bar.cc file, HIPCC treats the foo.o file as a
-		    # source file, and invokes Clang with "clang -x
-		    # hip foo.o", which fails.
+		    # bar.cc file, the compiler treats the foo.o file
+		    # as a source file, and invokes Clang with
+		    # "clang -x hip foo.o", which fails.
```
palves: Does amdclang++ invoke clang, or is amdclang++ clang itself?

gdb/testsuite/lib/gdb.exp
```diff
@@ -6796,19 +6815,19 @@ proc gdb_compile {source dest type options} {
...
 		    #
-		    # HIPCC works correctly if we put the .o file
-		    # first, though, like:
+		    # It works correctly if we put the .o file first,
+		    # though, like:
 		    #
-		    #  $ hipcc set_unbuffered_mode_obj.o source.cc
+		    #  $ amdclang++ set_unbuffered_mode_obj.o source.cc
```
palves: Is this still true with amdclang++?

gdb/testsuite/lib/future.exp
```diff
@@ -456,7 +465,11 @@ proc gdb_default_target_compile_1 {source destfile type options} {
 	}
     }
 
-    if {[info exists HIPCC_FOR_TARGET]} {
+    if {[info exists HIP_COMPILER_FOR_TARGET]} {
+	if {$compiler_type == "hip"} {
+	    set compiler $HIP_COMPILER_FOR_TARGET
+	}
+    } elseif {[info exists HIPCC_FOR_TARGET]} {
```
palves: Do we still manage to run the testsuite with hipcc if we set HIPCC_FOR_TARGET?  Did you try that?

If it turns out it doesn't work, then we should either make it work, or just remove the HIPCC_FOR_TARGET support completely.

gdb/testsuite/lib/gdb.exp
spatrang: Fixed. amdclang++ is clang (it's a symlink/wrapper that invokes clang directly), so the old comment about "invokes Clang with..." was misleading. The updated comment simply says amdclang++ treats the .o file as HIP source due to -x hip, without the incorrect indirection.

gdb/testsuite/lib/gdb.exp
spatrang: Tested: with amdclang++, putting the .o first does not help on its own -- -x hip in early_flags applies to all inputs regardless of ordering. However, this code path still works correctly because the conditional -x hip logic in gdb_compile only adds -x hip when actual source files (.cpp/.c) are present, and omits it when linking .o-only inputs. Updated the comment to reflect this.

gdb/testsuite/lib/future.exp
spatrang: Tested: `HIPCC_FOR_TARGET=/path/to/hipcc` does not work -- our new `-x hip` and `--hip-link` flags are added unconditionally for HIP compilations in gdb_compile, and hipcc doesn't handle them the same way. Since hipcc is deprecated and the flags are fundamentally different, I'll remove `HIPCC_FOR_TARGET` support entirely rather than adding complexity to support both code paths.

gdb/testsuite/lib/gdb.exp
palves: > However, this code path still works correctly because the conditional -x hip logic in gdb_compile only adds -x hip when actual source files (.cpp/.c) are present, and omits it when linking .o-only inputs. 

But we don't end up here with .o-only inputs.  We end up here compiling some source file, so gdb_compile will end up with both a .o file and and .cpp file in the intputs, so it'll add -x hip.

You can easily test this on linux with the following hack:

```
--- c/gdb/testsuite/lib/gdb.exp
+++ w/gdb/testsuite/lib/gdb.exp
@@ -6701,7 +6701,7 @@ proc gdb_compile {source dest type options} {
   }
 
     if { $type == "executable" } {
-       if { ([istarget "*-*-mingw*"]
+       if { 1 || ([istarget "*-*-mingw*"]
              || [istarget "*-*-*djgpp"]
              || [istarget "*-*-cygwin*"])} {
            # Force output to unbuffered mode, by linking in an object file
```
and that shows exactly that -x hip + .o issue:

```
Executing on host: /opt/therock-dist-linux-gfx103X-dgpu-7.13.0a20260416/llvm/bin/amdclang++  --hip-link -x hip -O0 -mllvm=-amdgpu-spill-cfi-saved-regs -Wno-unused-command-line-argument /home/pedro/rocm/gdb/build/gdb/testsuite/set_unbuffered_mode-hip.o /home/pedro/rocm/gdb/build/gdb/testsuite/temp/2878249/device_enumerator.cpp  -fdiagnostics-color=never -w -g  -lm   -o /home/pedro/rocm/gdb/build/gdb/testsuite/temp/2878249/device_enumerator.x    (timeout = 300)
builtin_spawn -ignore SIGHUP /opt/therock-dist-linux-gfx103X-dgpu-7.13.0a20260416/llvm/bin/amdclang++ --hip-link -x hip -O0 -mllvm=-amdgpu-spill-cfi-saved-regs -Wno-unused-command-line-argument /home/pedro/rocm/gdb/build/gdb/testsuite/set_unbuffered_mode-hip.o /home/pedro/rocm/gdb/build/gdb/testsuite/temp/2878249/device_enumerator.cpp -fdiagnostics-color=never -w -g -lm -o /home/pedro/rocm/gdb/build/gdb/testsuite/temp/2878249/device_enumerator.x
pid is 2878354 -2878354
/home/pedro/rocm/gdb/build/gdb/testsuite/set_unbuffered_mode-hip.o:1:1: error: 
      expected unqualified-id
    1 | <U+007F>ELF<U+0002><U+0001><U+0001><U+0000><U+0000><U+0000>...
      | ^
/home/pedro/rocm/gdb/build/gdb/testsuite/set_unbuffered_mode-hip.o:1:41: error: 
      source file is not valid UTF-8
    1 |   ...<98><U+0003><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000>...
      |      ^
...
```

Sounds to me like for hip we may need to make gdb_compile compile all .cpp/.c files to .o files first, and then do a separate final link step.

palves:
> The boards/hip.exp changes need to be tested with `--target_board=hip`. A smoke test of running a few testcases like for example gdb.base/break.exp, instead of the full testsuite, is probably enough.

Please make sure you test this.  I tried it now with:

$ make check RUNTESTFLAGS="--target_board=hip" TESTS="gdb.base/break.exp"

and that fails with:

```
gdb compile failed, /home/pedro/rocm/gdb/build/gdb/testsuite/../../../src/gdb/testsuite/lib/hip/hip-test-wrapper.cc:24:10: fatal error: 
      'hip/hip_runtime.h' file not found
   24 | #include <hip/hip_runtime.h>
      |          ^~~~~~~~~~~~~~~~~~~
1 error generated.

                === gdb Summary ===

# of untested testcases         1
```

Without your patch, it compiles OK.  (It then fails for other unrelated reasons.)

spatrang:
> > The boards/hip.exp changes need to be tested with `--target_board=hip`. A smoke test of running a few testcases like for example gdb.base/break.exp, instead of the full testsuite, is probably enough.
> 
> Please make sure you test this. I tried it now with:
> 
> $ make check RUNTESTFLAGS="--target_board=hip" TESTS="gdb.base/break.exp"
> 
> and that fails with:
> 
> ```
> gdb compile failed, /home/pedro/rocm/gdb/build/gdb/testsuite/../../../src/gdb/testsuite/lib/hip/hip-test-wrapper.cc:24:10: fatal error: 
>       'hip/hip_runtime.h' file not found
>    24 | #include <hip/hip_runtime.h>
>       |          ^~~~~~~~~~~~~~~~~~~
> 1 error generated.
> 
>                 === gdb Summary ===
> 
> # of untested testcases         1
> ```
> 
> Without your patch, it compiles OK. (It then fails for other unrelated reasons.)

Fixed. The --target_board=hip path broke because amdclang++ (unlike hipcc) doesn't automatically discover the HIP headers -- it needs --rocm-path to locate them. Added --rocm-path=$rocm_path to the board's compilation flags, along with the -x hip / --hip-link split (compile vs link). Tested with make check RUNTESTFLAGS="--target_board=hip" TESTS="gdb.base/break.exp" -- compilation succeeds, zero hip_runtime.h errors.

spatrang:
> > The boards/hip.exp changes need to be tested with `--target_board=hip`. A smoke test of running a few testcases like for example gdb.base/break.exp, instead of the full testsuite, is probably enough.
> 
> Please make sure you test this. I tried it now with:
> 
> $ make check RUNTESTFLAGS="--target_board=hip" TESTS="gdb.base/break.exp"
> 
> and that fails with:
> 
> ```
> gdb compile failed, /home/pedro/rocm/gdb/build/gdb/testsuite/../../../src/gdb/testsuite/lib/hip/hip-test-wrapper.cc:24:10: fatal error: 
>       'hip/hip_runtime.h' file not found
>    24 | #include <hip/hip_runtime.h>
>       |          ^~~~~~~~~~~~~~~~~~~
> 1 error generated.
> 
>                 === gdb Summary ===
> 
> # of untested testcases         1
> ```
> 
> Without your patch, it compiles OK. (It then fails for other unrelated reasons.)

@palves 
Implemented your suggestion -- for HIP, gdb_compile now pre-compiles each .cpp/.c source to its own .o before the final link step, mirroring the pattern boards/hip.exp already uses. In the unbuffered-mode handler, we now:

Loop over $source and recursively call gdb_compile to produce a .o for each source file (with -x hip, since they're sources)
Append set_unbuffered_mode.o to the now all-.o source list
Strip -x hip from early_flags for the final link (since we only have .o inputs)
Verified with your hack (if { 1 || ...): before the fix, set_unbuffered_mode-hip.o:1:1: error: expected unqualified-id reproduced exactly as you reported. After the fix, the final link becomes amdclang++ --hip-link -O0 ... device_enumerator.cpp.hip.o set_unbuffered_mode-hip.o ... -o device_enumerator.x (only .o inputs, no -x hip) and succeeds

gdb/testsuite/lib/gdb.exp
```diff
@@ -6806,18 +6807,52 @@ proc gdb_compile {source dest type options} {
...
+		    # Strip -x hip from early_flags in options since
+		    # we no longer have any source files in the
+		    # final link invocation.
+		    set stripped_options {}
+		    foreach opt $options {
+			if {[regexp {^early_flags=(.*)} $opt -> val]} {
+			    regsub -all {(^| )-x hip( |$)} \
+				$val { } val
+			    set val [string trim $val]
+			    lappend stripped_options "early_flags=$val"
```
palves: Can't we / shouldn't we skip the lappend if $val is empty?

palves:
> Fixed. The --target_board=hip path broke because amdclang++ (unlike hipcc) doesn't automatically discover the HIP headers -- it needs --rocm-path to locate them. Added --rocm-path=$rocm_path to the board's compilation flags, along with the -x hip / --hip-link split (compile vs link). Tested with make check RUNTESTFLAGS="--target_board=hip" TESTS="gdb.base/break.exp" -- compilation succeeds, zero hip_runtime.h errors.

I'm confused.   When running the normal gdb.rocm/ tests, I don't see "--rocm-path" passed to amdclang++.  Why is it needed here, and not there?

gdb/testsuite/lib/gdb.exp
```diff
@@ -6556,14 +6555,34 @@ proc gdb_compile {source dest type options} {
...
-	# the input files are .o files, like: hipcc foo.o -o foo ...
+	# the input files are .o files, like: amdclang++ foo.o -o foo ...
 	#
-	# HIPCC defaults to -O2, so add -O0 to early flags for the HIP
-	# language.  If "optimize" is also requested, another -O flag
-	# (e.g. -O2) will be added to the flags, overriding this -O0.
-	lappend new_options "early_flags=-O0\
+	# amdclang++ needs -x hip to treat .cpp/.c as HIP source when
+	# compiling, and --hip-link to link the HIP runtime when
+	# linking.  When linking only .o files, -x hip must NOT be
```
palves: As we've seen in the set_unbuffered_mode path below, the "only" in "only .o files" is not correct.  You can't have -x with any .o file.

Wouldn't it be better for the logic to force separate compilation be here instead of only in the set_unbuffered_mode path?

gdb/testsuite/lib/gdb.exp
```diff
@@ -6248,6 +6248,7 @@ proc gdb_compile {source dest type options} {
 	}
     }
 
+
```
palves: Spurious new empty line.

gdb/testsuite/lib/gdb.exp
spatrang: Done. Moved the pre-compile logic to the top of gdb_compile. For any HIP executable with any source input, gdb_compile now pre-compiles each .cpp/.c to its own .o first (with -x hip), then recurses with an all-.o source list. By the time the -x hip / --hip-link decision runs at the bottom, $source for type == "executable" always contains only .o files, so -x hip is safely omitted.

Also updated the comment to reflect this -- it now says "-x hip must NOT be passed when any .o input is present" and explains that the pre-compile step at the top ensures all executable-link invocations see only .o inputs.

lumachad:
Bringing this up here. @amd-shahab was asking questions on finding llvm-dwarfdump on a CI system. I think we should take the opportunity to make the tools search order correct in these changes (or a separate patch if desired).

See my previous comment here: https://github.com/ROCm/ROCgdb/pull/62#discussion_r3050310837.

I think we should honor PATH first. Then if ROCM_PATH is set, we go look for tools there. Then if everything fails, we search some hardcoded /opt/rocm path.

gdb/testsuite/lib/gdb.exp
```diff
@@ -6807,52 +6867,14 @@ proc gdb_compile {source dest type options} {
...
-		    }
-		    set options $stripped_options
+		    # For HIP, the generalized pre-compile step at
+		    # the top of gdb_compile has already turned any
+		    # source files into .o files.  Prepending the
+		    # unbuffered .o to the (all-.o) source list is
+		    # safe -- no mixed source/.o invocation happens.
+		    set source \
+			[linsert $source 0 \
+			     $gdb_saved_set_unbuffered_mode_obj($unbuf_obj_lang)]
```
palves: Would using the "ldflags=" non-hip path for hip too now work?

gdb/testsuite/lib/future.exp
```diff
@@ -121,20 +121,28 @@ proc gdb_find_rustc {} {
...
     if {![is_remote host]} {
-	set hipcc [lookfor_file $tool_root_dir hipcc]
-	if {$hipcc eq "" && [info exists ::env(ROCM_PATH)]} {
-	    set hipcc [lookfor_file $::env(ROCM_PATH)/bin hipcc]
+	set compiler [lookfor_file $tool_root_dir amdclang++]
+	if {$compiler eq "" && [info exists ::env(ROCM_PATH)]} {
+	    set compiler [lookfor_file $::env(ROCM_PATH)/llvm/bin amdclang++]
+	}
+	if {$compiler eq ""} {
+	    set compiler [lookfor_file /opt/rocm/llvm/bin amdclang++]
```
lancesix: Should be `/opt/rocmt/lib/llvm/bin`.

gdb/testsuite/lib/future.exp
```diff
@@ -121,20 +121,28 @@ proc gdb_find_rustc {} {
...
     } else {
-	set hipcc ""
+	set compiler ""
     }
-    return $hipcc
+    return $compiler
+}
+
+# Backwards compatibility alias.
+proc gdb_find_hipcc {} {
```
lancesix: If there are existing users of `find_hipcc` directly, would substuting `hipcc` with `amdclang++` be appropriate?  Doesn't `amdclang++` require extra flags (such as `-x hip`) to be considered a drop-in replacement of `hipcc`?

palves:
> Bringing this up here. @amd-shahab was asking questions on finding llvm-dwarfdump on a CI system. I think we should take the opportunity to make the tools search order correct in these changes (or a separate patch if desired).
> 
> See my previous comment here: [#62 (comment)](https://github.com/ROCm/ROCgdb/pull/62#discussion_r3050310837).
> 
> I think we should honor PATH first. Then if ROCM_PATH is set, we go look for tools there. Then if everything fails, we search some hardcoded /opt/rocm path.

I agree we should look into this, though I'd prefer if any changes to the search path order was done in a separate patch.   We have a little mess here, with search orders different for different tools.  E.g., "${rocm_path}/llvm/bin/clang-offload-bundler" in rocm.exp (which now misses the leading lib/), Shahab's patch that searches PATH first, the searches for all the different tools lib/future.exp which search PATH last.

lumachad:
> > Bringing this up here. @amd-shahab was asking questions on finding llvm-dwarfdump on a CI system. I think we should take the opportunity to make the tools search order correct in these changes (or a separate patch if desired).
> > See my previous comment here: [#62 (comment)](https://github.com/ROCm/ROCgdb/pull/62#discussion_r3050310837).
> > I think we should honor PATH first. Then if ROCM_PATH is set, we go look for tools there. Then if everything fails, we search some hardcoded /opt/rocm path.
> 
> I agree we should look into this, though I'd prefer if any changes to the search path order was done in a separate patch. We have a little mess here, with search orders different for different tools. E.g., "${rocm_path}/llvm/bin/clang-offload-bundler" in rocm.exp (which now misses the leading lib/), Shahab's patch that searches PATH first, the searches for all the different tools lib/future.exp which search PATH last.

Agreed. We need a separate change to clean this all up to use a single approach throughout.

gdb/testsuite/lib/gdb.exp
spatrang: Yes, that works.

gdb/testsuite/lib/gdb.exp
```diff
@@ -6248,6 +6248,53 @@ proc gdb_compile {source dest type options} {
...
 
+    # For HIP executables, amdclang++ needs -x hip to treat .cpp/.c as
+    # HIP source, but -x hip applied to a .o file makes amdclang++ try
+    # to parse the object as HIP source, which fails.  Since gdb_compile
+    # may later inject additional .o inputs (e.g. set_unbuffered_mode.o),
+    # we cannot safely mix any source files with any .o inputs in a
+    # single amdclang++ invocation.  To avoid that, pre-compile each
+    # source file to its own .o first (with -x hip), then recurse with
+    # an all-.o source list.  The recursive call will not add -x hip
+    # and the final link will work correctly.
```
lancesix: Couldn't we prefix the `.o` with `-x none`?

i.e. `amdclang++ -x hip foo.c -x none bar.o -x hip baz.c` to treat foo.c and baz.c as hip and let the compiler deduce what bar.o is.

Avoids having to have one invocation per input.

gdb/testsuite/lib/gdb.exp
```diff
@@ -6556,14 +6601,48 @@ proc gdb_compile {source dest type options} {
...
 	#
-	# HIPCC defaults to -O2, so add -O0 to early flags for the HIP
-	# language.  If "optimize" is also requested, another -O flag
-	# (e.g. -O2) will be added to the flags, overriding this -O0.
-	lappend new_options "early_flags=-O0\
+	# amdclang++ needs -x hip to treat .cpp/.c as HIP source when
+	# compiling, and --hip-link to link the HIP runtime when
+	# linking.  -x hip must NOT be passed when any .o input is
+	# present, otherwise amdclang++ tries to parse the .o file
+	# as HIP source.  The executable-linking path above pre-
```
lancesix: See previous comment, one could prefix the .o with `-x none` (and restore the language after the .o).

gdb/testsuite/lib/future.exp
lumachad: I'd do both old and new here, to keep things working for the old paths in case we ever want to run checks against an older tree. Then we should set a date by which we'll remove it. How does that sound?

lancesix: as far as I remember, `$ROCM_PREFIX/llvm` has been a symlink to `$ROCM_PREFIX/lib/llvm` for quite a long time, but I am happy having both if you feel we can come across configurations that do not have `lib/llvm` (arguably, having the compiler and other tools in the path would be the way to go anyway).

lumachad: If that's old news, I don't mind dropping it.

palves: I'd prefer leaving this falling back to /opt/rocm/ out of this patch.  That seems orthogonal to the amdclang++ change to me.

If we do the change as implemented, /opt/rocm overrides PATH, forcing you to set ROCM_PATH if you happen to also have a /opt/rocm dir you don't want to use.

gdb/testsuite/lib/gdb.exp
spatrang: Nice idea -- tested it and it works:

`amdclang++ --hip-link --rocm-path=... --offload-arch=gfx950 -x hip main.cpp -x none compute.o -o prog`

produces a working binary. I'll switch to this approach: prefix each input in the source list with -x hip (for .cpp/.c) or -x none (for .o/.a), drop -x hip from early_flags, and drop the pre-compile recursion + indexed-temp-file workaround. Single invocation, no temp collisions, closer to how hipcc worked. Will push the updated patch and re-test.

gdb/testsuite/lib/future.exp
spatrang: Good point -- removed the find_hipcc / gdb_find_hipcc aliases. amdclang++ isn't a drop-in for hipcc, so silently returning it would hide real breakage in out-of-tree callers.

gdb/testsuite/lib/future.exp
spatrang: Agreed with @palves -- the /opt/rocm fallback was added mid-review as a Copilot suggestion and is orthogonal to the hipcc → amdclang++ migration. Removed it.

Also updated $ROCM_PATH/llvm/bin → $ROCM_PATH/lib/llvm/bin per @lancesix -- that's where the actual binary lives; llvm is typically just a symlink. Users who want to rely on PATH or a non-standard layout can just set ROCM_PATH, use HIP_COMPILER_FOR_TARGET=/path/to/amdclang++, or have amdclang++ on PATH.

Search order is now:

$tool_root_dir (DejaGnu convention for locally-built tools)
$ROCM_PATH/lib/llvm/bin/amdclang++ (if ROCM_PATH is set)
Bare amdclang++ on PATH

gdb/testsuite/boards/hip.exp
```diff
@@ -149,10 +149,18 @@ proc skip_opencl_tests {} {
...
-proc allow_hipcc_tests {} {
+proc allow_hip_tests {} {
     return 0
 }
 
+# Backwards compatibility alias.  The testsuite previously used
+# allow_hipcc_tests when hipcc was the HIP compiler wrapper.  Keep
+# this alias so that any out-of-tree board files or scripts that
+# reference the old name continue to work.
+proc allow_hipcc_tests {} {
```
palves: I don't think we really need this.  Out-of-tree board files will probably need adjustment for other things anyhow.

gdb/testsuite/boards/hip.exp
```diff
@@ -290,7 +302,13 @@ proc gdb_compile {source dest type options} {
 	#    '-mllvm -amdgpu-spill-cfi-saved-regs'
 	#    [-Wunused-command-line-argument]
 	#
+	# Pass --rocm-path explicitly so amdclang++ can locate the HIP
+	# headers and device libraries when it's invoked from the board
+	# via PATH rather than an absolute path.  rocm_path comes from
+	# rocm.exp and defaults to /opt/rocm if ROCM_PATH is not set.
+	global rocm_path
```
palves: This sounds really weird to me.  

Is this documented behavior?  So regular compilation from the command line with plain "amdclang++" does not work unless you pass --rocm-path?

I find this super strange.  Clang is able to find e.g. libstdc++/gcc headers, the Microsoft SDK and Visual Studio headers etc. automatically, but not the rocm headers?  Huh.

I looked a little at the docs, and here:

https://rocm.docs.amd.com/projects/llvm-project/en/latest/reference/rocmcc.html

It says:

"Finding a HIP installation

First looks for HIP under the same parent directory as its own LLVM directory **and then falls back on /opt/rocm**. Users can use the --rocm-path option to instruct the compiler to use HIP from the specified ROCm installation."

So why didn't the /opt/rocm fallback case work?  From that, it looks like we hit an amdclang++ bug.  Or a documentation bug.

We probably still want to --rocm-path so that we respect the ROCM_PATH variable.  But still, that comment indicates something weird to me.

gdb/testsuite/lib/gdb.exp
palves: This is marked resolved, but I still see the pre-compile recursion workaround?

(unresolving)

gdb/testsuite/lib/rocm.exp
```diff
@@ -150,6 +150,14 @@ gdb_caching_proc allow_hipcc_tests {} {
     return 1
 }
 
+# Backwards compatibility alias.  The testsuite previously used
+# allow_hipcc_tests when the legacy hipcc wrapper was the HIP
+# compiler.  Keep this alias so that any out-of-tree test files
+# or scripts referencing the old name continue to work.
+gdb_caching_proc allow_hipcc_tests {} {
```
palves: I don't think we really need this. Out-of-tree board files will probably need adjustment for other things anyhow.  Was there a particular script of board file you were thinking of?

gdb/testsuite/lib/gdb.exp
spatrang: Addressed in 333c780d. The pre-compile recursion plus indexed-temp-file workaround is gone; gdb_compile now tags each input on the command line with -x hip (sources) or -x none (.o/.a) just before the target_compile call, per @lancesix's suggestion. A single amdclang++ invocation now correctly handles a mix of sources and object inputs, including the set_unbuffered_mode.o/hip-driver.o appended via ldflags.

`amdclang++ --rocm-path=... --hip-link break0.o break1.o -x none hip-driver.o -lm -o break`

gdb/testsuite/boards/hip.exp
spatrang: Tested empirically — with a fully clean environment, amdclang++ does auto-discover HIP correctly:
 
```
 $ env -i PATH=/usr/bin:/bin /usr/bin/amdclang++ -x hip -E -v - < /dev/null 2>&1 | grep "Found HIP"
Found HIP installation: /opt/rocm-7.3.0/lib/llvm/bin/../../.., version 7.2.0
```

That matches the documented behavior (parent of its own LLVM dir, then /opt/rocm fallback). The old comment was misleading; --rocm-path was never strictly necessary, just convenient for honoring an explicit ROCM_PATH.

Changed in 333c780d to only pass --rocm-path when ROCM_PATH is set in the environment; otherwise we defer to amdclang++'s built-in discovery.

gdb/testsuite/lib/rocm.exp
spatrang: Removed in 333c780d.

gdb/testsuite/boards/hip.exp
```diff
@@ -271,10 +271,14 @@ proc gdb_compile {source dest type options} {
 
 	# We need:
 	#
-	# . explicit -O0, because hipcc optimizes by default.
+	# . explicit -O0, because the legacy hipcc wrapper optimized by
+	#   default.  amdclang++ does not, but we keep -O0 explicit for
+	#   clarity and to stay safe if defaults ever change.
```
palves: Let's just remove the -O0.  There is no point in staying safe like that, it's not going to change without anyone noticing.

gdb/testsuite/boards/hip.exp
```diff
@@ -271,10 +271,14 @@ proc gdb_compile {source dest type options} {
...
 	#
-	# . -std=gnu++11 because hipcc defaults to -std=c++11, and
-	#   some testcases assume GNU extensions.
+	# . -std=gnu++11 because the legacy hipcc wrapper defaulted to
+	#   -std=c++11, and some testcases assume GNU extensions.
+	#   amdclang++ defaults to a later standard but we keep this
+	#   for consistency.
```
palves: Let's tweak the comment to talk about current state without comparing to old hipcc.  There is point in talking about the legacy hipcc wrapper, and there is no need to keep consistency with it.

gdb/testsuite/lib/gdb.exp
```diff
@@ -6768,14 +6766,37 @@ proc gdb_compile {source dest type options} {
...
 	# These happen when we're compiling an executable _and_ all
-	# the input files are .o files, like: hipcc foo.o -o foo ...
+	# the input files are .o files, like: amdclang++ foo.o -o foo ...
 	#
-	# HIPCC defaults to -O2, so add -O0 to early flags for the HIP
-	# language.  If "optimize" is also requested, another -O flag
-	# (e.g. -O2) will be added to the flags, overriding this -O0.
-	lappend new_options "early_flags=-O0\
+	# amdclang++ does not infer the input language from the file
+	# extension the way hipcc did.  Rather than passing a single
```
palves: Drop this "the way hipcc did" remark.

palves (commented):
I re-read this and it looks mostly good to me.  

Let's just remove comparisons to leggacy hipcc in code coments, we're switching away from it completely, it just ends up being useless filler.

At this point, I'd prefer if you just squashed this all into a single commit, as there is a lot of back and forth in the commits while the end result is surprisingly smaller than it would seem.  That should make you update the commit log to reflect current reality too, as the PR's text is out of date.

gdb/testsuite/boards/hip.exp
spatrang: Addressed in an update I'm about to push: removed `-O0` from the board's HIP flags. For consistency I also drop the explicit `-O0` from the native HIP path in `lib/gdb.exp` — amdclang++ defaults to `-O0`, so this is behavior-preserving, and an explicit `optimize` option still overrides it.

gdb/testsuite/boards/hip.exp
spatrang: Addressed in the upcoming update: reworded the comment to describe only the current amdclang++ flags, dropping the legacy-hipcc comparisons. The `-std=gnu++11` note now just states why it is needed (some testcases rely on GNU extensions).

gdb/testsuite/lib/gdb.exp
spatrang: Dropped in the upcoming update — along with the same "the way hipcc did" phrasing in the second occurrence (the per-input tagging comment) for consistency.

spatrang:
Thanks — addressing all of it in an update I'm about to push:

- Removed the legacy-hipcc comparisons from the code comments (a repo-wide sweep is now clean). I also removed the `hipcc` deprecation / equivalent-command block from `gdb.texinfo` and reworded the stale "Windows HIPCC" comment in `lib/gdb.exp`. (The only remaining `hipcc` token is a historical ABI reference in `gdb/amdgpu-tdep.c`, which is outside this patch.)
- Squashed everything into a single commit and rewrote the commit log to describe the final state (compiler discovery, the `-x hip` / `-x none` per-input tagging, `--hip-link`, conditional `--rocm-path`, the `-O0` removal, and the `allow_hipcc_tests` -> `allow_hip_tests` rename).
- Will refresh the PR description to match.

The branch is also rebased onto current `amd-staging`; while doing so I picked up the new `hip-lang-detect.exp`, `hip-catch-errors.exp` and `interrupt-twice.exp` tests and renamed their `require allow_hipcc_tests` to `allow_hip_tests`.

palves:
> I also removed the `hipcc` deprecation / equivalent-command block from `gdb.texinfo` 

Well, I think that change went too far.  I had only asked to remove the comments parts.   hipcc still exists and is widely used by users, so I think that part of the manual should have stayed.

palves:
What is that "--genco" commit about?  If you do fixup commits, please make sure to add some rationale in their commit logs.   That was needed because of these two tests, I guess:

testsuite/gdb.rocm/code-object-load-while-breakpoint-hit.exp:   { debug hip additional_flags=--genco additional_flags=-DDEVICE } ] != "" } {
testsuite/gdb.rocm/snapshot-objfile-on-load.exp:        { debug hip additional_flags=--genco } ] != "" } {

But why the mapping instead of just fixing the tests?

spatrang:
> Well, I think that change went too far. I had only asked to remove the comments parts. hipcc still exists and is widely used by users, so I think that part of the manual should have stayed.

Agreed, that was overreach on my part. I've restored the `hipcc` section in `gdb.texinfo` (the deprecation note plus the equivalent `hipcc` command); only the legacy-hipcc comparisons in *code comments* are dropped now.

> What is that "--genco" commit about? If you do fixup commits, please make sure to add some rationale in their commit logs. [...] But why the mapping instead of just fixing the tests?

Good point on both counts. I've dropped the `gdb_compile` mapping and instead changed the two tests directly to use `--cuda-device-only` (the amdclang++ equivalent of hipcc's `--genco`):

- `gdb.rocm/code-object-load-while-breakpoint-hit.exp`
- `gdb.rocm/snapshot-objfile-on-load.exp`

I also folded everything back into the single commit (no more fixup commit), and the commit log now documents the `--genco` -> `--cuda-device-only` change.

palves (commented):
The patch itself LGTM now.  Just a few comments on the commit log:

- The first paragraph has a few nits to pick:

> The GDB testsuite used hipcc for HIP compiler discovery, naming, and
> documentation, even though the underlying compiler is amdclang++.
> Complete the switch to amdclang++ throughout the testsuite.

"The GDB testsuite used hipcc" => "uses hipcc"

The "the underlying compiler is amdclang++" isn't what motivates this.  The real reason is that hipcc is deprecated and is going away at some point.

And then "complete the switch" implies the switch had been started already.  It hasn't, this patch starts and finishes it.

-  It doesn't mention the -fdiagnostics-color change in the "Compile and link flags" section.  It should mention that amdclang++ is clang-based and accepts the flag. 

- "Add HIP_COMPILER_FOR_TARGET" is not strictly correct, it's more rename "HIPCC_FOR_TARGET to HIP_COMPILER_FOR_TARGET".

- FYI, the PR text and the commit log are slightly out of sync, which is a bit confusing and made me wonder which one is up to date.  Since with "squash and merge" the PR text doesn't make it to git (the git log history is the canonical one), then feel free to simplify the PR text to give the short intro and then say to look at the individual commit log for details.  Alternatively, it's fine to have the info in both places, though I'd suggest using the gh tool to easily and automatically sync them.

lancesix:
Sorry I have not had much time to look at this PR.  I expect a fair amount of this can / should be upstream, right?  Having this only downstream would cause some divergence we'll need to fix soon.

palves:
> Sorry I have not had much time to look at this PR. I expect a fair amount of this can / should be upstream, right? Having this only downstream would cause some divergence we'll need to fix soon.

Agreed, you had already mentioned this earlier, even.  :-)

One issue is that upstream does not have any of the Windows bits in the testsuite (which needed adjustment in this patch), and it has fewer tests to adjust.  This is why I kept pushing for finishing up the patch downstream first.  

I'd vote for landing the patch downstream first, and then also send it upstream shortly after.  The upstream version will end up slightly simplified in some parts.

lancesix:
> 
> Agreed, you had already mentioned this earlier, even. :-)
> 
oups, forgot I already mentioned that :sweat_smile: 
> 
> I'd vote for landing the patch downstream first, and then also send it upstream shortly after. The upstream version will end up slightly simplified in some parts.

cool, works for me.

spatrang:
> Just a few comments on the commit log

Thanks - addressed all of these in the latest force-push:

- Reworded the opening paragraph: "uses hipcc", motivation is now that hipcc is deprecated and going away, and dropped "complete the switch" since this patch both starts and finishes it.
- Added a note that amdclang++ is clang-based and accepts `-fdiagnostics-color`, so we no longer skip it for HIP (keeps color escapes out of gdb.log).
- Reworded to "rename HIPCC_FOR_TARGET to HIP_COMPILER_FOR_TARGET" instead of "add".
- Synced the PR description with the commit log.

palves (approved):
Thanks, LGTM.  Let's merge this (with rebase and merge).

Next step is upstreaming this.  Let me/us know if you'd like to work on that.

lumachad:
Please don't merge. These tests are failing for both gcc and llvm:

```
2026-06-18 18:25:09,111 - INFO:        gdb.dwarf2/ada-thick-pointer.exp
2026-06-18 18:25:09,111 - INFO:        gdb.rocm/coop-group-grid-sync.exp
```

lumachad:
For the group sync one...

```
ERROR: -------------------------------------------
ERROR: in testcase /__w/ROCgdb/ROCgdb/build/tests/rocgdb/gdb/testsuite/gdb.rocm/coop-group-grid-sync.exp
ERROR:  invalid command name "allow_hipcc_tests"
ERROR:  tcl error code TCL LOOKUP COMMAND allow_hipcc_tests
ERROR:  tcl error info:
invalid command name "allow_hipcc_tests"
    while executing
"::gdb_tcl_unknown allow_hipcc_tests"
    ("uplevel" body line 1)
    invoked from within
"uplevel 1 ::gdb_tcl_unknown $args"
    (procedure "::unknown" line 5)
    invoked from within
"allow_hipcc_tests"
    ("uplevel" body line 1)
    invoked from within
"uplevel 1 $fn"
    (procedure "require" line 11)
    invoked from within
"require allow_hipcc_tests"
    (file "/__w/ROCgdb/ROCgdb/build/tests/rocgdb/gdb/testsuite/gdb.rocm/coop-group-grid-sync.exp" line 34)
    invoked from within
"source /__w/ROCgdb/ROCgdb/build/tests/rocgdb/gdb/testsuite/gdb.rocm/coop-group-grid-sync.exp"
    ("uplevel" body line 1)
    invoked from within
"uplevel #0 source /__w/ROCgdb/ROCgdb/build/tests/rocgdb/gdb/testsuite/gdb.rocm/coop-group-grid-sync.exp"
    invoked from within
"catch "uplevel #0 source $test_file_name" msg"
```

lumachad:
Don't think there's actually anything wrong with gdb.dwarf2/ada-thick-pointer.exp. It's just confusion from the CI script due to the group sync testcase errors trumpling over the gdb.dwarf2/ada-thick-pointer.exp testcase output.

spatrang:
Rebased onto the latest `amd-staging` and force-pushed.

What changed since the last push:
- Rebased onto current `amd-staging`, which pulled in the new `gdb.rocm/coop-group-grid-sync.exp` test (and the `supports_cooperative_groups` helper it needs).
- Renamed that test's `require allow_hipcc_tests` to `require allow_hip_tests`, matching the rest of the testsuite. This was the actual cause of the red CI (`invalid command name "allow_hipcc_tests"`); `gdb.dwarf2/ada-thick-pointer.exp` was only collateral noise from that test's error trampling the log, so it clears too.

I built rocgdb from this branch and ran `gdb.rocm/*.exp` + `gdb.dwarf2/*.exp` via `.github/scripts/test_rocgdb.py` (GCC and LLVM) against that tree:
- All `gdb.rocm` tests pass (incl. `coop-group-grid-sync`, `hip-lang-detect`, `interrupt-twice`, and the code-object tests).
- The only remaining unexpected failures are in `gdb.dwarf2` (`debug-names-non-ascending-cu`, `macro-source-path`, `dw2-ranges`). These are unrelated to this patch — it touches no `gdb.dwarf2` or GDB source files — and reproduce on `amd-staging` independently.

PR #64, opened 2026-04-09
[cherry-pick-amd-staging-rocgdb-16] gdb/dwarf2/expr: Cache the register reads

lancesix:
A recent change in the ROCm compiler started to show some timeout in the gdb.rocm/deep-stack.exp testcase.  In that test, we have a 12 frames deep stack, and we print a stack trace.  With the new compiler output, building this stacktrace would take about 190 seconds... Yes, about 3 minutes to unwind 12 frames.

Looking at the unwind information, we now see this new pattern for unwinding vector registers:

    VGPR40=[
      DW_OP_regx VGPR40
      DW_OP_swap
      DW_OP_LLVM_user
      DW_OP_LLVM_offset_uconst 0x0
      DW_OP_LLVM_user DW_OP_LLVM_call_frame_entry_reg EXEC
      DW_OP_deref_size 0x8
      DW_OP_LLVM_user DW_OP_LLVM_select_bit_piece 0x20 0x40
    ]

This new construct uses DW_OP_LLVM_select_bit_piece, which allows us to build a composite which has one piece per lane of the vector register, each piece referencing either
- the register itself (not touched because the lane was not active, so no need to spill), or
- a saved copy of the register to memory (at CFA).

The issue turned out to be that when building the composite, we ended up creating up to 64 pieces (one per lane), each can be a dwarf_register instance.  And even if all of those reference the same register, they do not share anything more than the regnum.

When we evaluate something like dwarf_composite::is_optimized_out, we end-up calling dwarf_register::is_optimized_out for each piece in sequence.  Each one of those calls will call dwarf_register::read and eventually frame_register_wind, i.e. we fully unwind the register once for each lane.  If the next frame, and the next, and the next use the same pattern, we end-up with an impossible complexity of O(lanes ** depth), which explains how we can end-up with 3 minutes to do a 12 frames unwinding.

This pattern does not only happen when calling value::optimized_out, this also happens when calling value->entirely_available for example, making things very inefficient..

To solve this, this patch makes two things:
- Make it so when we use DW_OP_LLVM_select_bit_piece, all the dwarf_register in the resulting composite share a storage,
- Make that storage cache the result of calling frame_register_unwind, so we do not have to repeat this for every piece.

The cache is cleared any time we call dwarf_register::write, or if we detect that it might be outdated (target_changed or memory_changed got called since we cached the data).

With this, doing the backtrace from the test now takes slightly more than .5 seconds, which is back to acceptable territory.

    (gdb) bt
    2026-04-02 17:40:05.195 - command started
    #0  base_case () at /home/.../gdb.rocm/deep-stack.cpp:40
    #1  0x00007ffff637649c in deep<0u> () at /home/.../gdb.rocm/deep-stack.cpp:54
    #2  0x00007ffff637655c in deep<1u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    #3  0x00007ffff637661c in deep<2u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    #4  0x00007ffff63766e4 in deep<3u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    #5  0x00007ffff63767a4 in deep<4u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    #6  0x00007ffff6376864 in deep<5u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    #7  0x00007ffff6376924 in deep<6u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    #8  0x00007ffff63769e4 in deep<7u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    #9  0x00007ffff6376aa4 in deep<8u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    #10 0x00007ffff6376b64 in deep<9u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    #11 0x00007ffff6376c24 in deep<10u> () at /home/.../gdb.rocm/deep-stack.cpp:47
    #12 0x00007ffff6376d40 in hip_deep () at /home/.../gdb.rocm/deep-stack.cpp:60
    2026-04-02 17:40:05.753 - command finished
    Command execution time: 0.558060 (cpu), 0.558099 (wall)

Change-Id: I976af7fc7d6a946967d3aa986bce5b990891da20
(cherry picked from commit f204e6a13acc73c30ebe029dfa81d0e95932bbd8)

### DIFF
--- a/gdb/dwarf2/expr.c
+++ b/gdb/dwarf2/expr.c
@@ -32,11 +32,20 @@
 #include "dwarf2/read.h"
 #include "frame.h"
 #include "gdbsupport/underlying.h"
+#include "gdbsupport/refcounted-object.h"
 #include "gdbarch.h"
 #include "gdbthread.h"
 #include "inferior.h"
 #include "objfiles.h"
 #include "observable.h"
+
+using context_generation_t = ULONGEST;
+
+/* A monotically increasing counter modified every time any inferior memory
+   is touched or any register in any target is modified.  This can be used to
+   know if state (register value) cached by the evaluator might be stale.  */
+
+static context_generation_t context_generation = 0;
 
 /* This holds gdbarch-specific types used by the DWARF expression
    evaluator.  See comments in execute_stack_op.  */
@@ -1243,13 +1252,12 @@ public:
 		  bool on_entry = false, LONGEST offset = 0,
 		  LONGEST bit_suboffset = 0)
     : dwarf_location (arch, offset, bit_suboffset),
-      m_regnum (regnum), m_on_entry (on_entry)
+      m_storage {register_storage_ptr::new_reference
+		   (new register_storage (regnum, on_entry))}
   {}
 
   dwarf_register (const dwarf_register &reg)
-    : dwarf_location (reg),
-      m_regnum (reg.m_regnum),
-      m_on_entry (reg.m_on_entry)
+    : dwarf_location (reg), m_storage (reg.m_storage)
   {}
 
   std::shared_ptr<dwarf_entry> clone () const override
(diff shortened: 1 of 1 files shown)


palves (approved):

lumachad (approved):
OK if CI is good.

PR #65, opened 2026-04-10
Fix ROCgdb CI trigger

lumachad:
- Pick up TheRock's revision that contains a fix to allow ROCgdb to be built
from an external source directory.

- Pass Python options to TheRock.

- Always use the latest manylinux image for the builds.

### DIFF
--- a/.github/workflows/therock-ci-linux.yml
+++ b/.github/workflows/therock-ci-linux.yml
@@ -14,7 +14,7 @@ permissions:
   contents: read
 
 env:
-  THEROCK_COMMIT_REF: c9529ebdedc2e52c7b4dc26d209725a31f616ac7 # 2026-03-13 commit
+  THEROCK_COMMIT_REF: 65a2a9cd5895a9cdb2a1352767377f35e70254d1 # 14/04/2026
 
 jobs:
   therock-build-linux:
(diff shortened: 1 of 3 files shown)


.github/workflows/therock-ci-linux.yml
```diff
@@ -14,7 +14,7 @@ permissions:
   contents: read
 
 env:
-  THEROCK_COMMIT_REF: c9529ebdedc2e52c7b4dc26d209725a31f616ac7 # 2026-03-13 commit
+  THEROCK_COMMIT_REF: 65a2a9cd5895a9cdb2a1352767377f35e70254d1 # 14/04/2026
```
lancesix, line 17: For my own education, could that be main?

Does this mean we need to update this ref every time we rely on something "new" in dbgapi?

lumachad: Since we own our workflows, it is up to us. Using top-of-tree main TheRock means we always use the latest, broken or not. Having a pinned revision helps with that, but indeed we need to keep updating things from time to time.

If we have a hard dependency on new code in dbgapi, we need to be mindful of that.

lancesix: Ok. For now, we can roll with that and stay with a stable version.  If needed, we can revisit that in the future.

lancesix (approved):
LGTM, thanks.

PR #66, opened 2026-04-10
gdb/dwarf2/expr: Fix cherrx-pick of "Cache the register reads"

lancesix:
The cherry-pick of "Cache the register reads" into the amd-staging-rocgdb-16 was incorrect, and was not detected by CI.

This patch fixes things.

Tested locally (build and gdb.rocm/*.exp + gdb.dwarf2/*.exp).

Change-Id: I054f684e5c17038252dd49980607b39992f21a2a


lumachad (approved):
CI is not trustworthy in this run. If you've validated it locally, that's fine.

PR #76, opened 2026-04-14
[github] Handle amd-staging-rocgdb-* branches in pre-commit

lumachad:
As the subject says, attempt to handle amd-staging-rocgdb-* branches in the pre-commit workflow, such that we don't have to customize things for those branches.

The idea is that we can just copy these workflows to a new amd-staging-rocgdb-* branch and it should work.


lancesix (approved):
LGTM, thanks.

PR #85, opened 2026-04-16
[testsuite] Fix failures in gdb.rocm/load-core-remote-system.exp

lumachad:
I spotted gdb.rocm/load-core-remote-system.exp failing when the agent ID in a system isn't 1.

FAIL: gdb.rocm/load-core-remote-system.exp: sysroot: info threads
FAIL: gdb.rocm/load-core-remote-system.exp: solib-search-path: info threads

This is due to the testcase's hardcoded Target ID. Fix this by expecting the correct pattern, but ignoring the specific numbers.


gdb/testsuite/gdb.rocm/load-core-remote-system.exp
```diff
@@ -91,7 +91,7 @@ proc do_test {} {
 	gdb_test_no_output "set sysroot $sysroot_base" "set sysroot"
 	gdb_test "core-file $corefile_path" "Core was generated by.*" "open corefile"
 	gdb_test "info threads" \
-	    "AMDGPU Wave 1:1:1:1 \\(0,0,0\\)/0${ws}\"do_an_addition\"${ws}do_an_addition.*"
+	    "AMDGPU Wave $::decimal:$::decimal:$::decimal:$::decimal \\($::decimal,$::decimal,$::decimal\\)/$::decimal${ws}\"do_an_addition\"${ws}do_an_addition.*"
```
lancesix: I'd say this becomes too lay.

The `:1 (0, 0, 0)/0` should always be true.  we always see the first wave of the only workgroup (so first wave of the dispatch), that would not change even for multi-GPU configs.

gdb/testsuite/gdb.rocm/load-core-remote-system.exp
```diff
@@ -100,7 +100,7 @@ proc do_test {} {
 	gdb_test_no_output "set solib-search-path $solib_search_path" "set solib-search-path"
 	gdb_test "core-file $corefile_path" "Core was generated by.*" "open corefile"
 	gdb_test "info threads" \
-	    "AMDGPU Wave 1:1:1:1 \\(0,0,0\\)/0${ws}\"do_an_addition\"${ws}do_an_addition.*"
+	    "AMDGPU Wave $::decimal:$::decimal:$::decimal:$::decimal \\($::decimal,$::decimal,$::decimal\\)/$::decimal${ws}\"do_an_addition\"${ws}do_an_addition.*"
```
lancesix: Same, keep `:1 (0,0,0)/0`

lumachad: And just to make sure, that would be the case regardless of concurrent use of the GPU?

gdb/testsuite/gdb.rocm/load-core-remote-system.exp
lumachad: Fixed now.

lancesix (approved):
LGTM, thanks.

gdb/testsuite/gdb.rocm/load-core-remote-system.exp
lancesix: Yes.

The numbering is:

```
gpu_id:queue_id:dispatch_id:wave_id (wg_x,wg_y,wg_z)/wave_in_wg
```

For this case, everything within the dispatch should always be the same, only gpu_id / queue_id / dispatch_id are subject to change.

PR #86, opened 2026-04-16
Use a fixed manylinux container SHA

lumachad:
Use of the latest image has proven to be unreliable at the moment. For now switch to using a fixed SHA that we manually update.


lancesix (approved):
LGTM, thanks.

PR #90, opened 2026-04-16
Add PR title prefixer workflow

lumachad:
Prefixes PR titles with the name of the branch being targeted.

### DIFF
--- a/.github/workflows/pr_prefixer.yml
+++ b/.github/workflows/pr_prefixer.yml
@@ -1,0 +1,50 @@
+name: Add Branch Prefix to PR Title
+
+on:
+  pull_request:
+    types: [opened, reopened]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  add-branch-prefix:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add branch prefix to PR title
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const baseBranch = context.payload.pull_request.base.ref;
+            const currentTitle = context.payload.pull_request.title;
+
+            // Strictly match:
+            // - amd-staging-rocgdb-<version>
+            // - amd-temp-<name>
+            const validBranchPattern = /^(amd-staging-rocgdb-\d+|amd-temp-[a-zA-Z0-9-]+)$/;
+            if (!validBranchPattern.test(baseBranch)) {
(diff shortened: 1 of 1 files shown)


lumachad:
@lancesix I think I'm happy with this prefixer. I like the idea. Should we go with this one or check with folks first?

lancesix:
> @lancesix I think I'm happy with this prefixer. I like the idea. Should we go with this one or check with folks first?

I think we can go with this one.  I'll get to review that "properly" when I have a chance (probably tomorrow).

.github/workflows/pr_prefixer.yml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+      - name: Add branch prefix to PR title
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const baseBranch = context.payload.pull_request.base.ref;
+            const currentTitle = context.payload.pull_request.title;
+
+            // Strictly match amd-staging or amd-staging-rocgdb-<version>
+            const validBranchPattern = /^(amd-staging|amd-staging-rocgdb-\d+)$/;
```
lancesix: For now, we also have the `amd-temp-windows` branch used for windows builds.  We might want to match that one as well.

lumachad: Meaning we'll also want to prefix those PR's with amd-temp-windows?

lancesix: Yes, I expect so.

My way of thinking of this is:
- Any PR not targeting the default branch (amd-staging) should have a `[$BRANCH_NAME]` prefix (makes it easier to have some context of where this needs to be applied)
- I expect we'll be submitting patches to `amd-staging(-rocgdb-[0-9]{2})?` and `amd-temp-windows` (could be `amd-temp-([a-zA-Z0-9-]+` to allow for future such topic branch if needed)

lumachad: Understood. Should we add the prefix to amd-staging anyway?

lancesix: I would not, because this is the implied default, and makes the UI (list of PRs) lighter if most PR do not have it, but I think this is just a personal taste.  To me the `[...]` prefix is just to grab the attention of the reviewer that this is not intended for the default branch.

lumachad: Fair enough.

lancesix (approved):

PR #91, opened 2026-04-17
New test gdb.rocm/write-all-registers.exp

palves:
This adds a test that writes to all GPU registers, confirming that:

- For registers with non-read-only bits, written values are really
written and can be read back.
- Writing to read-only registers is ignored.

Tested on Linux, on gfx90a, gfx942, gfx1030, and gfx1201.
Tested on Windows, on gfx1201.

Change-Id: I76599cb02bd0cdbe504398763d48a8112ad21bab

### DIFF
--- a/gdb/testsuite/gdb.rocm/write-all-registers.exp
+++ b/gdb/testsuite/gdb.rocm/write-all-registers.exp
@@ -1,0 +1,265 @@
+# Copyright 2026 Free Software Foundation, Inc.
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test writing to all registers in the GPU, and confirming that:
+#  - For registers with non-read-only bits, written values are really
+#    written and can be read back.
+#  - Writing to read-only registers is ignored.
+
+load_lib rocm.exp
+
+require allow_hipcc_tests
+
+standard_testfile .cpp
+
+if {[build_executable "failed to prepare" $testfile $srcfile {debug hip}]} {
+    return
+}
+
+# Make an array value for a vector register, with COUNT number of
+# values starting at START.
+proc make_vector_val {start count} {
+    set end [expr {$start + $count}]
+    set val "{"
+    for {set i $start} {$i < $end} {incr i} {
+	if {$i == $start} {
+	    append val "$i"
+	} else {
+	    append val ", $i"
+	}
+    }
+    append val "}"
+    return $val
+}
+
+# Return register REG's size in bytes.
+proc register_size {reg} {
+    return [get_valueof "/d" "sizeof(\$$reg)" INVALID "get size"]
+}
+
+# Save the original value of REG, run BODY, and then restore the
+# original value of REG.  The register's original value is saved in
+# ORG_VALUE_NAME.  Saving/restoring is important because modifying
+# some registers has side effects on what values can be written on
+# others.
+proc save_restore_register { reg org_value_name body } {
+    upvar 1 $org_value_name org_value
+    set org_value [get_valueof "/x" "\$$reg" INVALID "get original value"]
+
+    set code [catch {uplevel 1 $body} result]
+
+    gdb_test "p/x \$$reg = $org_value" \
+	" = $org_value" \
+	"restore original value"
+
+    if {$code == 1} {
+	global errorInfo errorCode
+	return -code $code -errorinfo $errorInfo -errorcode $errorCode $result
+    } else {
+	return -code $code $result
+    }
+}
+
+# Registers that are fully read-only from the client's perspective.
+proc is_client_readonly {reg} {
+    if {[regexp "^ttmp\[0-9\]+" $reg]} {
+	return 1
+    }
+    return [expr {$reg in {
+	excp_flag_priv
+	flat_scratch
+	scratch_base
+	state_priv
+	status
+	trap_ctrl
+	trapsts
+    }}]
+}
+
+# Registers that have some (but not all) bits that are read-only.
+proc is_some_readonly {reg} {
+    # The single list contains all the special registers from all
+    # supported architectures, sorted in alphabetical order.
+    return [expr {$reg in {
+	excp_flag_user
+	mode
+    }}]
+}
+
+# True if REG is a vector register.
+proc is_vector_register {reg} {
+    return [regexp "^(v|a)\[0-9\]+" $reg]
+}
+
+# Write VALUE to register REG, and then read it back to confirm the
+# write worked.  FMT is the format to pass to the print command.
+# TEST_SUFFIX is appended to the test messages.
+proc write_read_register_fmt {fmt reg value test_suffix} {
+    set value_re [string_to_regexp $value]
+    gdb_test "p/$fmt \$$reg = $value" \
+	" = $value_re" \
+	"write value $test_suffix"
+    gdb_test "p/$fmt \$$reg" \
+	" = $value_re" \
+	"read value $test_suffix"
+}
+
+# Like write_read_register_fmt, for decimal values.
+proc write_read_register_dec {reg dec_value test_suffix} {
+    return [write_read_register_fmt "d" $reg $dec_value $test_suffix]
+}
+
+# Like write_read_register_fmt, for hexadecimal values.
+proc write_read_register_hex {reg hex_value test_suffix} {
+    return [write_read_register_fmt "x" $reg $hex_value $test_suffix]
+}
+
+proc do_test {} {
+    clean_restart
+    gdb_load $::binfile
+
+    with_rocm_gpu_lock {
+	if {![runto kern allow-pending message]} {
+	    return
+	}
+
+	# Collect the names of all registers.
+	set all_registers {}
+	set any "\[^\r\n\]+"
+	gdb_test_multiple "info all-registers" "" -lbl {
+	    -re "^info all-registers(?=\r\n)" {
+		exp_continue
+	    }
+	    -re "^\r\n(\[a-z0-9_\]+) ${any}(?=\r\n)" {
+		set reg "$expect_out(1,string)"
+		verbose -log "got $reg"
+		lappend all_registers $reg
+		exp_continue
+	    }
+	    -re "^\r\n$::gdb_prompt $" {
+		pass $gdb_test_name
+	    }
+	}
+
+	# Build values for the vector registers.
+	set vector_length \
+	    [get_valueof "/d" "sizeof(\$v0)/sizeof(\$v0\[0\])" INVALID \
+		"get number of elements of \$v0"]
+	set vec_val1 [make_vector_val 1 $vector_length]
+	set vec_val2 [make_vector_val 128 $vector_length]
+
+	# For each register, write to it, and confirm the value did
+	# change.  Write two different values, in case the register
+	# happened to already have the first value.
+	foreach_with_prefix reg $all_registers {
+	    if {[is_vector_register $reg]} {
+		write_read_register_dec $reg $vec_val1 "1"
+		write_read_register_dec $reg $vec_val2 "2"
+	    } elseif {$reg == "pc"} {
+		# Must be multiples of 4.
+		write_read_register_hex $reg 0x223344556600 "1"
+		write_read_register_hex $reg 0x778899aabb04 "2"
+	    } elseif {[is_client_readonly $reg] || [is_some_readonly $reg]} {
+		set reg_size [register_size $reg]
+		if {$reg_size == 4} {
+		    set all_ones_val 0xffffffff
+		} elseif {$reg_size == 8} {
+		    set all_ones_val 0xffffffffffffffff
+		} else {
+		    fail "unsupported register size"
+		    continue
+		}
+
+		save_restore_register $reg org_value {
+		    if {[is_client_readonly $reg]} {
+			gdb_test "p/x \$$reg = $all_ones_val" \
+			    " = $all_ones_val" \
+			    "try set all ones"
+			gdb_test "p \$$reg == $org_value"  \
+			    " = true" \
+			    "read-only register did not change 1"
+
+			gdb_test "p/x \$$reg = 0x0" " = 0x0" \
+			    "try set all zeroes"
+			gdb_test "p \$$reg == $org_value"  \
+			    " = true" \
+			    "read-only register did not change 2"
+		    } elseif {[is_some_readonly $reg]} {
+			# Find the set of read-only registers:
+			#  - write all 1s, collect mask of bits that stay at 0
+			#  - write all 0s, collect mask of bits that stay at 1
+			#  - OR the two masks
+			# There should be at least one writable bit.
+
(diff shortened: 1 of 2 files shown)


lumachad (approved):
LGTM.

gdb/testsuite/gdb.rocm/write-all-registers.exp
```diff
@@ -0,0 +1,206 @@
...
+		# Must be multiples of 4.
+		write_read_register_hex $reg 0x223344556600 "1"
+		write_read_register_hex $reg 0x778899aabb04 "2"
+	    } elseif {[is_special $reg]} {
+		# Special registers.
+
+		set reg_size [register_size $reg]
+
+		# Set register to all ones.  Read-only bits will stay
+		# at zero.
```
lancesix: Just a nit.

We have registers with read-only bits which do not have to be 0s.  We can have 1s as well which are read-only.

lancesix: Just noting, the read-only bits in "special" registers do not have to be 0s.  Those can be ones as well.

For example, gfx12 has "STATUS.W64" to tell if a wave is in wave32 or wave64 mode.  When using wave64, this bit is 1 and readonly.

That being said, I don't think we currently expose any register where this could be an issue, but we could in the future add some completely read-only (no bit can be written) registers, in which case we'll need to adjust this testcase accordingly.

---

edit: `scratch_base` / `flat_scratch` might be good candidates of registers to become "fully read-only".

gdb/testsuite/gdb.rocm/write-all-registers.exp
```diff
@@ -0,0 +1,206 @@
...
+		    set all_ones_val 0xffffffffffffffff
+		} else {
+		    fail "unsupported register size"
+		    continue
+		}
+
+		save_restore_register $reg {
+		    gdb_test "p/x \$$reg = $all_ones_val" " = $all_ones_val"
+		    # Read back the value, which should have writable bits
+		    # set to one, read-only bits set to zero.
```
lancesix: Same, this comment is probably inaccurate.

To really get the set of read-only registers, you'd need to
- write all 1s, and collect a mask of bits staying at 0.
- write all 0s and collect a mask of bits staying at 1.
- OR those two masks.

This should give a mask of read-only bits, and only a very restricted set of regs (which we do not expose today) can have that mask equal to `$all_ones_val`.

gdb/testsuite/gdb.rocm/write-all-registers.exp
```diff
@@ -0,0 +1,206 @@
...
+	return -code $code -errorinfo $errorInfo -errorcode $errorCode $result
+    } else {
+	return -code $code $result
+    }
+}
+
+# Registers that have bits that are always zero.
+proc is_special {reg} {
+    # The single list contains all the special registers from all
+    # supported architectures, sorted in alphabetical order.
```
lancesix: A good dynamic approach to discover such registers is probably to look at `info registers system`.

For example, on gfx1031, I have:

```
(gdb) info registers system
mode           0x3f0               [ FP_ROUND.32=NEAREST_EVEN FP_ROUND.64_16=NEAREST_EVEN FP_DENORM.32=FLUSH_NONE FP_DENORM.64_16=FLUSH_NONE DX10_CLAMP IEEE ]
trapsts        0x40000000          [ EXCP_CYCLE=0 EXCP_GROUP_MASK=0 DP_RATE=2 ]
flat_scratch   0x7ffee4a00000      140732734111744
status         0x98041             [ SCC SPI_PRIO=0 USER_PRIO=0 TRAP_EN TTRACE_SIMD_EN VALID PERF_EN ]
```

Note that we see here `flat_scratch` (known as `scratch_base for gfx12) which I think should probably be fully read-only.  Where the scratch for a wave is is not something we should edit, it is a property of a wave we should not be able to touch.

`pc` is probably another register with read-only bits  which is not listed here, but it is handled in a special way anyway.

lancesix (commented):
Some comments, let me know what you think.

lancesix:
Also, I am surprised that the CI passes.  There is a known issue for `state_priv` in gfx12xx (https://github.com/ROCm/rocm-systems/pull/5149) which should have made the tests fail.

lumachad:
> Also, I am surprised that the CI passes. There is a known issue for `state_priv` in gfx12xx ([ROCm/rocm-systems#5149](https://github.com/ROCm/rocm-systems/pull/5149)) which should have made the tests fail.

We don't run rocgdb tests on gfx12xx at the moment. So that would need to be exercised manually.

lumachad:
Just a FYI, I observed the number of tests differing between a run with O0 and another with O3. About half as many tests passing on O3 compared to O0. Not sure if that's by design or not.

lancesix:
> Just a FYI, I observed the number of tests differing between a run with O0 and another with O3. About half as many tests passing on O3 compared to O0. Not sure if that's by design or not.

We had work internally to enable the testsuite at 01/02/03, some of that implied disabling some tests.  Not sure how much of that work landed (but we should eventually resurrect that).

palves:
> Also, I am surprised that the CI passes. There is a known issue for `state_priv` in gfx12xx ([ROCm/rocm-systems#5149](https://github.com/ROCm/rocm-systems/pull/5149)) which should have made the tests fail.

Just for the record, the test does fail for state_priv on gfx12 currently (both linux and windows the same way), that's the bug that triggered writing the testcase in the first place.

gdb/testsuite/gdb.rocm/write-all-registers.exp
palves: > Just noting, the read-only bits in "special" registers do not have to be 0s. Those can be ones as well.
> 
> For example, gfx12 has "STATUS.W64" to tell if a wave is in wave32 or wave64 mode. When using wave64, this bit is 1 and readonly.

Did we just find a bug then?  With this:

```
--- c/gdb/testsuite/gdb.rocm/write-all-registers.cpp
+++ w/gdb/testsuite/gdb.rocm/write-all-registers.cpp
@@ -26,7 +26,7 @@ kern ()
 int
 main ()
 {
-  kern<<<1, 1>>> ();
+  kern<<<1, 64>>> ();
   CHECK (hipDeviceSynchronize ());
   return 0;
 }
diff --git c/gdb/testsuite/gdb.rocm/write-all-registers.exp w/gdb/testsuite/gdb.rocm/write-all-registers.exp
index 9e2fe7bb00c..aa4167615b9 100644
--- c/gdb/testsuite/gdb.rocm/write-all-registers.exp
+++ w/gdb/testsuite/gdb.rocm/write-all-registers.exp
@@ -22,7 +22,10 @@ require allow_hipcc_tests
 
 standard_testfile .cpp
 
-if {[build_executable "failed to prepare" $testfile $srcfile {debug hip}]} {
+set opts {debug hip}
+lappend opts additional_flags=-mwavefrontsize64
+
+if {[build_executable "failed to prepare" $testfile $srcfile $opts]} {
     return
 }

```

I see 64 lanes in one wave, and the WAVE64 bit is set, but I can clear it from gdb and it sticks:

```
(gdb) p $status
$1 = [ TRAP_EN VCCZ VALID WAVE64 ]
(gdb) p $status = 0
$2 = [ ]
(gdb) p $status
$3 = [ ]
```

So in wave 64 mode the testcase as is still passes (except for the buggy state_priv).

lancesix: I think so, yes.

The definite test should be to do:

```
(gdb) p $status = 0
...
(gdb) si
....
(gdb) p $status
```

But re-looking at the [doc](https://docs.amd.com/v/u/en-US/rdna4-instruction-set-architecture) (sect 3.4.1), the `$status` register is fully read-only.  We should not be able to update it directly.

palves: If I issue a "si" after clearing the bit, the step succeeds normally, I still have 64 lanes, and the wave64 is back set to 1 again.

lancesix: Did anything change in the `$status` value? I expect it should be the same value (unless the single step changed something which impact status, like changing `$vcc` could change `$status.vccz`).

lumachad: I don't recall gdb having a concept or strong guarantee of read-only registers. You can write to anything, but it will just not stick. Once moved the buffer will be re-fetched and things will go back to a sane state.

The alternative is failing the write, but if we are writing in bulk, that will probably fail everything, even writes to r/w registers.

palves: Right, $status gets back its original value.

```
(gdb) p $status
$13 = [ TRAP_EN VCCZ VALID WAVE64 ]
(gdb) p $status = 0
$14 = [ ]
(gdb) si
0x00007ffff62c1548      24        for (int i = 0; i < 100; i++)
(gdb) p $status 
$15 = [ TRAP_EN VCCZ VALID WAVE64 ]
```

I also confirmed that changing $exec from gdb affects $status immediately:

```
(gdb) p $exec = 0
$17 = 0
(gdb) p $status 
$18 = [ TRAP_EN EXECZ VCCZ VALID WAVE64 ]
```

(I don't recall whether we have a testcase for this.)

lancesix: I don't think we have a testcase for this, but we have code for it in dbgapi.

i.e. we should have a testcase for this!

Similarly, changing $vcc should change the vccz bit.  Let me know if you want to add that to tests right away, or I'll create a ticket to do it later (I don't have bandwidth for this right now).

lumachad (changes requested):

gdb/testsuite/gdb.rocm/write-all-registers.exp
lancesix: > I don't recall gdb having a concept or strong guarantee of read-only registers. You can write to anything, but it will just not stick. Once moved the buffer will be re-fetched and things will go back to a sane state.
> 
> The alternative is failing the write, but if we are writing in bulk, that will probably fail everything, even writes to r/w registers.

dbgapi has logic for this.  Each register can have a "read-only mask". Any write to bits in this mask are dropped.

palves: > I don't recall gdb having a concept or strong guarantee of read-only registers. You can write to anything, but it will just not stick. 

The issue here is that the write stuck (somewhere in some buffer/cache) instead of being ignored.  The expectation is that reading the register back immediately, you'd see the write was ignored.

lumachad: > > I don't recall gdb having a concept or strong guarantee of read-only registers. You can write to anything, but it will just not stick. Once moved the buffer will be re-fetched and things will go back to a sane state.
> > The alternative is failing the write, but if we are writing in bulk, that will probably fail everything, even writes to r/w registers.
> 
> dbgapi has logic for this. Each register can have a "read-only mask". Any write to bits in this mask are dropped.

But things go first into gdb register buffers, and I'm guessing we can modify those at any time.

palves: > But things go first into gdb register buffers, and I'm guessing we can modify those at any time.

Note GDB's regsiter cache is write-through.  But I guess it could happen that gdb does the write-through and updates its cache at the same time, ending up with a stuck wrong value.  But I see some bits not being writable hence having written the testcase to detect the writable bits.  I'll need to take a better look at how it works for those.  Maybe I got confused?

Regardless, if this was trully a gdb regcache issue, then clearing the gdb register cache would make the bits go back to their original state.  But they don't:

```
(gdb) p $status
$23 = [ TRAP_EN EXECZ VCCZ VALID WAVE64 ]
(gdb) p $status = 0
$24 = [ ]
(gdb) p $status
$25 = [ ]
(gdb) maintenance flush register-cache 
Register cache flushed.
(gdb) p $status
$26 = [ ]
```

palves:
```
void
amd_dbgapi_target::store_registers (struct regcache *regcache, int regno)
{
...
  /* If the register has read-only bits, invalidate the value in the regcache
     as the value actually written may differ.  */
  if (tdep->register_properties[regno]
      & AMD_DBGAPI_REGISTER_PROPERTY_READONLY_BITS)
    regcache->invalidate (regno);
```

palves: AMD_DBGAPI_REGISTER_PROPERTY_READONLY_BITS is set for that register.

However, in dbgapi, each architecture returns the read-only bits for each register with the register_read_only_mask virtual method.  For the status register though, that is currently returning null.  Sounds like that should return an actual mask for this register.

I'm a bit puzzled looking for where does the register end up with AMD_DBGAPI_REGISTER_PROPERTY_READONLY_BITS set.  I don't see a breakpoint hit in gfx12_architecture_t::register_properties for that register.  I'll look a little deeper.

lancesix: Setting the READONLY_BITS property for registers with non-nulltpr `read_only_mask` is done [here](https://github.com/ROCm/rocm-systems/blob/develop/projects/rocdbgapi/src/architecture.cpp#L2129-L2136):

```c++
amd_dbgapi_register_properties_t
amdgcn_architecture_t::register_properties (amdgpu_regnum_t regnum) const
{
  amd_dbgapi_register_properties_t properties
    = register_read_only_mask (regnum) != nullptr
        ? AMD_DBGAPI_REGISTER_PROPERTY_READONLY_BITS
        : AMD_DBGAPI_REGISTER_PROPERTY_NONE;

  ...
}
```

palves: What was confusing me earlier is that there's pseudo_status and status, two different register numbers.  
I have something that works now, but I'm now wondering whether pseudo_state_priv should be writable too.  
From the same docs, state_priv can only be written from the trap handler (priv=1), so how come you wanted to be able to write to it from rocgdb, leading to that dbgapi patch?  Can one use rocgdb to debug the trap handler?

palves: > From the same docs, state_priv can only be written from the trap handler (priv=1), so how come you wanted to be able to write to it from rocgdb, leading to that dbgapi patch? Can one use rocgdb to debug the trap handler?

I've convinced myself that GDB+dbgapi should treat state_priv as read-only too.  

See dbgapi patch at https://github.com/ROCm/rocm-systems/pull/5211, and the comments/questions in patch 2/2 of that series.  Your dbgapi patch is still useful if we have internal dbgapi writes to state_priv, though I didn't find any.

I've uploaded a new version of the testcase that assumes that dbgapi patch is applied.  Since I'm ignoring writes to state_priv, the testcase no longer covers the bug you fixed.

gdb/testsuite/gdb.rocm/write-all-registers.exp
palves: This is now improved along those lines.  Thanks, I had only originally considered the reserved (always zero) bits.

gdb/testsuite/gdb.rocm/write-all-registers.exp
palves: I've now tweaked the dbgapi patch to consider flat_scratch / scratch_base read-only too, and updated the gdb patch.  Juggling too many machines at once, missed that one...  I haven't yet tested on gfx11.

gdb/testsuite/gdb.rocm/write-all-registers.exp
lancesix: I don't think I agree for `state_priv`  There is the `HALT` bit in this register that we might want to be able to control.

Shader code can do something `s_sethalt 1` which would result in the wave being halted, and then being able to write to the `state_priv.halt` bit from the debugger would be the only way to resume execution.

I'll go and check the patch you have in dbgapi.

palves:
I'm not working on this (and the corresponding dbgapi patch) at the moment and I don't know if/when I'll have time, and the result will probably look different in the end.  So I'm closing this for now.

PR #94, opened 2026-04-21
lib/rocm.exp: rework compiler_dwarf_language{}

amd-shahab:
```
This change introduces a few alterations to the
"compiler_dwarf_language{}" procedure in rocm.exp:

1. "llvm-dwarfdump" is used instead of "objdump", because
that's the dump tool officially delivered by our releases.

2. Since looking at a fully-linked test binary may result
in multiple CUs, each with their own language, we switch
to an object file compiled from a simple function, as a
reference, to capture the compiler behaviour (if it emits
C++ or HIP as the DWARF language when compiling).

3. Now that we use the same input (a simple function),
make it a cached procedure.

-------------------------------------------------------------------

Also, add a new test: hip-lang-detect. As a result, two other
hip-builtin-* tests become a tad simpler.
```

### DIFF
--- a/gdb/testsuite/gdb.rocm/hip-lang-detect.exp
+++ b/gdb/testsuite/gdb.rocm/hip-lang-detect.exp
@@ -1,0 +1,53 @@
+# Copyright (C) 2026 Free Software Foundation, Inc.
+# Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+# This file is part of GDB.
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Verify that GDB detects the HIP program language correctly based
+# on what the compiler has emitted as the DW_LANG attribute.
+
+load_lib rocm.exp
+
+require allow_hipcc_tests
+
+standard_testfile hip-builtin-variables.cpp
+
+if {[prepare_for_testing "failed to prepare ${testfile}" \
+	$testfile $srcfile {debug hip}]} {
+    return -1
+}
+
+with_rocm_gpu_lock {
+    if {![runto_main]} {
+	return -1
+    }
+
+    set dw_lang [compiler_dwarf_language]
+    set test "expected compiler language"
+    if {$dw_lang == "hip"} {
+	set expected_re "hip"
+    } elseif {$dw_lang == "cpp"} {
+	set expected_re "c\\+\\+"
+    } else {
+	fail "$test (found \"$dw_lang\")"
+	return
+    }
+    pass $test
+
+    gdb_test "show language" \
+	"The current source language is \"auto; currently $expected_re\"\." \
+	"expected gdb language"
+}
--- a/gdb/testsuite/lib/rocm.exp
+++ b/gdb/testsuite/lib/rocm.exp
@@ -568,54 +568,68 @@ proc enable_alu_exception {excp_name} {
 	"= .*$excp_name.*"
 }
 
+# Try to find llvm-dwardump first in the PATH, then in
+# (assumed) ROCM_PATH.  If not found, return empty string.
+
+gdb_caching_proc find_dwarfdump {} {
+    global rocm_path
+    foreach ddump [list "llvm-dwarfdump" \
+			"${rocm_path}/lib/llvm/bin/llvm-dwarfdump" ] {
+	set res [catch {exec $ddump --version} output]
+	if {$res == 0} {
+	    return $ddump
+	  }
+    }
+    return ""
+}
+
 # Read the DW_AT_language of the output binary and return:
 #
-#   "hip"       -- if DW_LANG_HIP   (48, 0x30)
-#   "cpp"       -- if DW_LANG_C++14 (33, 0x21)
-#   "unknown"   -- otherwise
+#   "hip"       -- if DW_LANG_HIP (48, 0x30)
+#   "cpp"       -- if DW_LANG_C_plus_plus_14 (33, 0x21)
+#   "others"    -- if other languages
+#   "unknown"   -- in case of failure
 
-proc compiler_dwarf_language {} {
-    # Use objdump because
-    #
-    # 1. readelf cannot process non-elf binaries such as PE/COFF.
-    # 2. If warnings are emitted, objdump still returns 0, unlike readelf.
-    #
-    # In "objdump ... | grep ..." command, if there's any text printed to
-    # stderr by "objdump", DejaGnu will add them to the final "output".
-    # Moreover, DejaGnu will set the "result" to "1" if it finds anything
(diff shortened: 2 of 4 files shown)


palves:
> This change introduces a few alternations to the "compiler_dwarf_language{}" procedure in rocm.exp:

alternations => alterations ?

> 
> 1. "llvm-dwarfdump" is used instead of "objdump".  There has been a few cases that the (system) objdump did not handle the DWARF (-W) correctly.

1.2 llvm-dwarfdump is shipped with rocm/therock, objdump is not.

> 2. Instead of using the output of dwarfdump from processing the test binary, to help us detect the DWARF language, we use a lean source code compiled into an object file.  That way, we end up with a simpler binary while still capturing the behaviour of the compiler.

This is missing stating the important part.  If we look at the fully-linked test binary, then we may see multiple compilation units, each with their own language.

> 3. The return values are now "hip", "others", and "unknown". The earlier return string of "c++" was not adding any valuble information.

valuble => valuable

But I disagree with #3.  If the compiler emits c++, then we want to be sure we treat that as c++ for the host side.  It is testing the arch check in this bit of dwarf2/read.c:

	  if (lang == language_cplus
	      && (cu->per_objfile->per_bfd->obfd->arch_info->arch
		  == bfd_arch_amdgcn))

This was what was tested by this bit in the testcases:

	    gdb_test "show language" \
		"The current source language is \"auto; currently c\\+\\+\"\." \
		"C++ as the auto language on the host side"

amd-shahab:
> If the compiler emits c++, then we want to be sure we treat that as c++ for the host side

why stop at "c++" then? what if the compiler emits "fortran", ...? are we going to check those too?

palves:
> > If the compiler emits c++, then we want to be sure we treat that as c++ for the host side
> 
> why stop at "c++" then? what if the compiler emits "fortran", ...? are we going to check those too?

I don't know if you this is tongue in cheek.   There are many releases of the rocm out there that emit c++.  They won't be emitting fortran or anything else.

amd-shahab:
> I don't know if you this is tongue in cheek

apparently, texts don't capture the tone right. that was a genuine question. I've heard there might be support for other languages.

gdb/testsuite/lib/rocm.exp
```diff
@@ -570,52 +570,43 @@ proc enable_alu_exception {excp_name} {
...
-    set cmd [list $objdump_program -Wi $bin 2>&1 | grep "DW_AT_language.*:"]
+    global rocm_path
+    if {![gdb_simple_compile check_my_lang {
+	    int foo (void) { return 42; }
+	  } object {hip}]} {
+	fail "language detection (compiling a sample program)"
+	return "unknown"
+    }
+
+    set dwarfdump $rocm_path/lib/llvm/bin/llvm-dwarfdump
```
lancesix: Might be a comment beyond the scope of this particular test. Is rocm-path (with a default of /opt/rocm) still something we want to rely on with the new practices to handle deployment? Don't we instead need to rely on clang/llvm-objdump from the path should be the ones for the AMD toolchain?

gdb/testsuite/lib/rocm.exp
```diff
@@ -570,52 +570,43 @@ proc enable_alu_exception {excp_name} {
...
     }
 
-    set matches [regexp -inline "DW_AT_language\[ \t\]*:\[ \t\]*($::decimal)\[ \t\]*" $output]
+    set matches [regexp -inline -all \
+		   "DW_AT_language\[ \t\]*\\((DW_LANG_.*)\\)" \
+		   $output]
+    if {[llength $matches] != 2} {
+	fail "language detection (extracting language)"
+	return "unknown"
+    }
```
lancesix, line 624: Wouldn't it work to just test for "any of the match is HIP"? Wasn't one of your issues that you end-up linking different CUs with different languages on some targets?

gdb/testsuite/lib/rocm.exp
```diff
@@ -570,52 +570,43 @@ proc enable_alu_exception {excp_name} {
...
 # Read the DW_AT_language of the output binary and return:
 #
-#   "hip"       -- if DW_LANG_HIP   (48, 0x30)
-#   "cpp"       -- if DW_LANG_C++14 (33, 0x21)
-#   "unknown"   -- otherwise
+#   "hip"       -- if DW_LANG_HIP (48, 0x30)
+#   "others"    -- if other languages
+#   "unknown"   -- in case of failure
 
 proc compiler_dwarf_language {} {
```
lancesix: Isn't what we want to test "does the compiler support emiting DW_LANG_hip"?  i.e. building this example, does that contain anything indicating hip (return true) or not (return false)?

palves: We also want to test that the fallback code in dwarf2/read.c doesn't treat host c++ as hip.  That's what the "show language" test was testing.  It shouldn't be removed.  It'll probably be a good idea to factor that out to its own testcase, though.  It's currently in two testcases.  I just talked with shahab about this, he'll give it a try.

gdb/testsuite/lib/rocm.exp
amd-shahab: now, we're "compiling" a simple function into a `*.o` file. I've updated the commit message (point number 2) to reflect that better.

amd-shahab:
- add `find_dwarfdump{}` procedure to handle different setups.

gdb/testsuite/lib/rocm.exp
amd-shahab: this has been addressed now in `find_dwarf_dump{}`.

- First it'll try to find it in `PATH`.
- If not found, then moves on to `ROCM_PATH/lib/llvm/bin`, if the `ROCM_PATH` is set.
- If `ROCM_PATH` is not set, then `/opt/rocm/lib/llvm/bin`.
- If not in any of them, then the procedure returns an empty string.

amd-shahab:
- remove the now unused `global rocm_path` line from `compiler_dwarf_language{}` procedure.

gdb/testsuite/gdb.rocm/hip-lang-detect.exp
```diff
@@ -0,0 +1,50 @@
...
+    if {$dw_lang == "hip"} {
+	gdb_test "show language" \
+	    "The current source language is \"auto; currently hip\"\." \
+	    "HIP as the auto language on the host side"
+    } elseif {$dw_lang == "cpp"} {
+	gdb_test "show language" \
+	    "The current source language is \"auto; currently c\\+\\+\"\." \
+	    "C++ as the auto language on the host side"
+    } else {
+	fail "unexpected language \"$dw_lang\" was found."
```
palves: As principle, let's make the pass/fail be balanced and issue the same messages to gdb.sum to make it easier to compare results with different compilers.  

Also, if the compiler outputs some other language, then it's a failure on their end, so make it an xfail instead of a fail.  

Let's put the compiler language in the fail message in a ()s block so it isn't considered part of the test message, as different runs could in theory produce different $dw_lang values.  (This also removes the spurious period at end of gdb.sum message.)  Or remove the $dw_lang part from the message, it's probably already visible in gdb.log.  Up to you.

Resulting in something like:

    set test "expected compiler language"
    if {$dw_lang == "hip"} {
        set expected_re "hip"
    } elseif {$dw_lang == "cpp"} {
        set expected_re "c\\+\\+"
    } else {
	   xfail "$test (found \"$dw_lang\")"
       return
    }
    pass $test

	gdb_test "show language" \
	    "The current source language is \"auto; currently $expected_re\"\." \
	    "expected gdb language"

amd-shahab: done. used `fail` instead of `xfail`.

amd-shahab:
- make `hip-lang-detect` test summary more uniform.
- add a comment about what `[llength $matches] != 2` is supposed to mean.

palves (approved):

PR #97, opened 2026-04-22
[amd-temp-windows] Cherry-pick changes to use dbgapi-0.80.0

lancesix:
Cherry-pick the following patches:

- c388bfa797d gdb/amdgcn: Dynamically select which bits to use to hold CORE_ADDR
- bee0a5144c8 gdb/amdgpu-tdep: Add process_id and wave_id args to address_dependency


lancesix:
Odd, looks there was some version of those patches in already, but not the one that eventually landed in amd-staging. I'll have another PR to bring the delta in.

PR #107, opened 2026-04-30
Migrate and adjust CI script from TheRock

lumachad:
These 2 commits migrate the CI script from TheRock into rocgdb's repo. It will live in .github/scripts for now. They also adjust the script so it works correctly from the new location.

We will need a TheRock change to make this work, but we need a rocgdb bump to take place first before we can switch TheRock to use this new location for the CI script.

### DIFF
--- a/.github/scripts/test_rocgdb.py
+++ b/.github/scripts/test_rocgdb.py
@@ -272,6 +272,14 @@ def parse_arguments() -> argparse.Namespace:
         help="Timeout value in seconds for individual tests (max: 600). Default is 100.",
     )
     parser.add_argument(
+        "--try-rocm-path",
+        action="store_true",
+        help="Try to use ROCM_PATH environment variable to determine the ROCm tree root. "
+        "If ROCM_PATH is set and valid, it takes precedence over OUTPUT_ARTIFACTS_DIR and script location detection. "
+        "If ROCM_PATH is not set, falls back to OUTPUT_ARTIFACTS_DIR or script location detection. "
+        "If ROCM_PATH is invalid, issues an error.",
+    )
+    parser.add_argument(
         "--check-type",
         type=str,
         choices=["check", "check-read1", "check-readmore"],
@@ -1334,19 +1342,40 @@ def main() -> None:
     # Determine paths either from arguments or environment variables.
     if args.testsuite_dir is None:
         # Determine the root of the ROCm tree.
-        output_artifacts_dir = os.getenv("OUTPUT_ARTIFACTS_DIR")
-        if output_artifacts_dir is not None:
-            # OUTPUT_ARTIFACTS_DIR is set, verify it exists.
-            rocm_dir = Path(output_artifacts_dir).resolve()
-            if not rocm_dir.is_dir():
-                _log_error_and_exit(
-                    f"OUTPUT_ARTIFACTS_DIR is set to '{rocm_dir}' but the path does not exist"
+        # Priority: ROCM_PATH (if --try-rocm-path) > OUTPUT_ARTIFACTS_DIR > script location
+        rocm_dir = None
+
+        if args.try_rocm_path:
+            rocm_path_env = os.getenv("ROCM_PATH")
+            if rocm_path_env is not None:
+                tmp_dir = Path(rocm_path_env).resolve()
+                if not tmp_dir.is_dir():
+                    _log_error_and_exit(
+                        f"ROCM_PATH is set to '{tmp_dir}' but the path does not exist"
+                    )
+                rocm_dir = tmp_dir
+                logger.info(f"Using ROCM_PATH: {rocm_dir}")
+            else:
+                logger.info(
+                    "--try-rocm-path flag set but ROCM_PATH environment variable is not set. "
+                    "Falling back to OUTPUT_ARTIFACTS_DIR or script location."
                 )
-        else:
-            # OUTPUT_ARTIFACTS_DIR is not set, use script location to find root.
-            script_path = Path(__file__).resolve()
-            rocm_dir = script_path.parent.parent.parent
-            logger.info(f"OUTPUT_ARTIFACTS_DIR not set, using script-based path: {rocm_dir}")
+
+        if rocm_dir is None:
+            output_artifacts_dir = os.getenv("OUTPUT_ARTIFACTS_DIR")
+            if output_artifacts_dir is not None:
+                # OUTPUT_ARTIFACTS_DIR is set, verify it exists.
+                rocm_dir = Path(output_artifacts_dir).resolve()
+                if not rocm_dir.is_dir():
+                    _log_error_and_exit(
+                        f"OUTPUT_ARTIFACTS_DIR is set to '{rocm_dir}' but the path does not exist"
+                    )
+                logger.info(f"Using OUTPUT_ARTIFACTS_DIR: {rocm_dir}")
+            else:
+                # OUTPUT_ARTIFACTS_DIR is not set, use script location to find root.
+                script_path = Path(__file__).resolve()
+                rocm_dir = script_path.parent.parent.parent
+                logger.info(f"Using script-based path: {rocm_dir}")
 
         rocgdb_bin = rocm_dir / "bin" / "rocgdb"
         rocgdb_testsuite_dir = rocm_dir / "tests" / "rocgdb" / "gdb" / "testsuite"
--- a/.github/scripts/test_rocgdb.py
+++ b/.github/scripts/test_rocgdb.py
@@ -1,0 +1,1589 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+import argparse
+import glob
+import logging
+import os
+import platform
+import re
+import resource
+import shlex
+import shutil
+import subprocess
+import sys
+import sysconfig
+import time
+from collections import defaultdict
+from pathlib import Path
+from typing import Dict, List, Optional, Set, Tuple
+
+logging.basicConfig(
+    level=logging.INFO, format="%(asctime)s - %(levelname)s: %(message)s"
+)
+logger = logging.getLogger(__name__)
+
+# Status symbols used throughout.
+STATUS_PASS = "[✓]"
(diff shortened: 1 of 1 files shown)


.github/scripts/test_rocgdb.py
```diff
@@ -0,0 +1,1589 @@
...
+    "TIMEOUT",
+    "UNTESTED",
+    "UNSUPPORTED",
+    "XFAIL",
+    "KFAIL",
+    "FLAKY",  # Tests that failed initially but passed on retry.
+]
+
+# Category display configuration: (prefix, suffix).
+CATEGORY_DISPLAY = {
```
lancesix, line 83: Seems that the suffix is always empty, is this needed at all?

.github/scripts/test_rocgdb.py
```diff
@@ -0,0 +1,1589 @@
...
+        env: Environment variables for command.
+        capture_output: Whether to capture stdout/stderr.
+        check: Whether to raise on non-zero exit.
+        error_msg: Custom error message on failure.
+
+    Returns:
+        CompletedProcess result.
+
+    Exits:
+        Code 1 if check=True and command fails.
```
lancesix, line 132: The comment seems odd.  If `check` is true and the command fails, this raises a CalledProcessError.  Otherwise, return a `CompletedProcess`.

.github/scripts/test_rocgdb.py
```diff
@@ -0,0 +1,1589 @@
...
+  python %(prog)s --dump-failed-test-log
+
+        """,
+    )
+
+    parser.add_argument(
+        "--dump-failed-test-log",
+        action="store_false",
+        help="For failed tests, dump gdb.log to the console at the end of the run. Default is on.",
+    )
```
lancesix, line 215: Feels like the flag name should be reversed (to "skip-failed-test-log" or similar) to match the logic.

Currently, not specifying this flag would print the log, but specifying `--dump-failed-test-log` prevents the dump.

.github/scripts/test_rocgdb.py
```diff
@@ -0,0 +1,1589 @@
...
+    parser.add_argument(
+        "--no-xfail",
+        action="store_true",
+        help="Do not use ignore lists for failed tests. Default is to use the ignore lists.",
+    )
+    parser.add_argument(
+        "--optimization",
+        type=str,
+        default="",
+        help="Optimization level to pass to compiler (e.g., -O0, -Os, -Og).",
```
lancesix, line 236: This translates to `CFLAGS_FOR_TARGET`, so feels the name might be a bit restrictive, isn't it?

.github/scripts/test_rocgdb.py
```diff
@@ -0,0 +1,1589 @@
...
+    parser.add_argument(
+        "--runtestflags",
+        type=str,
+        default="",
+        help="Additional flags for RUNTESTFLAGS (e.g., '--target_board=hip -debug').",
+    )
+    parser.add_argument(
+        "--tests",
+        nargs="+",
+        default=["gdb.rocm", "gdb.dwarf2"],
```
lancesix, line 260: Is this missing the `/*.exp` bits?

lancesix: from later, no.  This will be added later, but then the help should reflect that and sate the /*.exp suffix is automatically added.

.github/scripts/test_rocgdb.py
```diff
@@ -0,0 +1,1589 @@
...
+    if not (0 < args.timeout <= 600):
+        _log_error_and_exit(
+            f"Timeout must be between 1 and 600 seconds. Got {args.timeout}."
+        )
+
+    # Validate max_failed_retries value.
+    if args.max_failed_retries < 0:
+        _log_error_and_exit(
+            f"Max failed retries must be non-negative. Got {args.max_failed_retries}."
+        )
```
lancesix, line 318: Those 2 last checks (timeout <= 600 and max_failed_retries >= 0) would be better done in the `add_argument` directly.

i.e.
```python
def positive_int(st):
    v = int(st)
    if v < 0:
        raise argparse.ArgumentTypeError(f"{st} is not a positive integer")
    return v

parser.add_argument(
        "--max-failed-retries",
        type=positive_int,
        default=3,
        help="Maximum number of times to retry failed tests. Default is 3.",
    )
```

.github/scripts/test_rocgdb.py
```diff
@@ -0,0 +1,1589 @@
...
+        #   line output by dejagnu).
+        self.test_data = defaultdict(lambda: defaultdict(lambda: defaultdict(list)))
+
+        # Track currently failed tests by compiler_label.
+        self.failed_tests: Dict[str, Set[str]] = defaultdict(set)
+
+        # Track flaky tests (failed initially but passed on retry) by compiler_label.
+        self.flaky_tests: Dict[str, Set[str]] = defaultdict(set)
+
+    def cleanup_old_entries(self, compiler_label: str, tests: str) -> None:
```
lancesix, line 345: why isn't `tests` a `List[str]`?

.github/scripts/test_rocgdb.py
```diff
@@ -0,0 +1,1589 @@
...
+            # Detect ERROR lines.
+            if line.startswith("ERROR:"):
+                msg = line[len("ERROR:") :].strip()
+                test_file = current_test_file or "UNKNOWN"
+                errors_dict[test_file].append(f"{test_file}: {msg}")
+
+        return dict(errors_dict)
+
+    def update_results(
+        self, compiler_label: str, tests: str, results_file: str
```
lancesix, line 404: same, feels like `tests` should be a `List[str]` here.

lancesix (commented):
sorry this is just a quick pass, but here a couple of comments

lumachad:
@lancesix While I don't dismiss your comments on the code itself, this PR is mostly to migrate the script and adapt it to the new location and workflow. I'd rather not address test improvements here under the risk of blocking test script migrations from TheRock. But we can have a PR meant to polish this CI script if you want to.

lancesix:
> @lancesix While I don't dismiss your comments on the code itself, this PR is mostly to migrate the script and adapt it to the new location and workflow. I'd rather not address test improvements here under the risk of blocking test script migrations from TheRock. But we can have a PR meant to polish this CI script if you want to.

I am happy to have improvements in follow-up patches.

lumachad:
> > @lancesix While I don't dismiss your comments on the code itself, this PR is mostly to migrate the script and adapt it to the new location and workflow. I'd rather not address test improvements here under the risk of blocking test script migrations from TheRock. But we can have a PR meant to polish this CI script if you want to.
> 
> I am happy to have improvements in follow-up patches.

Excellent. I think that will be best.

.github/scripts/test_rocgdb.py
```diff
@@ -1334,19 +1340,42 @@ def main() -> None:
...
+                    logger.info(
+                        f"--use-rocm-path flag set but ROCM_PATH ('{rocm_dir}') does not exist. "
+                        "Falling back to OUTPUT_ARTIFACTS_DIR or script location."
+                    )
+                    rocm_dir = None
+                else:
+                    logger.info(f"Using ROCM_PATH: {rocm_dir}")
+            else:
+                logger.info(
+                    "--use-rocm-path flag set but ROCM_PATH environment variable is not set. "
```
lancesix: I think I have the same comment as for the debug agent. I'd suggest we settle on the discussion there, and whatever comes out of it will be applicable here as well.

lumachad:
Updated to match the use of --try-rocm-path from rocr-debug-agent's script.

.github/scripts/test_rocgdb.py
```diff
@@ -1334,19 +1341,42 @@ def main() -> None:
...
-            if not rocm_dir.is_dir():
-                _log_error_and_exit(
-                    f"OUTPUT_ARTIFACTS_DIR is set to '{rocm_dir}' but the path does not exist"
+        # Priority: ROCM_PATH (if --try-rocm-path) > OUTPUT_ARTIFACTS_DIR > script location
+        rocm_dir = None
+
+        if args.try_rocm_path:
+            rocm_path_env = os.getenv("ROCM_PATH")
+            if rocm_path_env is not None:
+                rocm_dir = Path(rocm_path_env).resolve()
```
lancesix: Just as a style thing, I'd probably assign this to a temp variable and only asign the value to `rocm_dir` after we validated that the value is valid.  I find the "reset with `rocm_dir = None` later odd.

lumachad: Yeah, I think that can be cleaned up.

lumachad: Done. Could you please check the latest?

lumachad:
@lancesix Thoughts on this one?

.github/scripts/test_rocgdb.py
```diff
@@ -0,0 +1,1617 @@
...
+        type=int,
+        default=100,
+        help="Timeout value in seconds for individual tests (max: 600). Default is 100.",
+    )
+    parser.add_argument(
+        "--try-rocm-path",
+        action="store_true",
+        help="Try to use ROCM_PATH environment variable to determine the ROCm tree root. "
+        "If ROCM_PATH is set and valid, it takes precedence over OUTPUT_ARTIFACTS_DIR and script location detection. "
+        "If ROCM_PATH is not set, falls back to OUTPUT_ARTIFACTS_DIR or script location detection.",
```
lancesix: Pedantically, this leaves the "ROCm path set but invalid" case unspecified.  Just saying "Otherwise" should cover it.

lancesix (approved):
Just a nit-pick about a comment, otherwise it seems reasonable (I take this is a pure copy of the file already used in TheRock).

.github/scripts/test_rocgdb.py
lumachad: Fixed now.

lancesix:
LGTM, thanks.

Please remember that there were comments that should be addressed in a follow-up patch.

lumachad:
> LGTM, thanks.
> 
> Please remember that there were comments that should be addressed in a follow-up patch.

Right. I left them open on purpose.

PR #103, opened 2026-04-27
gdb/testsuite: support optimized code in gdb.rocm testsuite

spatrang:
## Summary

This series makes the `gdb.rocm` testsuite pass when the test programs are
compiled with optimization (`-O1/-O2/-O3`, `-flto`) passed via
`GDB_TESTCASE_OPTIONS`, rather than the implicit `-O0` the suite assumes today.

Three commits, split by concern:

1. **Update `gdb.rocm` framework to support optimized code** — adds helpers in
   `gdb/testsuite/lib/rocm.exp`:
   - `check_for_any_flags {flags}` — detects whether `GDB_TESTCASE_OPTIONS`
     contains any of the listed flags.
   - `gdb_caching_proc no_optimized_code` — predicate for `-O1/-O2/-O3`.
   - `gdb_caching_proc no_lto` — predicate for `-flto`.

2. **Update some `gdb.rocm` tests to support optimized code** (22 files) —
   two patterns applied surgically:
   - In `.cpp`/`.c`: add `volatile` on locals the test inspects, and
     `__attribute__((optnone))` on helper functions the test breakpoints into.
     Add a trivial `volatile int` statement to empty "break here" functions
     so their body is not elided.
   - In `.exp`: relax line-number arithmetic, backtrace PC formatting,
     expected source strings after `step`, and add a bounded retry loop for
     extra ticks under optimization.

3. **Skip unsupported tests in case of optimized code** (19 files) — for tests
   whose semantics cannot be reconciled with optimized code, add
   `require no_optimized_code` at the top; `mi-aspace.exp` also gets
   `require no_lto`. `register-watchpoint.exp` soft-skips the tail via
   `check_for_any_flags` after the earlier portion has run.

Diffstat: **42 files changed, +146 / -46**.

## Context

This continues the work started in internal PR
[AMD-ROCm-Internal/ROCgdb#6](https://github.com/AMD-ROCm-Internal/ROCgdb/pull/6),
rebased onto the current `amd-staging` of the public repo. The commits are
content-identical to the internal PR, preserving the original authors
(Aleksandar Rikalo, Andrey Kasaurov) with their `Co-Authored-By:` trailers
and adding my `Signed-off-by`.

PR #108, opened 2026-05-01
gdb/testsuite/gdb.rocm/finish: Test small mixed struct

lancesix:
This patch extends the gdb.rocm/finish.exp testcase to cover small (up to 64 bytes) struct containing some sub 32 bits elements (short in this case) and a float element.  This test is here extend our coverage of ABI stability.

Suggested-by: Qadeer, Hafiz Abid <HafizAbid.Qadeer@amd.com>
Change-Id: I2a56ddd7712e58d2f4fa5da21182bd4e76b1873e

### DIFF
--- a/gdb/testsuite/gdb.rocm/finish.exp
+++ b/gdb/testsuite/gdb.rocm/finish.exp
@@ -49,6 +49,12 @@ proc_with_prefix emptyStruct {} {
     }
 }
 
+proc_with_prefix mixedSmallStruct {} {
+    gdb_breakpoint "returnMixedSmallSruct" "allow-pending"
+    gdb_continue_to_breakpoint "returnMixedSmallSruct"
+    gdb_test "finish" "Value returned is \\\$$::decimal = \\{a = 3.1400001, b = \\{4, 7\\}\\}"
+}
+
 proc_with_prefix structWithCharArray {} {
     # When dealing with a struct of size up to 8 bytes, it is returned packed
     # in registers (V0 and V1).  Above that, we allocate one register per element
@@ -156,6 +162,7 @@ proc_with_prefix struct_with_static {} {
 
 with_rocm_gpu_lock {
     emptyStruct
+    mixedSmallStruct
     structWithCharArray
     unionMember
     flexibleAraryMember
--- a/gdb/testsuite/gdb.rocm/finish.exp
+++ b/gdb/testsuite/gdb.rocm/finish.exp
@@ -39,13 +39,13 @@ proc_with_prefix emptyStruct {} {
     gdb_breakpoint "returnEmpty" "allow-pending"
     with_test_prefix "empty struct" {
 	gdb_continue_to_breakpoint "returnEmpty"
-	gdb_test "finish" "Value returned is \\\$$::decimal = {<No data fields>}"
+	gdb_test "finish" "Value returned is \\\$$::decimal = \\{<No data fields>\\}"
     }
 
     gdb_breakpoint "returnEmptyWithStatic" "allow-pending"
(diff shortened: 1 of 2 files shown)


abidh (approved):
LGTM.

gdb/testsuite/gdb.rocm/finish.exp
```diff
@@ -49,6 +49,12 @@ proc_with_prefix emptyStruct {} {
     }
 }
 
+proc_with_prefix mixedSmallStruct {} {
+    gdb_breakpoint "returnMixedSmallSruct" "allow-pending"
+    gdb_continue_to_breakpoint "returnMixedSmallSruct"
+    gdb_test "finish" "Value returned is \\\$$::decimal = \\\{a = 3.1400001, b = \\\{4, 7\\\}\\\}"
```
amd-shahab: I could get the _expected_ string match with one less `\`

```
"Value returned is \\$$::decimal = \\{a = 3.1400001, b = \\{4, 7\\}\\}"
```

am I missing something?

lancesix: Isn't it `\\` so expect parsing turns it into `\`, then `\$` for `$`, so the effective regex is `\$`, no?

Let me double check.

lancesix: I guess both of us have bits right, but both end up working.

First, for the `\\\$` case, we have `\\` (for `\`), then `\$` for `$`, so this gives `\$` as expected.  If we do `\\$`, we have the first `\` escaped properly, then the `$` should start a variable substitution.  However, because we have `$$::decimal`, the character after `$` is another `$`, so not a variable name, so TCL abandons the substitution and just keeps the `$`.

So I'd say that `\\\$` is the proper way, `\\$` only work because what comes next is not a valid name.

The other case here is `\\{` v.s. `\\\{`.  This time, again the first `\\` is not in question, and we have `\{` v.s. `{`.  This one I got wrong, in double quote environment, `{` has no special meaning so does not need to be escaped.  However, because `\{` is not particular substitution, TCL just drops the `\` and we are left with the expected `{`.

So long story short, I believe the proper way should be

```
"Value returned is \\\$$::decimal = \\{a = 3.1400001, b = \\{4, 7\\}\\}"
```

I'll update that shortly.

lancesix:
I have added another patch tightening the escaping of {} in TCL.

lumachad:
If possible, we should make sure this doesn't introduce additional optimized testsuite run failures. See https://github.com/ROCm/ROCgdb/pull/103/changes#diff-e3a7e92eaec07842270f3368e58638fc70df14eaa828b7e2fcf00a4f2dab72ca for example on how this test was adjusted to not cause FAIL's when compiled with O1~O3.

lancesix:
> If possible, we should make sure this doesn't introduce additional optimized testsuite run failures. See https://github.com/ROCm/ROCgdb/pull/103/changes#diff-e3a7e92eaec07842270f3368e58638fc70df14eaa828b7e2fcf00a4f2dab72ca for example on how this test was adjusted to not cause FAIL's when compiled with O1~O3.

I have added `__attribute__((noinline))` to the current patch.

amd-shahab:
2nd patch commit message:

The gdb.rocm/finish.exp seem**s** to have somewhat invalid

shouldb e -> should be


1st patch commit message:
This test is here extend our coverage of ABI stability --> ???

amd-shahab (approved):
with a few nits mentioned separately.

lumachad (approved):
The commit messages have some typos in them. Might want to clean those up.

Otherwise looks OK to me and confirmed they run OK.

PR #109, opened 2026-05-04
gdb/amd-dbgapi: fix -Wdeprecated-enum-enum-conversion warning

simark:
With gcc 16, I get:

    /home/smarchi/src/binutils-gdb/gdb/amd-dbgapi-target.c: In function ‘void info_dispatches_command(const char*, int)’:
    /home/smarchi/src/binutils-gdb/gdb/amd-dbgapi-target.c:4729:26: error: bitwise operation between different enumeration types ‘amd_dbgapi_dispatch_barrier_t’ and ‘amd_dbgapi_dispatch_fence_scope_t’ is deprecated [-Werror=deprecated-enum-enum-conversion]
     4729 |             if ((barrier | acquire) && release)
          |                  ~~~~~~~~^~~~~~~~~

IIUC, the intent of this is just to know if we've written something in `ss` already, and if we're going to write more.  And if so, write a `|` separator.  The `|` should probably have been a `||`, or because we like to be explicit in GDB:

    if ((barrier != AMD_DBGAPI_DISPATCH_BARRIER_NONE
         || acquire != AMD_DBGAPI_DISPATCH_FENCE_SCOPE_NONE)
        && release != AMD_DBGAPI_DISPATCH_FENCE_SCOPE_NONE)
      ss << "|";

But I think it's simpler to rewrite this to append to a string, and append a `|` if we're about to append something and the string is not empty.  A stringstream has no advantage over using a string in this case.

Change-Id: I48307186f00a9943f6be388960c85ce53d10b0c3

## Submission Checklist

### DIFF
--- a/gdb/amd-dbgapi-target.c
+++ b/gdb/amd-dbgapi-target.c
@@ -4714,27 +4714,30 @@ info_dispatches_command (const char *args, int from_tty)
 	      error (_("amd_dbgapi_dispatch_get_info failed (%s)"),
 		     get_status_string (status));
 
-	    std::stringstream ss;
-	    if (barrier == AMD_DBGAPI_DISPATCH_BARRIER_PRESENT)
-	      ss << "B";
+	    std::string str;
 
-	    if (barrier && acquire)
-	      ss << "|";
+	    auto append_str = [&str] (const char *s)
+	      {
+		if (!str.empty ())
+		  str += "|";
+
+		str += s;
+	      };
+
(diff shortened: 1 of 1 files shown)


lumachad (approved):
Looks OK to me. I think the readability took a hit here, but I don't feel strongly about it if it works.

lancesix (approved):
LGTM, thanks.

I think this makes things cleaner as well.

lumachad:
@simark Merge?

PR #111, opened 2026-05-05
[05-05-2026] Merge upstream/master into amd-staging

lancesix:
Handle conflicts introduced by the following upstream patches:

    commit https://github.com/ROCm/ROCgdb/commit/57cb42b69f90784bb57b8c5342765d7439dc693b
    Date:   Tue Apr 28 18:23:55 2026 +0200

        [gdb/symtab] Factor out new_symbol variant

This patch heavily reworks new_symbol in gdb/dwarf2/read.c.  In the end,
we did not have much changes downstream to this file, we only touched
code to treat HIP as a C++ dialect.  I took the code from upstream, and
made sure to have the HIP as a C++ dialect reinstated.

    commit https://github.com/ROCm/ROCgdb/commit/329a53a6d590e2e90f590c89473990040a86c8e0
    Date:   Sat Nov 22 11:03:57 2025 -0700

        Some cleanups to "pretend language" handling

Minor change which did not cause a conflict, but caused a build failure.
It turned a LANG variable into an optional, just adjust the place using
it.

    commit https://github.com/ROCm/ROCgdb/commit/5a3726062a2c50b28592551cadf20118fe603b29
    Date:   Tue Apr 28 20:55:42 2026 -0400

        gdb/dwarf2: pass around DWARF expressions as gdb::array_view

This upstream change caused most of the conflicts, but the resolution
was quite automatic.

Finally, starting with:

    commit https://github.com/ROCm/ROCgdb/commit/c76bd1b0f757cbb5f361ec5c994a17769a12825a
    Date:   Wed Apr 8 07:00:59 2026 -0600

        Remove symbol lookup of arguments

upstream GDB seems to change how argument pack are handled.

Consider the following code:

    template<typename... Args>
    void
    foo (Args... args)
    {
    }

    int
    main ()
    {
      foo<int, long, float> (1, 2, 3.14);
      return 0;
    }

Prior to this commit, stepping into FOO would have GDB show:

    (gdb) frame
    #0  foo<int, long, float> (args=3.1400001, args=3.1400001, args=3.1400001) at tmp.cpp:7
    7       }

This is wrong.  Even if all of the arguments of the pack have the same
name (args), the do not have the same type nor value.

With this upstream change, we now see:

    (gdb) frame
    #0  foo<int, long, float> (args=1, args=2, args=3.1400001) at tmp.cpp:7
    7       }

which is correct.

We had one test downstream exercising something like this, but expected
the "wrong" behaviour: gdb.rocm/names.exp.  This merge also adjusts it
to expect the new behaviour.

### DIFF
--- a/gdb/dwarf2/expr.c
+++ b/gdb/dwarf2/expr.c
@@ -783,9 +783,8 @@ dwarf_expr_context::fetch (int n)
 
 /* See expr.h.  */
 
-void
-dwarf_expr_context::get_frame_base (const gdb_byte **start,
-				    size_t * length)
+gdb::array_view<const gdb_byte>
+dwarf_expr_context::get_frame_base ()
 {
   ensure_have_frame (this->m_frame, "DW_OP_fbreg");
 
@@ -804,9 +803,8 @@ dwarf_expr_context::get_frame_base (const gdb_byte **start,
      something has gone wrong.  */
   gdb_assert (framefunc != NULL);
 
-  func_get_frame_base_dwarf_block (framefunc,
-				   get_frame_address_in_block (this->m_frame),
-				   start, length);
+  return func_get_frame_base_dwarf_block
+    (framefunc, get_frame_address_in_block (this->m_frame));
 }
 
 /* See expr.h.  */
@@ -848,7 +846,7 @@ dwarf_expr_context::dwarf_call (cu_offset die_cu_off)
   /* DW_OP_call_ref is currently not supported.  */
   gdb_assert (block.per_cu == this->m_per_cu);
 
-  this->eval (block.data, block.size);
+  this->eval (block.expr ());
 }
 
 /* See expr.h.  */
@@ -921,13 +919,12 @@ dwarf_expr_context::push_dwarf_reg_entry_value (call_site_parameter_kind kind,
     = dwarf_expr_reg_to_entry_parameter (this->m_frame, kind, kind_u,
 					 &caller_per_cu,
 					 &caller_per_objfile);
-  const gdb_byte *data_src
-    = deref_size == -1 ? parameter->value : parameter->data_value;
-  size_t size
-    = deref_size == -1 ? parameter->value_size : parameter->data_value_size;
+  auto expr = (deref_size == -1
+	       ? parameter->value_expr ()
+	       : parameter->data_value_expr ());
 
   /* DEREF_SIZE size is not verified here.  */
-  if (data_src == nullptr)
+  if (expr.empty ())
     throw_error (NO_ENTRY_VALUE_ERROR,
 		 _("Cannot resolve DW_AT_call_data_value"));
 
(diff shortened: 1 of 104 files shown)


gdb/dwarf2/expr.h
```diff
@@ -59,29 +59,28 @@ type *address_type (gdbarch *arch, int addr_size);
 void dwarf_expr_require_composition (const gdb_byte *, const gdb_byte *,
 				     const char *);
 
-/* If <BUF..BUF_END] contains DW_FORM_block* with single DW_OP_reg* return the
+/* IfBLOCK contains DW_FORM_block* with single DW_OP_reg* return the
```
simark: Missing space.

gdb/dwarf2/expr.c
```diff
@@ -2918,7 +2917,7 @@ struct dwarf_expr_context
   /* Return the location expression for the frame base attribute, in
      START and LENGTH.  The result must be live until the current
```
simark: Comment needs updating.

gdb/dwarf2/expr.c
simark: I suggest you change this one to take an array_view.  Its 2 callers have an array view ready to pass.

gdb/dwarf2/read.c
```diff
@@ -15494,497 +15482,494 @@ new_symbol_file_line (struct die_info *die, struct dwarf2_cu *cu,
...
+      suppress_add = true;
+      [[fallthrough]];
+    case DW_TAG_class_type:
+    case DW_TAG_interface_type:
+    case DW_TAG_structure_type:
+    case DW_TAG_union_type:
+    case DW_TAG_set_type:
+    case DW_TAG_enumeration_type:
+      if (cu->lang () == language_c
+	  || is_cplus_dialect (cu->lang ())
```
simark, line 15779: This seems fine, but I see more `== language_cplus` in this file that I think are introduced in this merge.  They should probably be changed too?

gdb/dwarf2/read.c
```diff
@@ -18269,7 +18249,7 @@ cutu_reader::prepare_one_comp_unit (struct dwarf2_cu *cu,
 	     with a "set language hip" command in GDB itself.  Still, this
 	     workaround works better than checking the "producer", since
 	     various providers tend to name them slightly differently.  */
-	  if (lang == language_cplus
+	  if (is_cplus_dialect (*lang)
```
simark: Does this one really need to change?  I think in this case we really just want to check for `language_cplus`, because it's the spot where we change cplus -> hip, if needed.

simark (changes requested):

gdb/dwarf2/expr.c
lancesix: I wanted to do that as a follow-up commit for that, but can also do it in the merge commit.

lumachad: > I wanted to do that as a follow-up commit for that, but can also do it in the merge commit.

I'd vouch for doing a follow-up, as this merge is holding other PR's.

lancesix: I have the change as an independent commit on the top of the branch now.

gdb/dwarf2/read.c
lancesix: Thanks for catching this.  I somehow ended up doing the opposite of what I wanted: change the `lang == language_cplus` in the fallback detection code (next comment) while keeping other instances around when should have gone.

I rectified all of those now.

lancesix:
I should have addressed all the issues found by @simark (thanks for the meticulous review), should be good for another round now.

simark (approved):

PR #114, opened 2026-05-07
Some minor cleanups around fetch/store registers.

aktemur:


lancesix (commented):
The patches look good to me, but I expect those should apply upstream. I'd suggest sending them upstream, they'll reach amd-staging from the automatic merge (if possible, when sending upstream, can you CC me?)

aktemur:
> The patches look good to me, but I expect those should apply upstream. I'd suggest sending them upstream, they'll reach amd-staging from the automatic merge (if possible, when sending upstream, can you CC me?)

Sure.

aktemur:
Closing the PR to submit to upstream.

PR #115, opened 2026-05-07
[amd-staging-rocgdb-16] gdb: Skip rocm watchpoint tests in case not all GPUs are supported

motokultivator:
In the case where there are multiple GPUs on the system, and some of them are not supported, rocgdb issues a warning about this during the "run" command. This change introduces recognition of this warning and skips tests that would fail in such cases.

A new procedure has been introduced, in rocm.exp, for this purpose, which can be extended to recognize other specific warnings.

The tests affected by this are:

gdb.rocm/aspace-watchpoint.exp
gdb.rocm/watchpoint-at-end-of-shader.exp
gdb.rocm/precise-memory-warning-watchpoint.exp


lumachad:
Please target amd-staging. amd-staging-rocgdb-16 is only supposed to get cherry-picks.

simark:
Also, if there are changes that would apply to upstream, would you like to send them upstream?  I don't mind doing it but if you want to do it by yourself it's fine to, just let me know.

lumachad (changes requested):
Wrong branch.

lancesix (commented):
First, please submit against amd-staging.

Secondly: is it really true that we cannot support watchpoint tests?  I really don't see them as any different from the other tests:  if we run the workload on the GPU which does not support the watchpoint, then yes, we can't do anything about the watchpoint, but nothing works, for any test.

However, if we do not use the unsupported GPU, we should still be able to use watchpoint on the supported GPU (unless of course there is a strong reason not to, but I don't really see one at the moment).

lumachad:
Please open against amd-staging.

PR #117, opened 2026-05-07
Move updated rocgdb CI script from TheRock to rocgdb

lumachad:
The CI scripts should live within their components as opposed to within
TheRock. This change makes that happen.

### DIFF
--- a/.github/scripts/sanitizers/asan.supp
+++ b/.github/scripts/sanitizers/asan.supp
@@ -1,0 +1,3 @@
+# Add suppression patterns here, one per line.
+# See https://clang.llvm.org/docs/AddressSanitizer.html#suppressing-reports-in-external-libraries
+# for the AddressSanitizer suppression file format.
--- a/.github/scripts/test_rocgdb.py
+++ b/.github/scripts/test_rocgdb.py
@@ -1,8 +1,10 @@
+#!/usr/bin/env python3
 # Copyright Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
 
 import argparse
 import glob
+import json
 import logging
 import os
 import platform
@@ -10,13 +12,14 @@ import re
 import resource
 import shlex
 import shutil
+import signal
 import subprocess
 import sys
 import sysconfig
 import time
 from collections import defaultdict
 from pathlib import Path
-from typing import Dict, List, Optional, Set, Tuple
+from typing import Dict, List, NoReturn, Optional, Set, Tuple, Union
 
 logging.basicConfig(
     level=logging.INFO, format="%(asctime)s - %(levelname)s: %(message)s"
@@ -28,42 +31,6 @@ STATUS_PASS = "[✓]"
 STATUS_FAIL = "[X]"
 STATUS_WARN = "[!]"
 STATUS_FLAKY = "[~]"
-
-# A list of tests we know FAIL but we are OK with it.
-#
-# The "Generic" key tracks failures we should ignore for any compiler
-# label.
-#
-# Each compiler label key contains a list of tests for which failures should
-# be ignored.
-XFAILED_TESTS = {
-    "Generic": [
-        "gdb.rocm/corefile.exp",
-        "gdb.rocm/device-interrupt.exp",
-        "gdb.rocm/load-core-remote-system.exp",
-    ],
-    "GCC": [],
-    "LLVM": [
-        "gdb.dwarf2/ada-valprint-error.exp",
-        "gdb.dwarf2/dw2-case-insensitive.exp",
-        "gdb.dwarf2/dw2-cp-infcall-ref-static.exp",
-        "gdb.dwarf2/dw2-empty-inline-ranges.exp",
-        "gdb.dwarf2/dw2-entry-pc.exp",
-        # Testcase expects ASLR OFF and amd-llvm generates
-        # PIE executables by default.  Fix being pursued upstream.
-        # REMOVE when the fix makes its way to our branch.
-        "gdb.dwarf2/dw2-entry-value.exp",
-        "gdb.dwarf2/dw2-inline-param.exp",
-        "gdb.dwarf2/dw2-param-error.exp",
-        "gdb.dwarf2/dw2-skip-prologue.exp",
-        "gdb.dwarf2/dw2-undefined-ret-addr.exp",
-        "gdb.dwarf2/dw2-unresolved.exp",
-        "gdb.dwarf2/dw2-unexpected-entry-pc.exp",
-        "gdb.dwarf2/fission-base.exp",
-        "gdb.dwarf2/fission-dw-form-strx.exp",
-        "gdb.dwarf2/pr13961.exp",
-    ],
-}
 
 # Test result categories.
 TEST_CATEGORIES = [
@@ -79,22 +46,38 @@ TEST_CATEGORIES = [
     "FLAKY",  # Tests that failed initially but passed on retry.
 ]
 
-# Category display configuration: (prefix, suffix).
+# Default wall-clock timeout (seconds) for each `make check` invocation in
+# --one-by-one mode when --one-by-one-test-timeout is not provided.
+ONE_BY_ONE_DEFAULT_WALL_CLOCK_TIMEOUT = 3600
+
+# Sanitizers we pre-configure on every run. If rocgdb wasn't built with the
+# corresponding -fsanitize=..., the matching *_OPTIONS env var is silently
+# ignored at runtime, so listing extras here is harmless.
+SANITIZERS = ("asan", "tsan", "ubsan", "msan", "lsan", "hwasan")
+
+# Patterns for parsing DejaGnu .sum lines.
+RESULT_LINE_RE = re.compile(
+    r"^(PASS|FAIL|XFAIL|UNTESTED|UNSUPPORTED|KFAIL|UNRESOLVED): (.+)$"
+)
+TIMEOUT_RE = re.compile(r"\(timeout\)")
+RUNNING_EXP_RE = re.compile(r"Running\s+(\S+\.exp)")
+
+# Category display configuration: prefix symbol.
 CATEGORY_DISPLAY = {
-    "PASS": (STATUS_PASS, ""),
-    "FAIL": (STATUS_FAIL, ""),
-    "ERROR": (STATUS_FAIL, ""),
-    "UNRESOLVED": (STATUS_FAIL, ""),
-    "TIMEOUT": (STATUS_WARN, ""),
-    "UNTESTED": (STATUS_WARN, ""),
-    "UNSUPPORTED": (STATUS_WARN, ""),
-    "XFAIL": (STATUS_WARN, ""),
-    "KFAIL": (STATUS_WARN, ""),
-    "FLAKY": (STATUS_FLAKY, ""),
+    "PASS": STATUS_PASS,
+    "FAIL": STATUS_FAIL,
+    "ERROR": STATUS_FAIL,
+    "UNRESOLVED": STATUS_FAIL,
+    "TIMEOUT": STATUS_WARN,
+    "UNTESTED": STATUS_WARN,
+    "UNSUPPORTED": STATUS_WARN,
+    "XFAIL": STATUS_WARN,
+    "KFAIL": STATUS_WARN,
+    "FLAKY": STATUS_FLAKY,
 }
 
 
-def _log_error_and_exit(message: str, exit_code: int = 1) -> None:
+def _log_error_and_exit(message: str, exit_code: int = 1) -> NoReturn:
     """
     Log an error message and exit the program.
 
@@ -102,7 +85,7 @@ def _log_error_and_exit(message: str, exit_code: int = 1) -> None:
         message: Error message to log.
         exit_code: Exit code (default 1).
     """
-    logger.info(f"{STATUS_FAIL} Error: {message}")
+    logger.error(f"{STATUS_FAIL} Error: {message}")
     sys.exit(exit_code)
 
 
@@ -113,6 +96,8 @@ def _run_command(
     capture_output: bool = False,
     check: bool = True,
     error_msg: Optional[str] = None,
+    timeout: Optional[int] = None,
+    kill_process_group: bool = False,
 ) -> subprocess.CompletedProcess:
     """
     Run a command with consistent error handling.
@@ -124,26 +109,65 @@ def _run_command(
         capture_output: Whether to capture stdout/stderr.
         check: Whether to raise on non-zero exit.
         error_msg: Custom error message on failure.
+        timeout: Optional wall-clock timeout in seconds; raises
+            subprocess.TimeoutExpired if exceeded.
+        kill_process_group: Run the command in a new process group/session and,
+            on TimeoutExpired, SIGKILL the entire group so descendants
+            (e.g. runtest/expect/gdb spawned by `make`) are not orphaned.
 
     Returns:
         CompletedProcess result.
 
     Exits:
-        Code 1 if check=True and command fails.
+        Code 1 if check=True, command exits with non-zero status, and error_msg is provided.
+
+    Raises:
+        CalledProcessError if check=True, command exits with non-zero status, and error_msg is None.
+        TimeoutExpired if timeout is set and exceeded.
     """
+    stdout = subprocess.PIPE if capture_output else None
+    stderr = subprocess.PIPE if capture_output else None
     try:
-        return subprocess.run(
+        process = subprocess.Popen(
             cmd,
             cwd=str(cwd) if cwd else None,
-            check=check,
-            capture_output=capture_output,
             env=env,
             text=True,
+            stdout=stdout,
+            stderr=stderr,
+            start_new_session=kill_process_group,
         )
-    except subprocess.CalledProcessError as e:
+    except OSError as e:
         if error_msg:
             _log_error_and_exit(f"{error_msg}: {e}")
+        else:
+            _log_error_and_exit(f"Failed to execute {cmd[0]!r}: {e}")
+    try:
+        out, err = process.communicate(timeout=timeout)
+    except BaseException:
+        # Cover TimeoutExpired and any other escaping exception
+        # (KeyboardInterrupt, OSError mid-read, etc.) so the make /
+        # runtest / expect / gdb process tree is never orphaned. With a new
+        # session we can SIGKILL the whole group; otherwise just the child.
+        if kill_process_group:
+            try:
+                os.killpg(process.pid, signal.SIGKILL)
+            except ProcessLookupError:
+                pass
+        else:
+            process.kill()
+        try:
+            out, err = process.communicate()
+        except Exception:
+            out, err = None, None
         raise
+    result = subprocess.CompletedProcess(cmd, process.returncode, out, err)
+    if check and process.returncode != 0:
+        exc = subprocess.CalledProcessError(process.returncode, cmd, out, err)
+        if error_msg:
+            _log_error_and_exit(f"{error_msg}: {exc}")
+        raise exc
+    return result
 
 
 def _read_file_lines(file_path: Path, error_context: str = "") -> List[str]:
@@ -160,9 +184,10 @@ def _read_file_lines(file_path: Path, error_context: str = "") -> List[str]:
     try:
         with open(file_path, encoding="utf-8") as f:
             return [line.strip() for line in f]
-    except FileNotFoundError:
+    except OSError as e:
         _log_error_and_exit(
-            f"{file_path} not found{'. ' + error_context if error_context else ''}"
+            f"Failed to read {file_path}: {e}"
+            f"{'. ' + error_context if error_context else ''}"
         )
 
 
@@ -176,7 +201,115 @@ def _extract_test_file_from_name(test_name: str) -> str:
     Returns:
         Test file path without description (e.g., "gdb.rocm/foo.exp").
     """
-    return test_name.split(": ", 1)[0] if ": " in test_name else test_name
+    return test_name.split(": ", 1)[0]
+
+
+def load_ignore_list_from_json(json_path: Path) -> Dict[str, List[str]]:
+    """
+    Load test ignore list from JSON file.
+
+    Args:
+        json_path: Path to JSON file containing ignore list.
+
+    Returns:
+        Dictionary mapping compiler labels to lists of ignored test paths.
+        Returns empty dict if file doesn't exist.
+
+    Exits:
+        Code 1 if JSON file exists but is malformed.
+    """
+    if not json_path.exists():
+        logger.warning(
+            f"Ignore list file not found: {json_path}. Using empty ignore list."
+        )
+        return {}
+
+    try:
+        with open(json_path, encoding="utf-8") as f:
+            data = json.load(f)
+    except OSError as e:
+        _log_error_and_exit(f"Failed to read {json_path}: {e}")
+    except json.JSONDecodeError as e:
+        _log_error_and_exit(f"Failed to parse JSON from {json_path}: {e}")
+
+    if not isinstance(data, dict):
+        _log_error_and_exit(
+            f"Invalid ignore list format in {json_path}: expected a JSON object"
+        )
+
+    # Validate structure: each key should map to a list of strings.
+    for key, value in data.items():
+        if not isinstance(value, list):
+            _log_error_and_exit(
+                f"Invalid ignore list format in {json_path}: "
+                f"key '{key}' should map to a list"
+            )
+        if not all(isinstance(item, str) for item in value):
+            _log_error_and_exit(
+                f"Invalid ignore list format in {json_path}: "
+                f"all items in list for key '{key}' should be strings"
+            )
+
+    return data
+
+
+def _non_negative_int(value: str) -> int:
+    """
+    Validate argument is a non-negative integer.
+
+    Args:
+        value: String value from command line.
+
+    Returns:
+        Non-negative integer value.
+
+    Raises:
+        argparse.ArgumentTypeError: If value is negative.
+    """
+    ivalue = int(value)
+    if ivalue < 0:
+        raise argparse.ArgumentTypeError(f"{value} is not a non-negative integer")
+    return ivalue
+
+
+def _timeout_value(value: str) -> int:
+    """
+    Validate timeout is a positive integer (seconds).
+
+    Args:
+        value: String value from command line.
+
+    Returns:
+        Integer timeout value.
+
+    Raises:
+        argparse.ArgumentTypeError: If value is not a positive integer.
+    """
+    ivalue = int(value)
+    if ivalue <= 0:
+        raise argparse.ArgumentTypeError(
+            f"timeout must be a positive integer (got {ivalue})"
+        )
+    return ivalue
+
+
+def _positive_nonzero_int(value: str) -> int:
+    """
+    Validate argument is a positive non-zero integer.
+
+    Args:
+        value: String value from command line.
+
+    Returns:
+        Positive non-zero integer value.
+
+    Raises:
+        argparse.ArgumentTypeError: If value is not positive and non-zero.
+    """
+    ivalue = int(value)
+    if ivalue <= 0:
+        raise argparse.ArgumentTypeError(f"{value} is not a positive non-zero integer")
+    return ivalue
 
 
 def parse_arguments() -> argparse.Namespace:
@@ -191,27 +324,36 @@ def parse_arguments() -> argparse.Namespace:
     """
     parser = argparse.ArgumentParser(
         description="Run ROCgdb test suite with different compilers.",
-        formatter_class=argparse.RawDescriptionHelpFormatter,
+        formatter_class=lambda prog: argparse.RawDescriptionHelpFormatter(
+            prog, width=100
+        ),
         epilog="""Examples:
   python %(prog)s --testsuite-dir /path/to/testsuite --rocgdb-bin /path/to/rocgdb
+  python %(prog)s --rocgdb-bin /path/to/rocgdb
   python %(prog)s --tests gdb.base/break.exp gdb.base/call-ar-st.exp
+  python %(prog)s --gpu-tests
+  python %(prog)s --cpu-tests
   python %(prog)s --parallel
   python %(prog)s --group-results
-  python %(prog)s --timeout 600
+  python %(prog)s --default-timeout 300
+  python %(prog)s --retry-timeout 600
   python %(prog)s --max-failed-retries 2
   python %(prog)s --optimization="-O0"
   python %(prog)s --runtestflags="--target_board=hip -debug"
   python %(prog)s --no-xfail
   python %(prog)s --quiet
-  python %(prog)s --dump-failed-test-log
+  python %(prog)s --skip-failed-test-log
+  python %(prog)s --output-ignore-list-file custom_ignore_list.json
+  python %(prog)s --one-by-one --tests gdb.rocm/foo.exp
 
         """,
     )
 
     parser.add_argument(
-        "--dump-failed-test-log",
+        "--skip-failed-test-log",
         action="store_false",
-        help="For failed tests, dump gdb.log to the console at the end of the run. Default is on.",
+        dest="dump_failed_test_log",
+        help="Skip dumping gdb.log to the console for failed tests. Default is to dump the log.",
     )
     parser.add_argument(
         "--group-results",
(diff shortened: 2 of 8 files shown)


lumachad:
FYI, the changes in the testing script won't be reflected in the CI. Manual testing required.

lumachad:
Marking ci:skip as this won't get auto-tested anyway.

lumachad:
Please don't review yet. I have a few more changes I would like to make before we go through the whole file.

lumachad:
Should be good for a review now.

.github/scripts/test_rocgdb-review.py
```diff
@@ -0,0 +1,2423 @@
+# Copyright Advanced Micro Devices, Inc.
```
lancesix: Could use a `#! /usr/bin/env python3` or something.  WDYT?

.github/scripts/test_rocgdb-review.py
```diff
@@ -0,0 +1,2423 @@
...
+            options.append("detect_leaks=0")
+        if log_dir_ready:
+            options.append(f"log_path={log_dir}/log")
+
+        suppression_file = suppressions_dir / f"{sanitizer}.supp"
+        if suppression_file.is_file():
+            options.append(f"suppressions={suppression_file}")
+
+        env_var_name = f"{sanitizer.upper()}_OPTIONS"
+        env_vars[env_var_name] = ":".join(options)
```
lancesix: Should that append options to any pre-existing one if any?

.github/scripts/test_rocgdb-review.py
```diff
@@ -0,0 +1,2423 @@
...
+    ld_library_parts = [
+        f"{artifacts_dir}/lib",
+        f"{artifacts_dir}/llvm/lib",
+        f"{artifacts_dir}/lib/llvm/lib",
+    ]
+    _prepend_to_path_var(env_vars, "PATH", path_parts)
+    _prepend_to_path_var(env_vars, "LD_LIBRARY_PATH", ld_library_parts)
+
+    # Configure GPU core dump pattern.
+    env_vars["HSA_COREDUMP_PATTERN"] = "gpucore.%p"
```
lancesix: Aren't the testcases setting this when needed?  If a test does not set it, isn't the runtime default good?

Not saying this is bad, just not sure to see why this is needed.

.github/scripts/test_rocgdb-review.py
```diff
@@ -0,0 +1,2423 @@
...
+
+    Returns:
+        None
+
+    Exits:
+        Code 1 if ROCgdb fails to execute or its python validation fails.
+    """
+    print_section("ROCgdb launcher data")
+    env = os.environ.copy()
+    env["ROCGDB_WRAPPER_DEBUG"] = "1"
```
lancesix: Could be in the try block below, not needed after that initial run

lancesix (commented):
Just very minor comments.
I expect this is mostly moving a file from TheRock (where it was reviewed already) to gdb, do I did not check it too deeply, but overall that looks good to me.

Thanks

lancesix (approved):

lumachad:
> Just very minor comments. I expect this is mostly moving a file from TheRock (where it was reviewed already) to gdb, do I did not check it too deeply, but overall that looks good to me.
> 
> Thanks

Actually there are quite a few differences from the current version in TheRock. A number of bug fixes, additional options etc. You might want to go through them to make sure it looks OK from your side. I've opened a PR like this (a single new file) so it gives you the overall code as opposed to dozens of commits on top of the current one in TheRock.

Edit: An earlier PR did move the file from TheRock (test_rocgdb.py, unmodified by this PR). But the new file (test_rocgdb-review.py) is an improved version.

.github/scripts/test_rocgdb-review.py
lumachad: Yes.

.github/scripts/test_rocgdb-review.py
lumachad: Yes, that would be nice indeed.

.github/scripts/test_rocgdb-review.py
lumachad: This is a bit historical. We used to have a bug here where the runtime defaulted to something different the tests didn't quite expect. The CI script came about around that time. I think it's safe to drop this setting completely then.

.github/scripts/test_rocgdb-review.py
lumachad: True.

lumachad:
Ready to merge. I'll put up a PR to enable use of this script by our ROCgdb CI before merging it.

lumachad:
Missing the xfail file. I need to check where I'll get that from.

.github/scripts/sanitizers/asan.supp
```diff
@@ -0,0 +1,3 @@
+# Add suppression patterns here, one per line.
+# See https://clang.llvm.org/docs/AddressSanitizer.html#suppressing-reports-in-external-libraries
+# for the AddressSanitizer suppression file format.
```
lancesix, line 3: note that GDB testsuite already sets detect_leaks=0 (https://github.com/ROCm/ROCgdb/blob/amd-staging/gdb/testsuite/lib/gdb.exp#L109-L114).  Should this be re-stated here just for clarity?

.github/scripts/test_rocgdb.py
```diff
@@ -79,30 +46,46 @@
...
+EMULATOR_DEFAULT_WALL_CLOCK_TIMEOUT = 3600
+
+# Sanitizers we pre-configure on every run. If rocgdb wasn't built with the
+# corresponding -fsanitize=..., the matching *_OPTIONS env var is silently
+# ignored at runtime, so listing extras here is harmless.
+SANITIZERS = ("asan", "tsan", "ubsan", "msan", "lsan", "hwasan")
+
+# Patterns for parsing DejaGnu .sum lines.
+RESULT_LINE_RE = re.compile(
+    r"^(PASS|FAIL|XFAIL|UNTESTED|UNSUPPORTED|KFAIL|UNRESOLVED): (.+)"
```
lancesix: should there be a `$` marker at the end of the expression?

.github/scripts/test_rocgdb.py
```diff
@@ -124,13 +109,63 @@ def _run_command(
...
+            else:
+                _log_error_and_exit(f"Failed to execute {cmd[0]!r}: {e}")
+        try:
+            out, err = process.communicate(timeout=timeout)
+        except BaseException:
+            # Cover TimeoutExpired and any other escaping exception
+            # (KeyboardInterrupt, OSError mid-read, etc.) so the make /
+            # runtest / expect / gdb process group is never orphaned.
+            try:
+                os.killpg(process.pid, signal.SIGKILL)
```
lancesix: It it different from doing `process.kill()`?

.github/scripts/test_rocgdb.py
```diff
@@ -124,13 +109,63 @@ def _run_command(
...
+                out, err = None, None
+            raise
+        result = subprocess.CompletedProcess(cmd, process.returncode, out, err)
+        if check and process.returncode != 0:
+            exc = subprocess.CalledProcessError(process.returncode, cmd, out, err)
+            if error_msg:
+                _log_error_and_exit(f"{error_msg}: {exc}")
+            raise exc
+        return result
+
```
lancesix: Feels that most of this section could always be used, just with `start_new_session=kill_process_group`.  Wouldn't this replace the `subprocess.run` you fallback to?  Would allow to only have one code path.

.github/scripts/test_rocgdb.py
```diff
@@ -246,30 +412,62 @@ def parse_arguments() -> argparse.Namespace:
...
+        "Default (if no test option specified) is gdb.rocm/*.exp and gdb.dwarf2/*.exp",
+    )
+    test_group.add_argument(
+        "--gpu-tests",
+        action="store_true",
+        help="Run GPU-specific tests (gdb.rocm directory only). "
+        "Mutually exclusive with --tests and --cpu-tests.",
     )
+    test_group.add_argument(
+        "--cpu-tests",
```
lancesix, line 429: do we plan to run the full GDB testsuite (excl gdb.rocm)?

.github/scripts/test_rocgdb.py
```diff
@@ -284,38 +482,88 @@ def parse_arguments() -> argparse.Namespace:
...
-        help="Which GDB testsuite target to invoke. 'check' is the normal run."
-        "'check-read1' runs under an LD_PRELOAD shim that forces read(2) to"
-        "return 1 byte at a time (stress-tests GDB's incremental I/O parsing)."
-        "'check-readmore' forces large reads (the opposite extreme). These"
-        "are upstream GDB testsuite modes and can surface real GDB bugs that"
+        help="Which GDB testsuite target to invoke. 'check' is the normal run. "
+        "'check-read1' runs under an LD_PRELOAD shim that forces read(2) to "
+        "return 1 byte at a time (stress-tests GDB's incremental I/O parsing). "
+        "'check-readmore' forces large reads (the opposite extreme). These "
+        "are upstream GDB testsuite modes and can surface real GDB bugs that "
```
lancesix: What this usually exposes are testcase bugs (invalid regexpr) more than GDB bugs (at least as far as I know / use those).

.github/scripts/test_rocgdb.py
```diff
@@ -284,38 +482,88 @@ def parse_arguments() -> argparse.Namespace:
...
+        help="Path to output JSON file for generating a new ignore list from test failures. "
+        "When specified, the testsuite runs normally and all failures are written to this file "
+        "at the end of the run. Common failures go into 'Generic' category, compiler-specific "
+        "failures go into compiler-specific categories (e.g., 'GCC', 'LLVM'). "
+        "This option requires a value.",
+    )
+    parser.add_argument(
+        "--emulator",
+        action="store_true",
+        help="Run each test in its own `make check` invocation and save its gdb.log "
```
lancesix, line 500: doesn't that impact timeout values as well?

.github/scripts/test_rocgdb.py
```diff
@@ -342,94 +590,88 @@ def __init__(self) -> None:
         # Track flaky tests (failed initially but passed on retry) by compiler_label.
         self.flaky_tests: Dict[str, Set[str]] = defaultdict(set)
 
-    def cleanup_old_entries(self, compiler_label: str, tests: str) -> None:
+        # Track harness-level errors (ERROR lines with no preceding
+        # "Running ....exp") per compiler_label. Replaced on each
+        # update_results call so it reflects the most recent run's state.
+        self.harness_errors: Dict[str, List[str]] = defaultdict(list)
```
lancesix, line 582: Just a suggestion, we could make the type name even clearer with typename aliases

```python
compiler = str
testname = str
self.harness_errors: Dict[compiler, List[testname]] = defaultdict(list)
```

but maybe having the dummy variables is the part for helping

.github/scripts/test_rocgdb.py
```diff
@@ -342,94 +590,88 @@ def __init__(self) -> None:
...
+            if running_match:
+                segments = os.path.normpath(running_match.group(1)).split(os.path.sep)
+                current_test_file = "/".join(segments[-2:])
+
+            if line.startswith("ERROR:"):
+                msg = line[len("ERROR:") :].strip()
+                if current_test_file is None:
+                    orphan_errors.append(msg)
+                else:
+                    new_errors[current_test_file].append(f"{current_test_file}: {msg}")
```
lancesix, line 653: seems redundant to record the `current_test_file` in each entry, isn't it?

.github/scripts/test_rocgdb.py
```diff
@@ -556,10 +806,11 @@ def _print_grouped_tests(self, test_list: Dict[str, List[str]]) -> None:
...
 
-        for file in test_list.keys():
-            for test_description in test_list[file]:
-                test_dir, test_file = os.path.split(file)
-                grouped[test_dir][test_file].append(test_description[len(file) + 2 :])
+        for file, descriptions in test_list.items():
+            test_dir, test_file = os.path.split(file)
+            prefix_len = len(file) + 2
+            for test_description in descriptions:
+                grouped[test_dir][test_file].append(test_description[prefix_len:])
```
lancesix, line 799: here we strep the file name, seems we could just not have it in the first place (see previous comment).

.github/scripts/test_rocgdb.py
```diff
@@ -783,6 +1077,35 @@ def print_section(
         logger.info(apply_color(border))
 
 
+def _log_aligned_fields(
+    fields: List[Tuple[str, Optional[str]]],
+    indent: str = "  ",
+) -> None:
+    """
+    Log (label, value) pairs with colons aligned to the longest label.
```
lancesix: longest label with an associated value.

.github/scripts/test_rocgdb.py
```diff
@@ -1026,67 +1440,93 @@ def set_test_timeout(test_suite_dir: Path, timeout_value: int) -> None:
...
-    return sorted(set(_extract_test_file_from_name(name) for name in test_names))
+    if exclude is None:
+        exclude = []
 
+    test_dirs = []
+    for item in testsuite_dir.iterdir():
+        if item.is_dir() and item.name.startswith("gdb.") and item.name not in exclude:
+            # Check if directory contains any .exp files.
+            exp_files = list(item.glob("*.exp"))
+            if exp_files:
```
lancesix: probably no need to materialize the list of files.  Could be

```python
if any(item.glob("*.exp"):
    test_dirs.appent(item.name)
```

(implicitly relies on any valid name being considered True)

otherwise, maybe more natural could be

```python
if len(item.glob("*.exp")) > 0:
```

.github/scripts/test_rocgdb.py
```diff
@@ -1182,34 +1626,39 @@ def validate_rocgdb(rocgdb_bin: Path) -> None:
...
+        # Parse "label: value" pairs printed by the embedded script and
+        # re-emit them with colons aligned, matching the surrounding
+        # banner blocks. Lines without a colon are passed through verbatim.
+        parsed: List[Tuple[str, Optional[str]]] = []
+        for line in result.stdout.splitlines():
+            label, sep, value = line.partition(":")
+            if sep:
+                parsed.append((label.strip(), value.strip()))
+            else:
+                parsed.append((line, None))
```
lancesix: given the usage here, it really feels like line is not a key

`_log_aligned_fields` argument could be `Union[str, Pair[str, str]]` to not create dummy pair with the second to None

.github/scripts/test_rocgdb.py
```diff
@@ -1236,46 +1685,83 @@ def run_tests(
         None
     """
     max_iterations = 1 + args.max_failed_retries
-    current_tests = tests
+    current_tests = list(tests)
```
lancesix: do we really need to copy the list?  (this could be necessary if tests was an iterator, but the type says otherwise).

.github/scripts/test_rocgdb.py
```diff
@@ -1236,46 +1685,83 @@ def run_tests(
...
+        else:
+            cmd = [
+                "make",
+                args.check_type,
+                f"RUNTESTFLAGS={runtestflags_str}",
+                f"TESTS={' '.join(current_tests)}",
+            ]
+            if args.parallel:
+                if args.jobs is not None:
+                    cmd += ["FORCE_PARALLEL=1", f"-j{args.jobs}"]
```
lancesix: could just be

```python
cmd += ["FORCE_PARALLEL=1", f"-j{args.jobs or ""}"]
```

to avoid this if / else.

.github/scripts/test_rocgdb.py
```diff
@@ -1409,13 +1998,15 @@ def main() -> None:
...
+    setup_sanitizer_environment(env_vars, rocgdb_testsuite_dir)
 
     # Validate that we can run rocgdb.
-    validate_rocgdb(rocgdb_bin)
+    validate_rocgdb(rocgdb_bin, env_vars)
 
     # Verify executables presence.
     check_executables(
-        ["hipcc", "gcc", "g++", "gfortran", "clang", "clang++", "flang", "runtest"]
+        ["hipcc", "gcc", "g++", "gfortran", "clang", "clang++", "flang", "runtest"],
```
lancesix: `make` should probably also be in this list.

.github/scripts/test_rocgdb.py
```diff
@@ -1501,20 +2094,196 @@ def _build_runtestflags(
...
+
+    Returns:
+        None
+
+    Exits:
+        Code 1 if the aggregated gdb.sum cannot be written.
+    """
+    aggregated_lines: List[str] = []
+    base_suffix = "" if iteration == 1 else f".retry-{iteration - 1}"
+    retry_number = iteration - 1
```
lancesix, line 2146: can swap and re-use

```python
retry_number = iteration - 1
base_suffix = "" if iteration == 1 else f".retry-{retry_number}"
```

.github/scripts/test_rocgdb.py
```diff
@@ -1501,20 +2094,196 @@ def _build_runtestflags(
...
+
+    for test in tests:
+        cmd = [
+            "make",
+            check_type,
+            f"RUNTESTFLAGS={runtestflags_str}",
+            f"TESTS={test}",
+        ]
+        logger.info(f"[emulator] {compiler_label} - {test}: {shlex.join(cmd)}")
```
lancesix: I know this is only called in emulator mode, but does not feel emulator specific.  One could use this even outside of emulation, for example to rule out leftover TCL variable that could influence a following test, or stuff like this.

.github/scripts/test_rocgdb.py
```diff
@@ -1501,20 +2094,196 @@ def _build_runtestflags(
...
+            )
+
+        aggregated_lines.extend(per_test_lines)
+
+        # Determine pass/fail using the same definition as TestResults, and
+        # pick the most informative status label by scanning for the highest
+        # severity prefix present (TIMEOUT > ERROR > UNRESOLVED > FAIL > PASS).
+        has_error = any(line.startswith("ERROR:") for line in per_test_lines)
+        has_unresolved = any(line.startswith("UNRESOLVED:") for line in per_test_lines)
+        has_fail = any(line.startswith("FAIL:") for line in per_test_lines)
```
lancesix: feels like we could avoid iterating 3 times on `per_test_lines`.  Probably not the end of the world, but feels odd to read.

.github/scripts/test_rocgdb.py
```diff
@@ -1501,20 +2094,196 @@ def _build_runtestflags(
...
+            )
+
+        status_content = (
+            f"test: {test}\n"
+            f"compiler: {compiler_label}\n"
+            f"iteration: {iteration}\n"
+            f"retry_number: {retry_number}\n"
+            f"status: {status_text}\n"
+            f"duration_seconds: {duration:.3f}\n"
+            f"wall_clock_timeout: {'true' if wall_clock_timed_out else 'false'}\n"
```
lancesix:
```python
f"wall_clock_timeout: {bool(wall_clock_timed_out)}\n"
```

.github/scripts/test_rocgdb.py
```diff
@@ -1581,6 +2357,10 @@ def print_configuration(
             Path to the root of the ROCm tree.
         args (argparse.Namespace):
             Parsed command-line arguments containing additional configuration values.
+        ignore_list_file_status (str):
```
lancesix: style seems inconsistent.  The type is given in the type annotation, having it here is redundant, and many places do not have type in the docstr.

.github/scripts/test_rocgdb.py
```diff
@@ -1591,27 +2371,59 @@ def print_configuration(
...
-    logger.info(f"  Tests:                {' '.join(args.tests)}")
-    logger.info(f"  Check Type:           {args.check_type}")
-    logger.info(f"  Parallel Execution:   {'Enabled' if args.parallel else 'Disabled'}")
-    logger.info(f"  Use FAIL ignore list: {'Not using' if args.no_xfail else 'Using'}")
-    logger.info(
-        f"  Group Results:        {'Enabled' if args.group_results else 'Disabled'}"
+    if not args.parallel:
+        parallel_info = "Disabled"
+    elif args.jobs is None:
+        parallel_info = "Enabled (unlimited)"
```
lancesix: just noting here, but I am not sure a bare `-j` (unlimited) is a sensitive thing to do in the first place.

lancesix (commented):
(I know I am not currently assigned, I still had a crack at it)

I have left a few comments, mostly minor.

Overall remarks on the emulator mode: I see it can be used in unstable environment, but what it does (run tests one at a time, and timeout) does not seem emulation specific.  It feels it could just be referred as "one-by-one" mode.

.github/scripts/sanitizers/asan.supp
lumachad: Given the script forces that setting, I think we're good. The suppression file is intentionally empty for now — a place to add patterns if/when needed. I do recall getting the warnings for leaks even with gdb's testsuite forcing that setting though. We might want to revisit that.

.github/scripts/test_rocgdb.py
lumachad: Good catch. Added.

.github/scripts/test_rocgdb.py
lumachad: Yes — `process.kill()` signals only the process itself, while `os.killpg` kills the entire process group. Since `make` spawns `runtest` → `expect` → `gdb`, we need to kill the whole group to avoid leaving orphaned processes behind on timeout.

.github/scripts/test_rocgdb.py
lumachad: Keeping the two paths — `start_new_session=True` changes signal propagation for all callers, not just the long-running `make check` ones.

.github/scripts/test_rocgdb.py
lumachad: Yes, the intent is to run all non-GPU tests (every gdb.* directory with .exp files except gdb.rocm). This should be the case when we start validating upstream gdb.

.github/scripts/test_rocgdb.py
lumachad: Fair point. Updated the help text to mention testcase regex bugs instead.

.github/scripts/test_rocgdb.py
lumachad: Yes, each individual `make check` invocation gets its own wall-clock timeout. Added a cross-reference to `--one-by-one-test-timeout` in the help text. Also took the opportunity to rename `--emulator` to `--one-by-one` since the behaviour isn't emulation-specific. Does this make it better?

.github/scripts/test_rocgdb.py
lumachad: I'm not sure how much improvement this would give us. The aliases would need to be module-wide. I'd rather keep it like this for now.

.github/scripts/test_rocgdb.py
lumachad: The prefix is actually used in the flat (non-grouped) display path where entries are printed as-is with no surrounding context. Fixing it properly would require changes to the print logic too, so I'd rather leave it for now.

.github/scripts/test_rocgdb.py
lumachad: The prefix is actually used in the flat (non-grouped) display path where entries are printed as-is with no surrounding context. Fixing it properly would require changes to the print logic too, so I'd rather leave it for now.

.github/scripts/test_rocgdb.py
lumachad: Fixed.

.github/scripts/test_rocgdb.py
lumachad: Done.

.github/scripts/test_rocgdb.py
lumachad: Done. Changed the signature to `Union[str, Tuple[str, str]]` and updated the one call site that used the `None` sentinel.

.github/scripts/test_rocgdb.py
lumachad: Done. Since `current_tests` is only ever reassigned (never mutated in place), the copy wasn't needed.

.github/scripts/test_rocgdb.py
lumachad: Done, and also changed the behaviour so that omitting `-j` defaults to `os.cpu_count()` rather than unlimited parallelism.

.github/scripts/test_rocgdb.py
lumachad: Done.

.github/scripts/test_rocgdb.py
lumachad: Done.

.github/scripts/test_rocgdb.py
lumachad: Good point. Renamed `--emulator` to `--one-by-one` throughout (flag, log dir, timeout option, constant, and all references). Does that work?

.github/scripts/test_rocgdb.py
lumachad: Done. Combined the three passes into a single loop.

.github/scripts/test_rocgdb.py
lumachad: Done.

.github/scripts/test_rocgdb.py
lumachad: Done. Removed the redundant type annotations from the Args section.

.github/scripts/test_rocgdb.py
lumachad: Good point. Changed the default to `os.cpu_count()` when `--jobs` is not specified, so we never pass bare `-j`.

lumachad:
Addressed all review comments in a single updated commit. Summary of changes:

**Renaming**
- `--emulator` / `--emulator-log-dir` / `--emulator-test-timeout` → `--one-by-one` / `--one-by-one-log-dir` / `--one-by-one-test-timeout` (the mode is not emulator-specific)
- `EMULATOR_DEFAULT_WALL_CLOCK_TIMEOUT` → `ONE_BY_ONE_DEFAULT_WALL_CLOCK_TIMEOUT`

**Bug fixes**
- Anchor `RESULT_LINE_RE` with `^` and `$` to prevent spurious partial matches

**API / behaviour changes**
- `_log_aligned_fields` now accepts `Union[str, Tuple[str, str]]` so callers can pass plain pass-through strings alongside `(label, value)` pairs
- `--parallel` defaults to `os.cpu_count()` jobs when `--jobs` is not given, instead of passing bare `-j` (unlimited parallelism)

**Minor cleanups**
- Add `make` to the executable presence check at startup
- Simplify `per_test_lines` status scan into a single pass
- Simplify `Path.glob` existence check with `any()`
- Remove unnecessary `current_tests` list copy
- Swap `retry_number` / `base_suffix` computation order for clarity
- Use `bool()` when writing `wall_clock_timeout` to the status file
- Remove `(Type)` annotations from `print_configuration` docstring
- Update `--check-type` help text to mention testcase regex bugs

Also set argparse help formatter width to 100 columns (by @lumachad, in addition to the above).

lumachad:
Labelling ci:skip, as we're missing the piece that adds the xfails file. Needs to come from TheRock.

lumachad:
Back to a review, since we have TheRock's fix.

.github/scripts/test_rocgdb.py
```diff
@@ -783,6 +1080,35 @@ def print_section(
...
+    Args:
+        fields: List of (label, value) tuples or plain strings for pass-through lines.
+        indent: Leading whitespace prefix for every row.
+
+    Returns:
+        None
+    """
+    if not fields:
+        return
+    labelled = [label for entry in fields if isinstance(entry, tuple) for label, _ in [entry]]
```
lancesix: I find this one quite odd to read, especially the `for label, _ in [entry]` part.  Why create 1 element list just to iterate over it, just to have unpacking?

I would find this much more readable:

```python
labelled = [entry[0] for entry in fields if isinstance(entry, tuple)]
```

.github/scripts/test_rocgdb.py
```diff
@@ -1409,13 +1996,15 @@ def main() -> None:
...
+    setup_sanitizer_environment(env_vars, rocgdb_testsuite_dir)
 
     # Validate that we can run rocgdb.
-    validate_rocgdb(rocgdb_bin)
+    validate_rocgdb(rocgdb_bin, env_vars)
 
     # Verify executables presence.
     check_executables(
-        ["hipcc", "gcc", "g++", "gfortran", "clang", "clang++", "flang", "runtest"]
+        ["make", "hipcc", "gcc", "g++", "gfortran", "clang", "clang++", "flang", "runtest"],
```
lancesix: I think there are changes pending (#62) to use `amdclang++`.  Should we include it already?

.github/scripts/test_rocgdb.py
```diff
@@ -1501,20 +2092,201 @@ def _build_runtestflags(
...
-        f"GDB={rocgdb_bin}",
-        f"CC_FOR_TARGET={cc}",
-        f"CXX_FOR_TARGET={cxx}",
-        f"F77_FOR_TARGET={fc}",
-        f"F90_FOR_TARGET={fc}",
+        f"GDB={shlex.quote(str(rocgdb_bin))}",
+        f"CC_FOR_TARGET={shlex.quote(cc)}",
+        f"CXX_FOR_TARGET={shlex.quote(cxx)}",
+        f"F77_FOR_TARGET={shlex.quote(fc)}",
+        f"F90_FOR_TARGET={shlex.quote(fc)}",
```
lancesix, line 2093: If we are explicitly specifying compilers, should we also include

`HIPCC_FOR_TARGET` (will soon be legacy) and `HIP_COMPILER_FOR_TARGET` (`amdclang++`, soon to be used)?

.github/scripts/test_rocgdb.py
lumachad: Probably.

.github/scripts/test_rocgdb.py
lumachad: Probably.

.github/scripts/test_rocgdb.py
```diff
@@ -1240,42 +1690,76 @@ def run_tests(
...
-            cmd,
-            cwd=test_suite_dir,
-            env=env_vars,
-            capture_output=args.quiet,
-            check=False,
-        )
+            logger.info(
+                f"Executing tests with {compiler_label} - Iteration {iteration}: {shlex.join(cmd)}"
+            )
+            _run_command(
```
lancesix, line 1739: Could use `kill_process_group=True` so all `make` invocations run in the same way (one-by-one or batch).

That, or this could be an opportunity to simplify `_run_command` and split it in 2: `_run_make_command` (which would have the `kill_process_group=True`), and `_run_command` which uses the simpler path.

.github/scripts/test_rocgdb.py
lumachad: Done. Unified into a single `Popen` path with `start_new_session=kill_process_group`. Exception cleanup branches on the flag: `os.killpg` for new sessions, `process.kill()` otherwise.

.github/scripts/test_rocgdb.py
lumachad: Done. Simplified to `[entry[0] for entry in fields if isinstance(entry, tuple)]`.

.github/scripts/test_rocgdb.py
lumachad: Done. Replaced `hipcc` with `amdclang++` now that #62 has landed.

.github/scripts/test_rocgdb.py
lumachad: Done. Added `HIP_COMPILER_FOR_TARGET=amdclang++` to RUNTESTFLAGS now that #62 has landed. `HIPCC_FOR_TARGET` is gone from `future.exp` so no need to set it. Unless you think we should add `HIPCC_FOR_TARGET` still.

.github/scripts/test_rocgdb.py
lumachad: Done. Added `kill_process_group=True` to the batch `make check` call so both paths tear down the process tree consistently.

.github/scripts/test_rocgdb.py
lancesix: I think this is fine.

We can still use this script to run upstream's testsuite, which still uses hipcc (for now, next step is to upstream the transition to amdclang++).  But I doubt this is worth the effort.  We can make the script for branches as they'll be in a week or so.

lumachad: Worth mentioning we might not be able to pick up the changes from #62 if they conflict with amd-staging-rocgdb-16.

lancesix: Part of the cherry-pick process includes adjusting the patch as necessary.

lancesix (approved):
Ok, should be good to go, if CI runs fine.

PR #123, opened 2026-05-13
gdb, amd_dbgapi_target: revise store_registers to handle regno == -1

aktemur:
This is not submitted to upstream master because there the `fetch_reg` lambda patch (9e6a95ff9f29) is not available. Alternatively we can submit both to upstream. @lancesix

### DIFF
--- a/gdb/amd-dbgapi-target.c
+++ b/gdb/amd-dbgapi-target.c
@@ -2754,38 +2754,57 @@ amd_dbgapi_target::store_registers (struct regcache *regcache, int regno)
   struct gdbarch *gdbarch = regcache->arch ();
   gdb_assert (is_amdgpu_arch (gdbarch));
 
-  gdb_byte raw[AMDGPU_MAX_REGISTER_SIZE];
-  ULONGEST reg_size = register_type (gdbarch, regno)->length ();
-  gdb_assert (reg_size <= AMDGPU_MAX_REGISTER_SIZE);
-  regcache->raw_collect (regno, &raw);
-
   amdgpu_gdbarch_tdep *tdep = get_amdgpu_gdbarch_tdep (gdbarch);
+  amd_dbgapi_wave_id_t wave_id = get_amd_dbgapi_wave_id (regcache->ptid ());
 
-  /* If the register has read-only bits, invalidate the value in the regcache
-     as the value actually written may differ.  */
-  if (tdep->register_properties[regno]
-      & AMD_DBGAPI_REGISTER_PROPERTY_READONLY_BITS)
-    regcache->invalidate (regno);
+  bool invalidate_volatile = false;
 
-  /* Invalidate all volatile registers if this register has the invalidate
-     volatile property.  For example, writing to VCC may change the content
-     of STATUS.VCCZ.  */
-  if (tdep->register_properties[regno]
-      & AMD_DBGAPI_REGISTER_PROPERTY_INVALIDATE_VOLATILE)
+  auto store_reg
+    = [wave_id, tdep, gdbarch, regcache, &invalidate_volatile] (int reg)
+      {
+	gdb_byte raw[AMDGPU_MAX_REGISTER_SIZE];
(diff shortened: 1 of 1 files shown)


gdb/amd-dbgapi-target.c
```diff
@@ -2751,38 +2751,51 @@ amd_dbgapi_target::store_registers (struct regcache *regcache, int regno)
...
-    = amd_dbgapi_write_register (wave_id, tdep->register_ids[regno], 0,
-				 reg_size, raw);
+	amd_dbgapi_status_t status
+	  = amd_dbgapi_write_register (wave_id, tdep->register_ids[reg], 0,
+				       reg_size, raw);
 
-  if (status != AMD_DBGAPI_STATUS_SUCCESS)
-    warning (_("Couldn't write register %s (#%d)."),
-	     gdbarch_register_name (gdbarch, regno), regno);
+	if (status != AMD_DBGAPI_STATUS_SUCCESS && warn)
```
palves: Did you consider moving the warning to the caller, avoiding the "mode" boolean?

Like:

```
  if (regno == -1)
    {
      for (int i = 0; i < gdbarch_num_regs (gdbarch); i++)
	store_reg (i);
    }
  else
   {
      if (!store_reg (regno))
        warning (_("Couldn't write register %s (#%d)."),
	       gdbarch_register_name (gdbarch, regno), regno);
   }
```

aktemur: > Did you consider moving the warning to the caller, avoiding the "mode" boolean?

I followed `fetch_register`, but I agree this would look cleaner. I'll add a patch to modify `fetch_reg` first.

gdb/amd-dbgapi-target.c
```diff
@@ -2754,38 +2754,53 @@ amd_dbgapi_target::store_registers (struct regcache *regcache, int regno)
...
+	/* Invalidate all volatile registers if this register has the
+	   invalidate volatile property.  For example, writing to VCC
+	   may change the content of STATUS.VCCZ.  */
+	if (tdep->register_properties[reg]
+	    & AMD_DBGAPI_REGISTER_PROPERTY_INVALIDATE_VOLATILE)
+	  {
+	    for (size_t r = 0; r < tdep->register_properties.size (); ++r)
+	      if (tdep->register_properties[r]
+		  & AMD_DBGAPI_REGISTER_PROPERTY_VOLATILE)
+		regcache->invalidate (r);
```
palves: I think we could take the opportunity to handle the INVALIDATE_VOLATILE case better, when writing all registers.

Inside the outer loop, if register ```i``` has INVALIDATE_VOLATILE, we invalidate every volatile register.  This is quadratic behavior that we could avoid if we do the invalidation loop just once.   

But in addition, if ```i + N``` is volatile, its ```store_reg (i + N)``` turn hits ```regcache->raw_collect (i + N, ...)``` on a register that was just invalidated, which will roundtrip through the target to re-read what we're about to overwrite.  That makes that write be "re-read from target, then write back the just-read value" AFAICS.

Seems to me like ideally we'd do all the writes first in one pass, and then if that wrote any INVALIDATE_VOLATILE register, do the volatile invalidation loop as a second pass.

aktemur: > Seems to me like ideally we'd do all the writes first in one pass, and then if that wrote any INVALIDATE_VOLATILE register, do the volatile invalidation loop as a second pass.

Agree.  I also assume storing incompatible values for volatile/read-only registers is checked by the debug api, btw. Because otherwise one could end up with value inconsistencies, regardless of when we invalidate.

palves (approved):
LGTM, thanks.

lumachad:
I'm seeing a few failures in the CI. Not sure if runner-related or something else.

lumachad:
Re-triggered and it looks good now.

aktemur:
> Re-triggered and it looks good now.

Thanks for checking.

PR #124, opened 2026-05-13
WIP: Lane divergence support

palves:
Lane divergence support prototype.

This prototypes support for lane divergence using DW_AT_LLVM_lane_pc from https://llvm.org/docs/AMDGPUUsage.html#dw-at-llvm-lane-pc .

It includes a testcase hand-written using the DWARF assembler from the testsuite.  It's actually the first example of a DWARF-assembler based GPU testcase, showing the contortions we need to go through.  Those could could/should probably be extracted out making it possible to write other DWARF-assembler based tests using that framework.


palves:
Currently based on amd-staging from 2024.  Needs to be properly rebased.  I've made it work with current compilers though.  Good enough for UI discussions.

aktemur:
@palves, given the purpose of this branch, I'm converting this PR to a draft. Please update if this change is incorrect.

PR #131, opened 2026-05-19
gdb/testsuite/gdb.rocm: add multi-inferior stress test

spatrang:
## Summary

Add a stress test that debugs many GPU programs on a single GPU at
once, to exercise the multi-process GPU debug path and GDB's
multi-inferior plumbing under load.

## What the test does

- Creates N inferiors of a simple kernel-launcher program with
  `add-inferior` (not fork), all pinned to the same (default) GPU, and
  runs each to `main`.
- Plants a breakpoint on the kernel, resumes everything in non-stop
  mode, waits for one kernel stop per inferior, then continues them all
  to a clean exit.
- Running enough programs on one device crosses the per-device limit on
  simultaneously-debugged processes and exercises the firmware context
  switching that kicks in beyond it.

## Notes

- Uses `add-inferior` rather than fork, so the test does not depend on
  fork being available.
- Deliberately runs everything on the default GPU; saturating one
  device is what exercises the per-device limit, and it keeps the test
  shard-friendly (documented in code comments).
- N is chosen to exceed the per-device limit; the test sweeps N/2 and
  N*2 around it.  The timeout scales with N, since bringing up many GPU
  runtimes on one device serializes.

## Gating

Skipped automatically on targets that do not support multi-process GPU
debugging, via `hip_devices_support_debug_multi_process`, plus
`allow_hip_tests` and `allow_multi_inferior_tests`.

## Files added

- `gdb/testsuite/gdb.rocm/multi-inferior-stress.cpp`
- `gdb/testsuite/gdb.rocm/multi-inferior-stress.exp`

### DIFF
--- a/gdb/testsuite/gdb.rocm/multi-inferior-stress.cpp
+++ b/gdb/testsuite/gdb.rocm/multi-inferior-stress.cpp
@@ -1,0 +1,38 @@
+/* This testcase is part of GDB, the GNU debugger.
+
+   Copyright (C) 2026 Free Software Foundation, Inc.
+   Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.  */
+
+/* GPU kernel launcher.  Launches a kernel on the default GPU and waits
+   for it to finish.  */
+
+#include <hip/hip_runtime.h>
+
+#include "rocm-test-utils.h"
+
+__global__ void
+kern ()
+{
+  __builtin_amdgcn_s_sleep (1);
+}
+
+int
+main ()
+{
+  kern<<<1, 1>>> ();
+  CHECK (hipDeviceSynchronize ());
+  return 0;
+}
--- a/gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
+++ b/gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
@@ -1,0 +1,161 @@
+# Copyright (C) 2026 Free Software Foundation, Inc.
+# Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+# This file is part of GDB.
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Stress test debugging many GPU programs on a single GPU at once.
+#
+# GDB creates N inferiors of a simple kernel-launcher program, all
+# pinned to the same GPU device, runs them concurrently in non-stop
+# mode, and drives every one to a kernel breakpoint and then to exit.
+# Running enough programs on one device exercises the multi-process GPU
+# debug path, including the firmware context switching that kicks in
+# once the number of simultaneously debugged processes crosses the
+# hardware limit, as well as GDB's multi-inferior plumbing.
+#
+# Inferiors are created with "add-inferior" rather than fork, so the
+# test does not depend on fork being available.
+
+load_lib rocm.exp
+
+standard_testfile .cpp
+
+require allow_hip_tests
+require hip_devices_support_debug_multi_process
+require allow_multi_inferior_tests
+
+if {[build_executable "failed to prepare" $testfile $srcfile {debug hip}]} {
+    return
+}
+
+# Run one stress iteration with NUM_INFERIORS inferiors, all on a
+# single GPU.
+
+proc do_test { num_inferiors } {
+    clean_restart
+    gdb_load $::binfile
+
+    gdb_test_no_output "set non-stop on"
+
+    with_rocm_gpu_lock {
+	# Create NUM_INFERIORS inferiors of the same program and run
+	# each to main.  Every inferior uses the default device (device
+	# 0), which is the first ROCR_VISIBLE_DEVICES-visible GPU, so
+	# they all run on the same single GPU.  Running that many
+	# programs on one device is what saturates it past the per-device
+	# limit and exercises the firmware context switching.
+	for {set i 1} {$i <= $num_inferiors} {incr i} {
+	    if { $i > 1 } {
+		gdb_test "add-inferior" "Added inferior $i.*" \
+		    "add inferior $i"
+		gdb_test "inferior $i" "Switching to inferior $i.*" \
+		    "switch to inferior $i"
+		gdb_load $::binfile
+	    }
+	    gdb_test "start" "Temporary breakpoint $::decimal, main .*" \
+		"start inferior $i"
+	}
+
+	# Plant a breakpoint on the kernel and resume every inferior.
+	gdb_breakpoint kern -allow-pending
+
+	gdb_test_multiple "continue -a &" "continue -a to kernel" {
+	    -re "Continuing\\.\r\n$::gdb_prompt " {
+		pass $gdb_test_name
+	    }
+	}
+
+	# Collect one kernel stop per inferior.
+	set seen [list]
+	set test "wait for all GPU stops"
+	gdb_test_multiple "" $test {
+	    -re "Thread ($::decimal)\\.$::decimal\[^\r\n\]* hit Breakpoint\[^\r\n\]*, kern \\(\\)\[^\r\n\]*\r\n" {
+		set inf $expect_out(1,string)
+		if { [lsearch -exact $seen $inf] == -1 } {
+		    lappend seen $inf
+		}
+		if { [llength $seen] < $num_inferiors } {
+		    exp_continue
+		} else {
+		    pass $test
+		}
+	    }
+	    -re "\\\[Inferior $::decimal \[^\r\n\]* exited\[^\r\n\]*\\\]\r\n" {
+		fail "$test (inferior exited before hitting the kernel)"
+		return
+	    }
+	    timeout {
+		fail $test
+		verbose -log "only [llength $seen] of $num_inferiors\
+			inferiors stopped before timeout"
+		return
+	    }
+	}
+
+	# Resume everything again.  Every inferior should run to a clean
+	# exit.
+	gdb_test_multiple "continue -a &" "continue -a to exit" {
+	    -re "Continuing\\.\r\n$::gdb_prompt " {
+		pass $gdb_test_name
+	    }
+	}
+
+	set exited [list]
+	set test "wait for all inferiors to exit"
+	gdb_test_multiple "" $test {
+	    -re "\\\[Inferior ($::decimal) \[^\r\n\]* exited normally\\\]\r\n" {
+		set inf $expect_out(1,string)
+		if { [lsearch -exact $exited $inf] == -1 } {
+		    lappend exited $inf
+		}
+		if { [llength $exited] < $num_inferiors } {
+		    exp_continue
+		} else {
+		    pass $test
+		}
+	    }
+	    -re "\\\[Inferior $::decimal \[^\r\n\]* exited with code\[^\r\n\]*\\\]\r\n" {
+		fail "$test (non-zero exit)"
+		return
+	    }
+	    timeout {
+		fail $test
+		verbose -log "only [llength $exited] of $num_inferiors\
+			inferiors exited before timeout"
+		return
+	    }
+	}
+    }
+}
+
+# The hardware allows a limited number of processes (around 8 to 16) to
+# be debugged simultaneously on a GPU.  Beyond that the firmware does
+# transparent context switching.  Use N as a count that reaches that
+# limit and run N/2 and N*2 processes: N*2 is comfortably above the
+# limit and drives the context switching, while N/2 is a lighter
+# baseline.  N is kept modest to keep the test practical.  Empirically
+# higher counts work too but mainly cost wall-clock, as bringing up that
+# many GPU runtimes on one device serializes.
+set n 16
+
+foreach_with_prefix num_inferiors [list [expr { $n / 2 }] [expr { $n * 2 }]] {
+    # Bringing up many GPU runtimes concurrently on one device
+    # serializes, so scale the timeout to roughly one unit per 8
+    # inferiors.
+    with_timeout_factor [expr { ($num_inferiors + 7) / 8 }] {
+	do_test $num_inferiors
+    }
+}


gdb/testsuite/gdb.rocm/multi-inferior-stress.cpp
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+  CHECK (hipSetDevice (idx % num_devices));
+  kern<<<1, 1>>> ();
+  CHECK (hipDeviceSynchronize ());
+  return 0;
+}
+
+static int
+parent (char **argv, int num_children)
```
lancesix: Having `char **argv` but not `int argc` really seems odd, I would classify this as a code-smell.

My understanding is that the caller checked that there is at least one arg, so `argv[0]` is valid (the only one used here), but if this is the intent you should just pass `argv[0]` as an arg to `parent`, not the full `argv`.

gdb/testsuite/gdb.rocm/multi-inferior-stress.cpp
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+  /* Break here.  */
+
+  for (int i = 0; i < num_children; i++)
+    {
+      char idx_buf[32] = {};
+      char ndev_buf[32] = {};
+      snprintf (idx_buf, sizeof (idx_buf), "%d", i);
+      snprintf (ndev_buf, sizeof (ndev_buf), "%d", num_devices);
```
lancesix: if you use snprintf, you should check the returned value.

Could also use `std::to_string` instead (from `<string>`).

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
```diff
@@ -0,0 +1,186 @@
...
+
+    gdb_test_no_output "set non-stop on"
+    gdb_test_no_output "set detach-on-fork off"
+    gdb_test_no_output "set follow-fork parent"
+    gdb_test_no_output "set args $num_children"
+
+    # Scale the timeout with N: every additional debugged process adds
+    # debug-attach setup cost, and once we cross the per-device limit
+    # we also pay for context switching.
+    set timeout_factor [expr { ($num_children + 3) / 4 }]
```
lancesix: Has that proven to be necessary?  I expect the default timeout is order of magnitude bigger than the time we actually need.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
```diff
@@ -0,0 +1,186 @@
...
+	    gdb_breakpoint [gdb_get_line_number "Break here"]
+	    gdb_breakpoint kern allow-pending
+	    gdb_breakpoint [gdb_get_line_number "Last break here"]
+
+	    # Run the parent up to the pre-fork sync point.  GDB prints the
+	    # source line of the stop on a second line, so the pattern must
+	    # account for breakpoint-hit + source-line before the prompt.
+	    set test "run to fork point"
+	    gdb_test_multiple "run" $test {
+		-re -wrap "At least one agent is busy.*" {
```
lancesix: Can this happen with `require hip_devices_support_debug_multi_process`?

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
```diff
@@ -0,0 +1,186 @@
...
+	    # each launches its kernel and we expect one
+	    # "hit Breakpoint ..., kern ()" notification per child
+	    # inferior.
+	    gdb_test_multiple "continue -a &" "continue -a in non-stop" {
+		-re "Continuing\\.\r\n$::gdb_prompt " {
+		    pass $gdb_test_name
+		}
+	    }
+
+	    set test "wait for $num_children GPU stops"
```
lancesix: I would probably not have "$num_children" in the test, we already have that in the "with_test_with_prefix".  Could be "wait for all GPU stops".

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
```diff
@@ -0,0 +1,186 @@
...
+			lappend seen_inferiors $inf
+			lappend stopped_gpu_threads $thr
+		    }
+		    if {[llength $seen_inferiors] < $num_children} {
+			exp_continue
+		    } else {
+			pass $test
+		    }
+		}
+		-re "At least one agent is busy" {
```
lancesix: Same, can we have that if we ensured `hip_devices_support_debug_multi_process`?

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
```diff
@@ -0,0 +1,186 @@
...
+	    }
+
+	    # Diagnostic check: GDB should know about N child inferiors
+	    # plus the parent.  A mismatch here usually means a fork was
+	    # silently lost.  Each row in `info inferiors` contains a
+	    # "<num>... process <pid>" substring that the header line does
+	    # not, so we count rows by matching that substring.
+	    set inf_count 0
+	    gdb_test_multiple "info inferiors" "info inferiors shows N+1" {
+		-re "\[0-9\]+\\s+process\\s+\[0-9\]+\[^\r\n\]*\r\n" {
```
lancesix: I am not sure the "<num>... process <pid>" is target (linux) dependant.  Should we have something more generic?

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
```diff
@@ -0,0 +1,186 @@
...
+		    incr inf_count
+		    exp_continue
+		}
+		-re "$::gdb_prompt $" {
+		    set expected [expr { $num_children + 1 }]
+		    if { $inf_count >= $expected } {
```
lancesix: is the ">" meant to be a pass?  Sounds this should be a fail unless the inferior count exactly matches the expected count.

lancesix (commented):
I have not re-read that test, but this new test seems to have a lot of similarities with `gdb.rocm/multi-inferior-gpu.exp`.  Couldn't both be merged as a fair amount is common?

gdb/testsuite/gdb.rocm/multi-inferior-stress.cpp
```diff
@@ -0,0 +1,142 @@
+/* This testcase is part of GDB, the GNU debugger.
```
aktemur, line 1: Could you please give a description in the commit message? It's currently empty.

spatrang: The follow-up commit 707b1191da9 ("gdb/testsuite/gdb.rocm: clean up multi-inferior-stress test") has a bulleted body describing the changes it makes.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
spatrang: Tightened to == in 707b1191da9. With the corrected regex, > would imply either a regex defect (matching something it shouldn't) or GDB reporting more inferiors than were forked

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
spatrang: Switched to a target-agnostic column-position regex in 707b1191da9: data rows always start with * (selected) or (other) followed by the inferior number, while the header line and any async output do not match that anchor. No longer depends on target-specific words like "process".

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
spatrang: Applying the same decision as the first arm — removed in 707b1191da9.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
spatrang: `foreach_with_prefix` already adds `num_children=N:`. Renamed to `"wait for all GPU stops"` in 707b1191da9.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
spatrang: Right - at that point only inferior 1 is attached, so that message can't appear. Arm removed in 707b1191da9.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
spatrang: Agreed — the scaling was defensive without measurement. Dropped in 707b1191da9. The current sweep runs comfortably under the default gdb_test_timeout.

gdb/testsuite/gdb.rocm/multi-inferior-stress.cpp
spatrang: Switched to `std::to_string from <string>` in 707b1191da9. That drops both snprintf calls and the scratch buffers entirely - cleaner than threading return-value checks through.

gdb/testsuite/gdb.rocm/multi-inferior-stress.cpp
spatrang: Good catch. Fixed in 707b1191da9: signature is now parent (const char *argv0, int num_children), with main passing argv[0] directly.

spatrang:
> I have not re-read that test, but this new test seems to have a lot of similarities with `gdb.rocm/multi-inferior-gpu.exp`. Couldn't both be merged as a fair amount is common?

I'd like to keep the two .exp files separate: multi-inferior-gpu.exp is a per-GPU smoke (one child per detected device), while this test deliberately scales N independently of the device count to stress the multi-process GPU debug path and GDB's multi-inferior plumbing. They also have different require gates - the stress test additionally requires allow_multi_inferior_tests. If the duplication concern is the driver loop itself (set non-stop / detach-on-fork / follow-fork → run → wait-for-N hits → continue-each → post-waitpid breakpoint), I'm happy to factor that into a helper in lib/rocm.exp in a follow-up so both .exp files share it.

lumachad:
I'm in favor of having the common part split out and reused. I think we should keep the tests separate for load-balancing/concurrency purposes.

lumachad:
@spatrang Could we please have a PR that splits out the common parts that will be reused by this new test, shared with the existing multi-inferior tests? Let's do that first, then we can come back to this one and clear see the differences.

palves:
I would be better for the long term if this didn't rely on fork.  E.g., instead, add inferiors with "add-inferior" and run them.

The reason is that a fork approach won't ever work on Windows.  GDB on Windows does not support debugging multiple processes currently, but it's a matter of time until it does.  And when it does, we'd want the test to work there too.

Or even, to make it work today on Windows too, we could launch one GDB per inferior.  Doable if you juggle spawn_ids.  Might give us more coverage if twe exercise both approaches, even.  But we could start with one only if it's too complicated.

lumachad:
> I would be better for the long term if this didn't rely on fork. E.g., instead, add inferiors with "add-inferior" and run them.

I think that's a fair requirement, though I think we should put together the testcase as-is for now, and deal with that refactoring at a later stage. Unless we want to adjust the scope of the ticket this is referencing.

spatrang:
Force-pushed a reworked version addressing the review/JIRA discussion. What changed since the last push:

- **No more fork.** The test now creates the inferiors with `add-inferior` + `start` instead of fork / `detach-on-fork off` / `follow-fork parent`, so it doesn't depend on fork being available.
- **Single GPU saturation.** All inferiors now run on the default GPU instead of being spread across devices with `hipSetDevice(idx % num_devices)`. Saturating one device is what actually exercises the per-device limit and the firmware context switching. This is deliberate and documented in code comments; it also keeps the test shard-friendly.
- **N basis.** I probed how high N can go on gfx942: no hard cap up to 64 inferiors (no errors), the only cost is wall-clock since bringing up that many GPU runtimes on one device serializes. Given the per-device limit is around 8-16, the test uses N = 16 and sweeps N/2 = 8 and N*2 = 32; 32 is comfortably above the limit. The timeout scales with N.
- The `.cpp` is now a simple kernel launcher (no parent/child/waitpid).

gdb/testsuite/gdb.rocm/multi-inferior-stress.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include <hip/hip_runtime.h>
+
+#include <stdlib.h>
+
+#include "rocm-test-utils.h"
+
+__global__ void
+kern ()
+{
+  asm ("s_sleep 1");
```
czidev-amd: Most interesting results you get when something more computational heavy is happening in the kernel, maybe some MAC instructions? It is just a suggestion, you can discard it as you please.

czidev-amd (approved):
Once all the other reviewers input are resolved, I have no other objections

gdb/testsuite/gdb.rocm/multi-inferior-stress.cpp
spatrang: Thanks for the suggestion — I actually tried it. I swapped the `s_sleep` for a multiply-accumulate loop and measured it on gfx942 with the sweep at N = {8, 32}:

- `s_sleep 1`: 166 pass / 0 fail, ~24s, stable across runs.
- MAC loop (1000 iters): mostly passed but **flaky** — repeated runs gave stragglers at N=32 (`only 29-31 of 32 inferiors exited before timeout`), with wall time swinging from ~16s to ~97s.
- MAC loop (100000 iters): consistently failed the exit phase.

The root cause is exactly the contention this test creates: with N processes context-switched on one GPU by the firmware, a compute-bound kernel produces stragglers — one or two processes get starved and don't finish in time — which makes the pass/fail outcome nondeterministic. A sleeping wave doesn't fight for the compute units, so every process makes prompt forward progress and the test stays reliable.

Since this test's job is a deterministic regression check of the multi-process *debug* path (not compute throughput), I'll keep `s_sleep` for stability. Appreciate the idea though — the heavier kernel is genuinely more stressful, just not in a way that gives a stable pass/fail signal here.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
```diff
@@ -0,0 +1,165 @@
...
+# Running enough programs on one device exercises the multi-process GPU
+# debug path -- including the firmware context switching that kicks in
+# once the number of simultaneously debugged processes crosses the
+# hardware limit -- as well as GDB's multi-inferior plumbing.
+#
+# Inferiors are created with "add-inferior" rather than fork, so the
+# test does not depend on fork being available.
+#
+# The test is gated on hip_devices_support_debug_multi_process, which
+# excludes targets that do not support multi-process GPU debugging.
```
lumachad: Don't think we need to mention this. It is obvious from the require below.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
```diff
@@ -0,0 +1,165 @@
...
+		gdb_test "add-inferior" "Added inferior $i.*" \
+		    "add inferior $i"
+		gdb_test "inferior $i" "Switching to inferior $i.*" \
+		    "switch to inferior $i"
+		gdb_load $::binfile
+	    }
+	    gdb_test_no_output "set args $device" "set args, inferior $i"
+	    gdb_test "start" "Temporary breakpoint $::decimal, main .*" \
+		"start inferior $i"
+	}
```
lumachad, line 70: This whole block is indented wrong. Please fix it.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
```diff
@@ -0,0 +1,165 @@
...
+proc do_test { num_inferiors device } {
+    clean_restart
+    gdb_load $::binfile
+
+    gdb_test_no_output "set non-stop on"
+
+    # Bringing up many GPU runtimes concurrently on one device
+    # serializes and can take a while, so scale the timeout with the
+    # number of inferiors.
+    with_timeout_factor [expr { ($num_inferiors + 7) / 8 }] {
```
lumachad: Why are we limiting it to 8 here? If there is a reason, explain it in a comment. Otherwise we should not limit this.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
```diff
@@ -0,0 +1,165 @@
...
+	    }
+	    timeout {
+		fail $test
+		verbose -log "only [llength $seen] of $num_inferiors\
+			inferiors stopped before timeout"
+		return
+	    }
+	}
+
+	# Resume everything again; every inferior should run to a clean
```
lumachad: Let's format these comments in the standard way for gdb. So drop these ; or -- or -.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
```diff
@@ -0,0 +1,165 @@
...
+	    }
+	    timeout {
+		fail $test
+		verbose -log "only [llength $exited] of $num_inferiors\
+			inferiors exited before timeout"
+		return
+	    }
+	}
+    }
+    }
```
lumachad: Indentation of these brackets is wrong. Please fix.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
```diff
@@ -0,0 +1,165 @@
...
+# work too but mainly cost wall-clock, as bringing up that many GPU
+# runtimes on one device serializes.
+#
+# The test deliberately runs everything on the default GPU (device 0)
+# rather than spreading across devices: saturating a single GPU is what
+# exercises the per-device limit and the firmware context switching,
+# and locking to one GPU keeps the test cheap and shard-friendly.
+foreach_with_prefix num_inferiors { 8 32 } {
+    do_test $num_inferiors 0
+}
```
lumachad, line 161: Same thing. Let's not have these ;, -- or - in the comments. Let's keep it standard for gdb.

Another thing is we shouldn't pin to GPU 0 as we don't know what kind of GPU that is. It might be unsupported. Besides that, we should honor ROCR_VISIBLE_DEVICES.

We should make it so we guarantee we will run in a single GPU, but that GPU is to be picked by either the system, runtime or a fallback in case none of these are set.

Also, the comment mentions N=16 but then proceeds to test N/2 and N x 2. We should probably set 16 and then have the test automatically run tests with N/2 and N*2. That makes it clearer and makes maintenance easier.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
```diff
@@ -0,0 +1,165 @@
...
+	    }
+	}
+    }
+    }
+}
+
+# The hardware allows a limited number of processes (around 8 to 16) to
+# be debugged simultaneously on a GPU; beyond that the firmware does
+# transparent context switching.  Take N = 16 as a count that reaches
+# that limit and, following the ticket, exercise N/2 and N*2 processes:
```
lumachad: We shouldn't reference tickets. They are not accessible by upstream.

gdb/testsuite/gdb.rocm/multi-inferior-stress.cpp
```diff
@@ -0,0 +1,47 @@
...
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.  */
+
+/* Simple GPU kernel launcher for the multi-inferior stress test.
+
+   The companion .exp launches this program as many separate inferiors,
```
lumachad: This should be an explanation for the .exp file, not the .cpp file.

gdb/testsuite/gdb.rocm/multi-inferior-stress.cpp
lumachad: Should we use __builtin_amdgcn_s_sleep(1) instead?

lumachad (changes requested):
One important point. I think this should go upstream first given it is a generic test, as long as we don't have lack of upstream features to support these tests.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
lumachad: > Another thing is we shouldn't pin to GPU 0 as we don't know what kind of GPU that is. It might be unsupported. Besides that, we should honor ROCR_VISIBLE_DEVICES.
> 
> We should make it so we guarantee we will run in a single GPU, but that GPU is to be picked by either the system, runtime or a fallback in case none of these are set.

I went to check this one. Turns out setting GPU 0 already takes into account the filtering by ROCR_VISIBLE_DEVICES. So if we make only GPU 2 visible, the choice of GPU 0 in the testcase will pick the first in the set, which is only GPU 2.

So I think this isn't a concern for us. Feel free to ignore it.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
spatrang: Removed the gating comment from the header, it's clear from the require lines.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
spatrang: Fixed.  The mis-indentation came from wrapping the body in an extra with_timeout_factor block; I moved with_timeout_factor out to wrap the do_test call instead, so the body sits at the correct single level.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
spatrang: The 8 is the timeout divisor (roughly one extra timeout unit per 8 inferiors), not a process limit.  Added a comment explaining it.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
spatrang: Done, reworded into plain sentences without the semicolons/dashes.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
spatrang: Fixed along with the block above (removed the extra with_timeout_factor nesting).

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
spatrang: Done, dropped the ; -- and - from the comments.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
spatrang: Removed the ticket reference.

gdb/testsuite/gdb.rocm/multi-inferior-stress.cpp
spatrang: Moved the test description into the .exp; the .cpp header now only describes the program.

gdb/testsuite/gdb.rocm/multi-inferior-stress.cpp
spatrang: Done, switched to __builtin_amdgcn_s_sleep(1).

spatrang:
Thanks for the review @lumachad. Pushed an updated version addressing all the comments (rebased on latest amd-staging):

**Cosmetic / comments**
- Fixed the block/bracket indentation (the extra `with_timeout_factor` nesting is gone; it now wraps the `do_test` call instead).
- Dropped the `;`, `--` and `-` connectors from comments; reworded into plain sentences.
- Removed the ticket reference and the redundant gating comment in the header.
- Moved the test description from the `.cpp` into the `.exp`; the `.cpp` header now only describes the program.

**Design**
- No longer pins to a hardcoded device. The test sets `ROCR_VISIBLE_DEVICES` to a single GPU, honoring an existing `ROCR_VISIBLE_DEVICES` (first entry) and otherwise falling back to the first device, applied per-inferior since `add-inferior` gives each inferior a fresh environment.
- `N = 16`, and the test now derives and runs `N/2` and `N*2` automatically.
- Explained the timeout divisor in a comment.
- `kern()` uses `__builtin_amdgcn_s_sleep(1)` instead of inline asm.

Verified onROCm 7.14: 166 pass, 0 fail.

On sending this upstream first: agreed it's a generic test with no downstream-only dependency. I'm happy to submit it to upstream GDB first and then let it flow back here. Let me know if you'd like me to proceed that way.

gdb/testsuite/gdb.rocm/multi-inferior-stress.exp
spatrang: Thanks for checking — you're right. Since device 0 already resolves to the first ROCR_VISIBLE_DEVICES-visible GPU, I dropped the explicit ROCR_VISIBLE_DEVICES handling I'd added and simplified: every inferior just uses the default device (device 0), so they all share the same single GPU and an existing ROCR_VISIBLE_DEVICES is honored automatically. A comment in do_test notes this. (N=16 with auto-derived N/2 and N*2 is unchanged.)

gdb/testsuite/gdb.rocm/multi-inferior-stress.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.  */
+
+/* GPU kernel launcher used by gdb.rocm/multi-inferior-stress.exp.
```
lumachad: Drop this. If we reuse this file for other tests, this is bound to get copied over.

lumachad (approved):
OK with the comment fixed.

gdb/testsuite/gdb.rocm/multi-inferior-stress.cpp
spatrang: Done — dropped the file-specific reference; the header is now generic so it stays correct if the program is reused by another test.

PR #132, opened 2026-05-20
[2026-05-20] Merge upstream/master into amd-staging

lancesix:
Handle conflicts introduced by:

- https://github.com/ROCm/ROCgdb/commit/e7fce05fcf47c487b4f996f08a68f5818d1d41ca gdb/python: fix use of frame_info_ptr within pending_frame_object

The change introduced a use SCOPE_EXIT, but compilation failed because
this macro was not available.  Fix it by explicitly including the
required header.

- https://github.com/ROCm/ROCgdb/commit/958d06262a7a1da5c5fa99cc7b1fd3aa382c5c46 gdb: fix coerce_pieced_ref for multi-piece values

This revealed that the implementation of the downstream
coerce_closure_ref.

This patch adds a new testcase, where we have a struct defined as
follows:

    struct S {
      int x;
      int &ref;
    };

The test defines a variable "s" of that time, whose implicit value is
given by the following DWARF expression:

    (gdb) info address s
    Symbol "s" is a complex DWARF expression:
         0: DW_OP_const4u 123
         5: DW_OP_stack_value
         6: DW_OP_piece 4 (bytes)
         8: DW_OP_GNU_implicit_pointer DIE 81 offset 0
        14: DW_OP_piece 8 (bytes)

That is, the value s's first (x) element is 123, and second field (ref)
references the an implicit value (value of DIE at offset 81).  However,
the original implementation for coerce_closure_ref was calling:

    return closure->get_location ()->indirect_implicit_ptr (frame, type);

In that case, the closure's location was the location of "s", so
indirect_implicit_ptr was looking for an implicit at the start of s, not
at the expected offset 4.  Fixed by adding the required parameter to the
call to indirect_implicit_ptr.

- https://github.com/ROCm/ROCgdb/commit/8f65ab7b71f12def1934ed0c296eef441198b4fa gdb: use value::embedded_offset in check_pieced_synthetic_pointer

Nothing major for this conflict, just make sure to use embedded_offset
in the downstream implementation to follow upstream changes.


simark (approved):
In the commit message, there is an incomplete sentence:

>     This revealed that the implementation of the downstream
    coerce_closure_ref.

Also:

>  The test defines a variable "s" of that time

time -> type

Also:

> references the an implicit value

drop "the"

The code in expr.c was changed to use `TARGET_CHAR_BIT`, we could perhaps switch to that instead of `HOST_CHAR_BIT`, to be a bit closer to upstream.  It's going to be 8 anyway.

I don't fully understand the changes, but the changes came with tests, so if they pass it should be fine.

lancesix:
> In the commit message, there is an incomplete sentence:
> 
> > ```
> > This revealed that the implementation of the downstream
> > ```
> 
> ```
> coerce_closure_ref.
> ```
> 
> Also:
> 
> > The test defines a variable "s" of that time
> 
> time -> type
> 
> Also:
> 
> > references the an implicit value
> 
> drop "the"
> 
> The code in expr.c was changed to use `TARGET_CHAR_BIT`, we could perhaps switch to that instead of `HOST_CHAR_BIT`, to be a bit closer to upstream. It's going to be 8 anyway.
> 
> I don't fully understand the changes, but the changes came with tests, so if they pass it should be fine.

Oh, if I use `HOST_CHAR_BIT` that was not intentional.  I really meant `TARGET_CHAR_BIT`.

I'll address the other remarks, thanks for catching this.

PR #134, opened 2026-05-20
gdb/testsuite/gdb.rocm: add OpenMP GPU offload tests

spatrang:
## Summary
Add OpenMP GPU offload coverage to the `gdb.rocm/` testsuite.
The existing `gdb.rocm/` tests cover HIP-style kernels (`__global__`
functions launched via `hipLaunchKernelGGL`), but there is no coverage
for OpenMP `target` offload — an increasingly common way of programming
AMD GPUs from C, C++ and Fortran. Without these tests, regressions in
GDB's handling of:
- the `__omp_offloading_<hash>_<func>_l<line>` outliner-frame symbols,
- OpenMP-runtime-loaded device ELFs (loaded via `libomptarget` plugins,
  not `libamdhip64`),
- DWARF emitted for OpenMP `map`-clause variables,
can land silently while the HIP path keeps working.
## Motivation
`gdb.threads/` already has CPU-side OpenMP tests (`omp-par-scope.exp`,
`omp-task.exp`), but nothing in the testsuite exercises the `target`
construct on an actual GPU. This PR adds those tests and the dejagnu
framework hooks needed to drive an OpenMP-offload-capable Clang/Flang
toolchain.
### Framework hooks in `gdb/testsuite/lib/rocm.exp` (+186 lines)
Reusable procs that future OpenMP-offload tests can also rely on:
- \`rocm_find_llvm_tool\` — locate a tool under \`\$ROCM_PATH/llvm/bin\`,
  then \`\$ROCM_PATH/lib/llvm/bin\`, then \`PATH\`.
- \`find_amdclang\`, \`find_amdclangpp\`, \`find_amdflang\` — cached
  wrappers for the three offload-capable compiler drivers.
- \`allow_omp_offload_tests\`, \`allow_omp_offload_fortran_tests\` —
  \`require\`-compatible predicates: amd-dbgapi support, toolchain
  availability, ≥ 1 AMD GPU.
- \`rocm_omp_offload_flags <lang>\` — build the
  \`-fopenmp -fopenmp-targets=amdgcn-amd-amdhsa --offload-arch=…\`
  flag set; omits the Clang-only \`-Wno-unused-command-line-argument\`
  for Fortran (the Flang driver rejects that diagnostic option).
- \`gdb_compile_omp_offload\`, \`gdb_compile_omp_offload_cpp\`,
  \`gdb_compile_omp_offload_fortran\` — \`gdb_compile\` wrappers that
  swap in the right offload-capable driver and add the offload flags.
- \`with_rocm_omp_gpu_lock\` — alias for \`with_rocm_gpu_lock\` so
  OpenMP tests serialize with HIP tests on the GPU.
### New tests under \`gdb/testsuite/gdb.rocm/\` (16 files)
- \`omp-target-break\` — basic \`#pragma omp target\` breakpoint by
  \`file:line\`, AMDGPU wave check.
- \`omp-target-step\` — \`next\` stepping inside the offloaded region
  keeps the program on a wave.
- \`omp-target-locals\` — inspect \`map(to:)\`, \`map(tofrom:)\`,
  \`firstprivate\` and a device-local automatic variable.
- \`omp-target-teams\` — \`target teams distribute parallel for\`
  produces multiple AMDGPU waves at one breakpoint; \`thread N\`
  switches to a wave; \`print i\` works inside it.
- \`omp-target-data\` — \`target data\` enclosing two consecutive
  \`target\` regions; the device buffer is reused between kernels
  (verified via \`c[0] == 100\` in the second kernel).
- \`omp-target-multi-kernel\` — two \`declare target\` device functions
  launched from separate target regions; named breakpoints land on the
  right kernel.
- \`omp-target-cpp\` — templated device function (\`add_t<int>\`) and
  struct functor (\`multiplier::operator()\`) compiled with the C++
  driver; \`print this->factor\` works.
- \`omp-target-fortran\` — \`!\$omp target\` compiled with the Fortran
  driver; breakpoint inside the loop hits an AMDGPU wave.
Each \`.exp\` follows the established \`gdb.rocm/\` patterns:
\`load_lib rocm.exp\`, \`require allow_omp_offload_tests\`,
\`with_rocm_omp_gpu_lock { … }\`, and uses the \`gdb_test_multiple\` +
\`exp_continue\` idiom (same as \`gdb.rocm/shared-memory.exp\`) so the
patterns survive multi-line GDB output.
## Testing

\`\`\`
make -C testsuite check TESTS="gdb.rocm/omp-target-*.exp"
…
# of expected passes            47
# of unexpected failures        0
\`\`\`
Per-test PASS / FAIL counts:
- \`omp-target-break.exp\` — 3 PASS, 0 FAIL
- \`omp-target-step.exp\` — 9 PASS, 0 FAIL
- \`omp-target-locals.exp\` — 7 PASS, 0 FAIL
- \`omp-target-teams.exp\` — 6 PASS, 0 FAIL
- \`omp-target-data.exp\` — 8 PASS, 0 FAIL
- \`omp-target-multi-kernel.exp\` — 5 PASS, 0 FAIL
- \`omp-target-cpp.exp\` — 6 PASS, 0 FAIL
- \`omp-target-fortran.exp\` — 3 PASS, 0 FAIL
- **Total — 47 PASS, 0 FAIL**
On hosts without an offload-capable Clang/Flang toolchain or without an
AMD GPU, the \`require allow_omp_offload_tests\` /
\`allow_omp_offload_fortran_tests\` gates report the tests as
\`UNSUPPORTED\` rather than \`FAIL\`, so this PR is safe to land in CI
immediately.
## Backwards compatibility
- \`lib/rocm.exp\` is only extended; no existing proc is renamed and
  no signature changes. Existing HIP tests are unaffected.
- New procs are uniquely named (e.g. \`find_amdclang\` vs the existing
  \`find_hipcc\`) so there is no collision.

### DIFF
--- a/gdb/testsuite/gdb.rocm/omp-target-basic.c
+++ b/gdb/testsuite/gdb.rocm/omp-target-basic.c
@@ -1,0 +1,57 @@
+/* Copyright 2026 Free Software Foundation, Inc.
+   Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+   This file is part of GDB.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.  */
+
+/* Basic OpenMP "target" offload program shared by the breakpoint,
+   data-sharing (locals) and single-stepping checks in
+   omp-target-basic.exp.  It offloads a single '#pragma omp target'
+   region with a mapped input array (to:), a mapped output array
+   (tofrom:) and a firstprivate scalar, plus a few distinguishable
+   statements at the top of the region for single-stepping.  */
+
+#include <stdio.h>
+
+#define N 8
+
+int
+main (void)
+{
+  int in_arr[N];
+  int out_arr[N];
+
+  for (int i = 0; i < N; i++)
+    {
+      in_arr[i] = 100 + i;
+      out_arr[i] = 0;
+    }
+
+  int firstpriv_val = 42;
+
+  #pragma omp target						\
+	      map(to:in_arr[0:N]) map(tofrom:out_arr[0:N])	\
+	      firstprivate(firstpriv_val)
+  {
+    int priv_val = 7;			/* first-target-line */
+    int step_two = priv_val + 1;
+    int step_three = step_two + 1;
+    out_arr[0] = (in_arr[0] + priv_val + firstpriv_val
+		  + step_two + step_three);
+  }
+
+  printf ("out[0]=%d\n", out_arr[0]);
+  return 0;
+}
--- a/gdb/testsuite/gdb.rocm/omp-target-basic.exp
+++ b/gdb/testsuite/gdb.rocm/omp-target-basic.exp
@@ -1,0 +1,82 @@
+# Copyright 2026 Free Software Foundation, Inc.
+# Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+#
+# This file is part of GDB.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Basic OpenMP-offload device-debugging checks, all sharing a single
+# '#pragma omp target' binary:
+#
+#   - "breakpoint": set a breakpoint inside the target region by
+#     file:line, hit it on the device, and observe an AMDGPU wave.
+#   - "locals": inspect variables in different data-sharing classes
+#     (map(to:), map(tofrom:), firstprivate).
+#   - "step": single-step through statements in the region, staying on
+#     the device.
+
+load_lib rocm.exp
+
+require allow_omp_offload_tests
+
+standard_testfile
+
+set first_line [gdb_get_line_number "first-target-line"]
+
+if {[prepare_for_testing "failed to prepare" $testfile $srcfile \
+	 {debug omp-rocm}] == -1} {
+    return -1
+}
+
+with_rocm_gpu_lock {
+    if {![runto_main]} {
+	return
+    }
+
+    gdb_breakpoint "$srcfile:$first_line" -allow-pending
+
+    with_test_prefix "breakpoint" {
+	gdb_continue_to_breakpoint "device breakpoint"
+	gdb_assert {[llength [info_thread_get_wave_list]] >= 1} \
+	    "AMDGPU wave at OMP-target stop"
+    }
+
+    # Still stopped at the first statement in the region (out_arr is
+    # not written yet), so the data-sharing values are deterministic.
+    with_test_prefix "locals" {
+	gdb_test "print firstpriv_val" "= 42" "firstprivate value"
+	gdb_test "print in_arr\[0\]" "= 100" "mapped-to in_arr\[0\]"
+	gdb_test "print in_arr\[7\]" "= 107" "mapped-to in_arr\[7\]"
+	gdb_test "print out_arr\[0\]" "= 0" \
+	    "mapped-tofrom out_arr\[0\] before fill"
+    }
+
+    # Step forward through a few statements; each step must stay on a
+    # GPU wave.  Line numbers are not pinned because OpenMP-offload
+    # line tables vary across compiler versions.
+    foreach which {1 2 3} {
+	with_test_prefix "next-$which" {
+	    # Require landing on a numbered source line (so a step that
+	    # errors out is not silently accepted); the exact line is
+	    # not pinned because OpenMP-offload line tables vary.
+	    gdb_test "next" "$::decimal\t\[^\r\n\]*" "next inside target region"
+	    gdb_assert {[llength [info_thread_get_wave_list]] >= 1} \
+		"still on AMDGPU wave"
+	}
+    }
+
+    # The OpenMP runtime's host threads emit "[Thread ... exited]"
+    # lines before the program exits, so allow extra output.
+    gdb_continue_to_end "" "continue" 1
+}
--- a/gdb/testsuite/gdb.rocm/omp-target-cpp.exp
+++ b/gdb/testsuite/gdb.rocm/omp-target-cpp.exp
@@ -1,0 +1,72 @@
+# Copyright 2026 Free Software Foundation, Inc.
+# Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+#
+# This file is part of GDB.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# C++-specific OpenMP-offload tests.  Verify that GDB can break
+# inside a templated device function and a struct member function
+# used in a target region.
+
+load_lib rocm.exp
+
+require allow_omp_offload_tests
+
+standard_testfile .cpp
+
+set tmpl_line    [gdb_get_line_number "tmpl-line"]
+set functor_line [gdb_get_line_number "functor-line"]
+
+if {[prepare_for_testing "failed to prepare" $testfile $srcfile \
+	 {debug c++ omp-rocm}] == -1} {
+    return -1
+}
+
+with_rocm_gpu_lock {
+    if {![runto_main]} {
+	return
+    }
+
+    with_test_prefix "templated add_t" {
+	gdb_breakpoint "$srcfile:$tmpl_line" -allow-pending
+	gdb_continue_to_breakpoint "add_t<int>"
+	# The kernel reaches add_t through sum_pair, so the device call
+	# stack shows both frames.
+	gdb_test "backtrace 2" \
+	    "#0 .*add_t<int> .*\r\n#1 .*sum_pair<int> .*" \
+	    "add_t called from sum_pair"
+	delete_breakpoints
+    }
+
+    with_test_prefix "multiplier::operator()" {
+	gdb_breakpoint "$srcfile:$functor_line" -allow-pending
+	gdb_continue_to_breakpoint "multiplier::operator()"
+	# Likewise, the functor is reached through scale.
+	gdb_test "backtrace 2" \
+	    "#0 .*multiplier::operator\\(\\) .*\r\n#1 .*scale .*" \
+	    "functor called from scale"
+
+	# 'this->factor' should be the value we set on the host (3),
+	# transferred via the implicit 'map(to:m)'.
+	gdb_test "print this->factor" \
+	    "= 3" \
+	    "factor visible via 'this'"
+	delete_breakpoints
+    }
+
+    # The OpenMP runtime's host threads emit "[Thread ... exited]"
+    # lines before the program exits, so allow extra output.
+    gdb_continue_to_end "" "continue" 1
+}
--- a/gdb/testsuite/gdb.rocm/omp-target-data.exp
+++ b/gdb/testsuite/gdb.rocm/omp-target-data.exp
@@ -1,0 +1,64 @@
+# Copyright 2026 Free Software Foundation, Inc.
+# Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+#
+# This file is part of GDB.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# 'target data' region with two nested 'target' regions that reuse
+# the same device buffer.  Verify GDB can break inside both kernels
+# in sequence.
+
+load_lib rocm.exp
+
+require allow_omp_offload_tests
+
+standard_testfile
+
+set k1_line [gdb_get_line_number "data-k1"]
+set k2_line [gdb_get_line_number "data-k2"]
+
+if {[prepare_for_testing "failed to prepare" $testfile $srcfile \
+	 {debug omp-rocm}] == -1} {
+    return -1
+}
+
+with_rocm_gpu_lock {
+    if {![runto_main]} {
+	return
+    }
+
+    gdb_breakpoint "$srcfile:$k1_line" -temporary -allow-pending
+
+    with_test_prefix "first kernel" {
+	gdb_continue_to_breakpoint "first kernel"
+	gdb_test "print a\[0\]" "= 0"   "a\[0\] in first kernel"
+	gdb_test "print b\[0\]" "= 100" "b\[0\] in first kernel"
+    }
+
+    gdb_breakpoint "$srcfile:$k2_line" -temporary -allow-pending
+
+    with_test_prefix "second kernel" {
+	gdb_continue_to_breakpoint "second kernel"
+	# After the first kernel ran to completion (its temporary
+	# breakpoint already auto-deleted), c[0] = a[0] + b[0]
+	# = 0 + 100 = 100.  The 'target data' region preserves this on
+	# the device, so we observe it.
+	gdb_test "print c\[0\]" "= 100" "c\[0\] in second kernel"
+    }
+
+    # The OpenMP runtime's host threads emit "[Thread ... exited]"
+    # lines before the program exits, so allow extra output.
+    gdb_continue_to_end "" "continue" 1
+}
--- a/gdb/testsuite/gdb.rocm/omp-target-fortran.exp
+++ b/gdb/testsuite/gdb.rocm/omp-target-fortran.exp
@@ -1,0 +1,70 @@
+# Copyright 2026 Free Software Foundation, Inc.
+# Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+#
+# This file is part of GDB.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Fortran OpenMP-offload test: stop in the offloaded loop and read the
+# loop index, then stop inside a '!$omp declare target' function and
+# check the device-side call stack.
+
+load_lib rocm.exp
+
+require allow_omp_offload_fortran_tests
+
+standard_testfile .f90
+
+set add_line [gdb_get_line_number "fadd-line"]
+
+# Build with -O0 so device locals (e.g. the loop index) stay visible:
+# amdflang optimizes the offloaded code by default, dropping their
+# debug info.  This is amdflang-specific (amdclang already defaults to
+# -O0) and is tracked upstream as ROCM-19898, so keep the workaround
+# local to this test rather than in the shared compile flags.
+if {[prepare_for_testing "failed to prepare" $testfile $srcfile \
+	 {debug f90 omp-rocm additional_flags=-O0}] == -1} {
+    return -1
+}
+
+with_rocm_gpu_lock {
+    if {![runto_main]} {
+	return
+    }
+
+    # Stop inside the '!$omp declare target' function and confirm the
+    # innermost frame is the device function.
+    gdb_breakpoint "$srcfile:$add_line" -temporary -allow-pending
+    gdb_continue_to_breakpoint "add_vals"
+    gdb_test "backtrace 1" "#0 .*add_vals .*" "stopped in add_vals"
+
+    # a(i) == i and b(i) == 2 * i, so the arguments satisfy y == 2 * x.
+    # Use distinct failure defaults so the relationship can't trivially
+    # hold if a read fails.
+    set x [get_integer_valueof "x" -1 "device function arg x"]
+    set y [get_integer_valueof "y" -2 "device function arg y"]
+    gdb_assert {$x >= 1 && $y == 2 * $x} "b(i) == 2 * a(i)"
+
+    # Read the device-local loop index in the kernel (caller) frame.
+    # 'i' is only visible because amdflang is built with -O0 (see
+    # ROCM-19898 for the default-optimization debug-info limitation).
+    gdb_test "frame 1" "__omp_offloading\[^\r\n\]*\r\n$::decimal\[^\r\n\]*" \
+	"select kernel frame"
+    set i [get_integer_valueof "i" 0 "loop index"]
+    gdb_assert {$i >= 1 && $i <= 32} "loop index is in range"
+
+    # The OpenMP runtime's host threads emit "[Thread ... exited]"
+    # lines before the program exits, so allow extra output.
+    gdb_continue_to_end "" "continue" 1
+}
--- a/gdb/testsuite/gdb.rocm/omp-target-fortran.f90
+++ b/gdb/testsuite/gdb.rocm/omp-target-fortran.f90
@@ -1,0 +1,53 @@
+! Copyright 2026 Free Software Foundation, Inc.
+! Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+!
+! This file is part of GDB.
+!
+! This program is free software; you can redistribute it and/or modify
+! it under the terms of the GNU General Public License as published by
+! the Free Software Foundation; either version 3 of the License, or
+! (at your option) any later version.
+!
+! This program is distributed in the hope that it will be useful,
+! but WITHOUT ANY WARRANTY; without even the implied warranty of
+! MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+! GNU General Public License for more details.
+!
+! You should have received a copy of the GNU General Public License
+! along with this program.  If not, see <http://www.gnu.org/licenses/>.
+!
+! Small Fortran program offloading a loop to the GPU via '!$omp
+! target'.  The loop calls a '!$omp declare target' function so the
+! test can also check the device-side call stack.
+
+module devmod
+  implicit none
+contains
+  function add_vals (x, y) result (r)
+    !$omp declare target
+    integer, intent(in) :: x, y
+    integer :: r
+    r = x + y                       ! fadd-line
+  end function add_vals
+end module devmod
+
+program omp_target_fortran
+  use devmod
+  implicit none
+  integer, parameter :: n = 32
+  integer :: a(n), b(n), c(n), i
+
+  do i = 1, n
+     a(i) = i
+     b(i) = 2 * i
+     c(i) = 0
+  end do
+
+  !$omp target map(to:a, b) map(tofrom:c)
+  do i = 1, n
+     c(i) = add_vals (a(i), b(i))
+  end do
+  !$omp end target
+
+  print *, 'c1=', c(1)
+end program omp_target_fortran
--- a/gdb/testsuite/gdb.rocm/omp-target-multi-kernel.exp
+++ b/gdb/testsuite/gdb.rocm/omp-target-multi-kernel.exp
@@ -1,0 +1,59 @@
+# Copyright 2026 Free Software Foundation, Inc.
+# Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+#
+# This file is part of GDB.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Two '#pragma omp declare target' device functions (square_dev and
+# cube_dev) launched from two separate target regions in a single
+# binary.  Verify breakpoints set on each named device function are
+# hit by their corresponding kernel only.
+
+load_lib rocm.exp
+
+require allow_omp_offload_tests
+
+standard_testfile
+
+if {[prepare_for_testing "failed to prepare" $testfile $srcfile \
+	 {debug omp-rocm}] == -1} {
+    return -1
+}
+
+with_rocm_gpu_lock {
+    if {![runto_main]} {
+	return
+    }
+
+    # This is a serial 'omp target' loop, so the first iteration
+    # (i == 0) handles in[0] == 1; the device argument x is therefore 1
+    # in both kernels.  Reading it confirms we stopped in the right
+    # device function with a live argument (which also implies a wave).
+    with_test_prefix "square kernel" {
+	gdb_breakpoint "square_dev" -temporary -allow-pending
+	gdb_continue_to_breakpoint "square_dev"
+	gdb_test "print x" "= 1"
+    }
+
+    with_test_prefix "cube kernel" {
+	gdb_breakpoint "cube_dev" -temporary -allow-pending
+	gdb_continue_to_breakpoint "cube_dev"
+	gdb_test "print x" "= 1"
+    }
+
+    # The OpenMP runtime's host threads emit "[Thread ... exited]"
+    # lines before the program exits, so allow extra output.
+    gdb_continue_to_end "" "continue" 1
+}
--- a/gdb/testsuite/gdb.rocm/omp-target-teams.c
+++ b/gdb/testsuite/gdb.rocm/omp-target-teams.c
@@ -1,0 +1,54 @@
+/* Copyright 2026 Free Software Foundation, Inc.
+   Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+   This file is part of GDB.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.  */
+
+/* Use 'target teams distribute parallel for' so that GDB sees
+   multiple GPU waves at the same breakpoint.  */
+
+#include <stdio.h>
+
+#define N 256
+
+int
+main (void)
+{
+  int a[N];
+  int b[N];
+  int c[N];
+
+  for (int i = 0; i < N; i++)
+    {
+      a[i] = i;
+      b[i] = N - i;
+      c[i] = 0;
+    }
+
+  #pragma omp target teams distribute parallel for	\
+	      map(to:a[0:N], b[0:N]) map(tofrom:c[0:N])
+  for (int i = 0; i < N; i++)
+    c[i] = a[i] + b[i];		/* teams-line */
+
+  int correct = 1;
+  for (int i = 0; i < N; i++)
+    {
+      if (c[i] != N)
+	correct = 0;
+    }
+
+  printf ("correct=%d\n", correct);
+  return correct ? 0 : 1;
+}
--- a/gdb/testsuite/gdb.rocm/omp-target-teams.exp
+++ b/gdb/testsuite/gdb.rocm/omp-target-teams.exp
@@ -1,0 +1,65 @@
+# Copyright 2026 Free Software Foundation, Inc.
+# Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+#
+# This file is part of GDB.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# 'target teams distribute parallel for' spreads the loop across GPU
+# lanes, with consecutive iterations on consecutive lanes.  Verify this
+# by reading the loop index on two different lanes (the difference must
+# match the lane distance) and checking a per-lane computed value.
+
+load_lib rocm.exp
+
+require allow_omp_offload_tests
+
+standard_testfile
+
+set break_line [gdb_get_line_number "teams-line"]
+
+if {[prepare_for_testing "failed to prepare" $testfile $srcfile \
+	 {debug omp-rocm}] == -1} {
+    return -1
+}
+
+with_rocm_gpu_lock {
+    if {![runto_main]} {
+	return
+    }
+
+    gdb_breakpoint "$srcfile:$break_line" -allow-pending
+
+    gdb_continue_to_breakpoint "teams breakpoint"
+
+    # 'distribute parallel for' runs consecutive iterations on
+    # consecutive lanes, so lane L sees loop index i = i0 + L.  Switch
+    # to lane 0 explicitly (rather than relying on which lane the stop
+    # selected), read i, then read it on lane 7, and check the two
+    # lanes are 7 iterations apart.
+    gdb_test "lane 0" "Switching to .*lane 0 .*" "switch to lane 0"
+    set i0 [get_integer_valueof "i" -1 "loop index on lane 0"]
+    gdb_test "lane 7" "Switching to .*lane 7 .*" "switch to lane 7"
+    set i7 [get_integer_valueof "i" -1 "loop index on lane 7"]
+    gdb_assert {$i7 - $i0 == 7} "lane 7 runs 7 iterations past lane 0"
+
+    # a[i] == i and b[i] == N - i, so a[i] + b[i] == N (256) on any
+    # lane (evaluated here on lane 7).
+    gdb_test "print a\[i\] + b\[i\]" "= 256" "a\[i\] + b\[i\] == N"
+
+    delete_breakpoints
+    # The OpenMP runtime's host threads emit "[Thread ... exited]"
+    # lines before the program exits, so allow extra output.
+    gdb_continue_to_end "" "continue" 1
+}
--- a/gdb/testsuite/lib/rocm.exp
+++ b/gdb/testsuite/lib/rocm.exp
@@ -693,3 +693,52 @@ gdb_caching_proc compiler_dwarf_language {} {
 	return "others"
     }
 }
+
+# OpenMP GPU offload support.
+#
+# The procs below extend the ROCm test framework to support running
+# tests that use the OpenMP "target" construct to offload code to
+# the GPU.
+
+# Return 1 if the C/C++ OpenMP-offload toolchain and a usable AMD GPU
+# are both available; otherwise return a {0 reason} pair suitable for
+# 'require'.
+#
+# The actual OpenMP-offload compile is driven by gdb_compile's
+# "omp-rocm" option, which selects the AMD compiler (amdclang /
+# amdclang++ / amdflang) and adds the offload flags.
+
+gdb_caching_proc allow_omp_offload_tests {} {
+    # Reuse the shared ROCm gate (native target, amd-dbgapi support and
+    # a usable AMD GPU device); the device check compiles a HIP program
+    # with amdclang++, proving the AMD LLVM toolchain that ships
+    # amdclang / amdclang++ is present.  OpenMP offload only adds a
+    # Linux-only restriction on top.
+    set base [allow_hip_tests]
+    if {[lindex $base 0] == 0} {
+	return $base
+    }
+
+    if {![istarget "*-linux*"]} {
+	return {0 "OpenMP offload is only tested on Linux"}
+    }
+
+    return 1
+}
+
+# Same as above, but also requires amdflang for Fortran tests.
+
+gdb_caching_proc allow_omp_offload_fortran_tests {} {
+    set base [allow_omp_offload_tests]
+    if {[lindex $base 0] == 0} {
+	return $base
+    }
+
+    # Fortran offload needs amdflang specifically.  Resolve it through
+    # the same discovery default_target_compile uses, and confirm it
+    # actually runs.
+    if {[catch {exec [gdb_find_amd_omp_compiler amdflang] --version}]} {
+	return {0 "amdflang not available"}
+    }
+    return 1
+}
--- a/gdb/testsuite/gdb.rocm/omp-target-cpp.cpp
+++ b/gdb/testsuite/gdb.rocm/omp-target-cpp.cpp
@@ -1,0 +1,95 @@
+/* Copyright 2026 Free Software Foundation, Inc.
+   Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+   This file is part of GDB.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.  */
+
+/* C++-specific OpenMP-offload features.  Exercises a templated device
+   function and a callable functor used inside a '#pragma omp target'
+   region.  */
+
+#include <cstdio>
+
+#define N 16
+
+#pragma omp declare target
+template <typename T>
+static T
+add_t (T a, T b)
+{
+  return a + b;		/* tmpl-line */
+}
+
+/* A named caller of add_t, so the backtrace at the add_t breakpoint
+   shows a stable intermediate frame (sum_pair -> add_t) rather than
+   only the compiler-generated kernel frame.  */
+template <typename T>
+static T __attribute__ ((noinline))
+sum_pair (T a, T b)
+{
+  return add_t<T> (a, b);
+}
+
+struct multiplier
+{
+  int factor;
+
+  int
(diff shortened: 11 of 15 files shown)


motokultivator:
With flang-23 I get:
`gdb compile failed, flang-23: error: unknown argument: '-fno-stack-protector'`
for omp-target-fortran.exp.
It comes from this part of gdb.exp:
```
    if { !$getting_compiler_info
         && [test_compiler_info {gcc-*-*}]
         && !([test_compiler_info {gcc-[0-3]-*}]
              || [test_compiler_info {gcc-4-0-*}])
         && [lsearch -exact $options rust] == -1
         && [lsearch -exact $options hip] == -1} {
        # Put it at the front to not override any user-provided value.
        lappend new_options "early_flags=-fno-stack-protector"
    }
```

```
flang --version
AMD flang version 23.0.0git (https://github.com/ROCm/llvm-project.git 43215c73116c407735c85a180d174f718798c328+PATCHED:2506c552d8428e2cc1778bef048b20f818e06bb3)
Target: x86_64-unknown-linux-gnu
Thread model: posix
```

spatrang (commented):
> With flang-23 I get:
> `gdb compile failed, flang-23: error: unknown argument: '-fno-stack-protector'`
> for omp-target-fortran.exp.
> It comes from this part of gdb.exp:
> ```
>     if { !$getting_compiler_info
>          && [test_compiler_info {gcc-*-*}]
>          && !([test_compiler_info {gcc-[0-3]-*}]
>               || [test_compiler_info {gcc-4-0-*}])
>          && [lsearch -exact $options rust] == -1
>          && [lsearch -exact $options hip] == -1} {
>         # Put it at the front to not override any user-provided value.
>         lappend new_options "early_flags=-fno-stack-protector"
>     }
> ```
> 
> ```
> flang --version
> AMD flang version 23.0.0git (https://github.com/ROCm/llvm-project.git 43215c73116c407735c85a180d174f718798c328+PATCHED:2506c552d8428e2cc1778bef048b20f818e06bb3)
> Target: x86_64-unknown-linux-gnu
> Thread model: posix
> ``` 

Applied fix in [1d0d3cd](https://github.com/ROCm/ROCgdb/pull/134/commits/1d0d3cd6aec0fab4b3bed3d173aef6c6c912b615) this commit

gdb/testsuite/lib/rocm.exp
```diff
@@ -633,3 +633,189 @@ gdb_caching_proc compiler_dwarf_language {} {
...
+}
+
+# Convenience wrapper: call BODY under both the GPU lock (so only one
+# GDB drives the device at a time) and with the rocm-specific cleanups.
+# This is identical to with_rocm_gpu_lock, but is named to make
+# OpenMP-offload tests read more naturally.
+
+proc with_rocm_omp_gpu_lock { body } {
+    with_rocm_gpu_lock {uplevel 1 $body}
+}
```
lumachad: Why is this needed? Can't we use with_rocm_gpu_lock directly like all the other gdb.rocm tests?

gdb/testsuite/lib/rocm.exp
```diff
@@ -633,3 +633,189 @@ gdb_caching_proc compiler_dwarf_language {} {
...
+	"${prog}"]
+
+    foreach cand $candidates {
+	set res [catch {exec $cand --version} output]
+	if {$res == 0} {
+	    return $cand
+	}
+    }
+    return ""
+}
```
lumachad: I'd imagine if our compiler supports OMP we would only need to run the tests with the right flags. Why do we need to find the compiler here?

gdb/testsuite/lib/rocm.exp
```diff
@@ -633,3 +633,189 @@ gdb_caching_proc compiler_dwarf_language {} {
...
+
+# Find the AMD OpenMP-offload C compiler (amdclang).
+gdb_caching_proc find_amdclang {} {
+    return [rocm_find_llvm_tool amdclang]
+}
+
+# Find the AMD OpenMP-offload C++ compiler (amdclang++).
+gdb_caching_proc find_amdclangpp {} {
+    return [rocm_find_llvm_tool amdclang++]
+}
```
lumachad: Same comment. Why do we need to find the compiler?

gdb/testsuite/lib/rocm.exp
```diff
@@ -633,3 +633,189 @@ gdb_caching_proc compiler_dwarf_language {} {
...
+
+# Find the AMD OpenMP-offload Fortran compiler (amdflang).
+gdb_caching_proc find_amdflang {} {
+    return [rocm_find_llvm_tool amdflang]
+}
```
lumachad: Same comment. Why do we need to find the compiler?

lumachad (commented):
I think we're on the right track, but it seems this patch does things that should be done by https://github.com/ROCm/ROCgdb/pull/62 or some other rework patch before this one.

gdb/testsuite/lib/rocm.exp
```diff
@@ -640,3 +640,189 @@ gdb_caching_proc compiler_dwarf_language {} {
 	return "others"
     }
 }
+
+# OpenMP GPU offload support.
+#
+# The procs below extend the rocm test framework to support running
```
aktemur: How should "rocm" be spelled? Isn't it ROCm?

gdb/testsuite/lib/rocm.exp
```diff
@@ -640,3 +640,189 @@ gdb_caching_proc compiler_dwarf_language {} {
...
+}
+
+# Find the AMD OpenMP-offload Fortran compiler (amdflang).
+gdb_caching_proc find_amdflang {} {
+    return [rocm_find_llvm_tool amdflang]
+}
+
+# Return 1 if the C/C++ OpenMP-offload toolchain and a usable AMD GPU
+# are both available; otherwise return a {0 reason} pair suitable for
+# 'require'.  Mirrors allow_hipcc_tests.
```
aktemur: This proc is almost exactly the same as `allow_hipcc_tests`. I think we can reuse most of the code.

gdb/testsuite/lib/rocm.exp
spatrang: Good point — dropped the alias and switched the OMP tests to call `with_rocm_gpu_lock` directly, like the rest of gdb.rocm.

gdb/testsuite/lib/rocm.exp
spatrang: Agreed. OpenMP offload doesn't need a separate compiler — any offload-capable clang/flang works with the right `-fopenmp -fopenmp-targets=…` flags, so this discovery is really the job of the compiler-configuration rework in #62 rather than this patch. Plan: rebase #134 on top of #62 and drive the build through the configured `*_FOR_TARGET` compiler (+ offload flags), dropping `rocm_find_llvm_tool` and the `find_amd*` helpers. Leaving them for now only so the suite is runnable standalone until #62 lands; I'll remove them in the rebase.

gdb/testsuite/lib/rocm.exp
spatrang: Same as the `find_amdclang` reply above — this overlaps with #62 and will be dropped when #134 is rebased on top of it.

gdb/testsuite/lib/rocm.exp
spatrang: Same as above — handled by the #62-based rework rather than a parallel discovery path here.

gdb/testsuite/lib/rocm.exp
```diff
@@ -640,3 +640,288 @@ gdb_caching_proc compiler_dwarf_language {} {
...
+    # Need amd-dbgapi support in GDB.
+    set output [remote_exec host $::GDB "$::INTERNAL_GDBFLAGS --configuration"]
+    if { [string first "--with-amd-dbgapi" $output] == -1 } {
+	return {0 "amd-dbgapi not supported"}
+    }
+
+    if {[find_amdclang] == ""} {
+	return {0 "amdclang not found in ROCM_PATH"}
+    }
+
```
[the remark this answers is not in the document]
spatrang: Correct on both counts. Once the discovery is reworked on top of #62 this gating goes away; if any capability check remains it will cover the C++ driver too, and the UNSUPPORTED reasons will say "ROCM_PATH or PATH" to match what is actually searched.

gdb/testsuite/lib/rocm.exp
```diff
@@ -640,3 +640,288 @@ gdb_caching_proc compiler_dwarf_language {} {
...
+# Same as above, but also requires amdflang for Fortran tests.
+
+gdb_caching_proc allow_omp_offload_fortran_tests {} {
+    set base [allow_omp_offload_tests]
+    if {[lindex $base 0] == 0} {
+	return $base
+    }
+    if {[find_amdflang] == ""} {
+	return {0 "amdflang not found in ROCM_PATH"}
+    }
```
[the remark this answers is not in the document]
spatrang: Same as above — the reason string will be corrected (it also searches PATH) as part of the #62-based rework.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
```diff
@@ -0,0 +1,74 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# 'target teams distribute parallel for' should produce multiple GPU
+# waves all hitting the same breakpoint.
```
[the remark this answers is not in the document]
spatrang: Updated the header comment to explain that the test requires at least one AMDGPU wave (not an exact count), matching the assertion below.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
```diff
@@ -0,0 +1,74 @@
...
+
+    gdb_breakpoint "$srcfile:$break_line" allow-pending
+
+    rocm_omp_continue_to_breakpoint "continue to teams breakpoint"
+
+    # Verify multiple waves are present.  We don't need an exact wave
+    # count - it depends on the chosen num_teams / thread_limit and
+    # GPU architecture - but there must be at least one.
+    set wave_list [info_thread_get_wave_list]
+    gdb_assert {[llength $wave_list] >= 1} "saw at least one AMDGPU wave"
```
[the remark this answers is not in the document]
spatrang: Reworded this comment to match the `>= 1` assertion.

spatrang:
Thanks for the review @lumachad. Following your note about not blocking #62: I'll keep the OpenMP-offload series here in #134 and base it on top of #62's branch, so #62 stays independent and #134 is reviewed as close-to-final code. The OMP-specific parts (offload flags, test helpers, the `allow_*` gate) stay here and build on #62's reworked compiler discovery/search-order rather than being pushed into #62. Replied inline on the discovery comments with specifics.

Already addressed in the meantime:
- Dropped the `with_rocm_omp_gpu_lock` alias; the tests now use `with_rocm_gpu_lock` directly.
- Reworded the `omp-target-teams` "multiple waves" comments to match the "at least one wave" assertion.

gdb/testsuite/gdb.rocm/omp-target-teams.c
```diff
@@ -0,0 +1,52 @@
...
+      c[i] = 0;
+    }
+
+  #pragma omp target teams distribute parallel for	\
+	      map(to:a[0:N], b[0:N]) map(tofrom:c[0:N])
+  for (int i = 0; i < N; i++)
+    c[i] = a[i] + b[i];		/* teams-line */
+
+  int correct = 1;
+  for (int i = 0; i < N; i++)
```
aktemur, line 46: We should use braces for the for-loop.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
```diff
@@ -0,0 +1,93 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# 'target teams distribute parallel for' should produce multiple GPU
+# waves all hitting the same breakpoint.
+
+load_lib rocm.exp
+
+require allow_omp_offload_tests
+
+standard_testfile .c
```
aktemur: ".c" is not needed.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
```diff
@@ -0,0 +1,93 @@
...
+    set saw_bp_hit 0
+    gdb_test_multiple "continue" "continue to teams breakpoint" {
+	-re "hit Breakpoint $::decimal" {
+	    set saw_bp_hit 1
+	    exp_continue
+	}
+	-re "$::gdb_prompt " {
+	    gdb_assert { $saw_bp_hit } $gdb_test_name
+	}
+    }
```
aktemur: Can't we use `gdb_continue_to_breakpoint`?

gdb/testsuite/gdb.rocm/omp-target-teams.exp
```diff
@@ -0,0 +1,93 @@
...
+
+    # Verify multiple waves are present.  We don't need an exact wave
+    # count - it depends on the chosen num_teams / thread_limit and
+    # GPU architecture - but there must be at least one.
+    set wave_list [info_thread_get_wave_list]
+    gdb_assert {[llength $wave_list] >= 1} "saw at least one AMDGPU wave"
+
+    if {[llength $wave_list] >= 1} {
```
aktemur: I'd remove this for the sake of simplicity. If there isn't at least one wave, the test above would fail anyway.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
```diff
@@ -0,0 +1,93 @@
...
+	    exp_continue
+	}
+	-re "$::gdb_prompt " {
+	    gdb_assert { $saw_bp_hit } $gdb_test_name
+	}
+    }
+
+    # Verify multiple waves are present.  We don't need an exact wave
+    # count - it depends on the chosen num_teams / thread_limit and
+    # GPU architecture - but there must be at least one.
```
aktemur: At least one or multiple?

gdb/testsuite/gdb.rocm/omp-target-teams.exp
```diff
@@ -0,0 +1,93 @@
...
+	set saw_switch 0
+	gdb_test_multiple "thread $first_wave" "switch to first AMDGPU wave" {
+	    -re "Switching to thread $first_wave," {
+		set saw_switch 1
+		exp_continue
+	    }
+	    -re "$::gdb_prompt " {
+		gdb_assert { $saw_switch } $gdb_test_name
+	    }
+	}
```
aktemur: A simple `gdb_test` should suffice here.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
```diff
@@ -0,0 +1,93 @@
...
+	gdb_test "print i >= 0" \
+	    "= (true|1)" \
+	    "loop index is non-negative"
```
aktemur: I don't think this is guaranteed. If we switch to the wave that got assigned to the first data chunk, the first lane's `i` would be 0.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
```diff
@@ -0,0 +1,93 @@
...
+    # GPU architecture - but there must be at least one.
+    set wave_list [info_thread_get_wave_list]
+    gdb_assert {[llength $wave_list] >= 1} "saw at least one AMDGPU wave"
+
+    if {[llength $wave_list] >= 1} {
+	# Switch to one of the waves and confirm the kernel-local
+	# index 'i' is a sensible non-negative integer.
+	set first_wave [lindex $wave_list 0]
+	set saw_switch 0
+	gdb_test_multiple "thread $first_wave" "switch to first AMDGPU wave" {
```
aktemur: Why do we need to switch? The breakpoint hit would've already put us in the wave context.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
```diff
@@ -0,0 +1,93 @@
...
+    set saw_exit 0
+    gdb_test_multiple "continue" "continue to exit" {
+	-re "exited normally" {
+	    set saw_exit 1
+	    exp_continue
+	}
+	-re "$::gdb_prompt " {
+	    gdb_assert { $saw_exit } $gdb_test_name
+	}
+    }
```
aktemur: Can't we use `gdb_continue_to_termination`?

gdb/testsuite/gdb.rocm/omp-target-teams.exp
```diff
@@ -0,0 +1,93 @@
...
+load_lib rocm.exp
+
+require allow_omp_offload_tests
+
+standard_testfile .c
+
+set break_line [gdb_get_line_number "teams-line"]
+
+if {[gdb_compile_omp_offload [list $srcdir/$subdir/$srcfile] \
+	 [standard_output_file $testfile] executable {debug}] != ""} {
```
aktemur: For HIP programs, we use the common `build_executable` but pass a build parameter "hip". Have you considered doing something similar? We can pass "omp", but note that amdclang is not the only omp compiler. However some compiler flags are also likely to be the same. So, maybe "omp" and "rocm" so that we set common OpenMP flags with "omp" and then we select the right compiler with "rocm". Rough idea to consider.

aktemur (commented):
I looked at the omp-target-teams test and rocm.exp only. I think we should first resolve infrastructure questions.

gdb/testsuite/lib/rocm.exp
lumachad: To speed things up a bit, you can add #134's patch/series to this PR so we only review close-to-final OpenMP GPU offload tests code. Otherwise #62 will be blocked waiting for #134 when it doesn't need to.

gdb/testsuite/lib/rocm.exp
lumachad: See my reasoning above.

gdb/testsuite/lib/rocm.exp
spatrang: Agreed — the Linux + `--with-amd-dbgapi` + `find_amdgpu_devices` checks are identical to `allow_hipcc_tests`. I'll factor the common part into a shared helper that both gates call. To keep #62 unblocked (per @lumachad's note), this shared helper lives here in #134 on top of #62's reworked checks, rather than being added to #62.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
spatrang: `rocm_omp_continue_to_breakpoint` was added to tolerate the multi-line stop report we observed for OMP-offload: an interleaved `Python Exception ...` line plus the auto-generated `__omp_offloading_*` frame between `hit Breakpoint` and the source line. That interleaving looks like an artifact of our test GDB missing its Python module rather than something intrinsic — if stock `gdb_continue_to_breakpoint` works against a properly installed build, I'll switch to it. I'll settle this together with the infra cleanup.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
spatrang: There's no `gdb_continue_to_termination` in our tree — the standard proc is `gdb_continue_to_end`. Same situation as the breakpoint helper: `rocm_omp_continue_to_exit` only exists to tolerate the multi-line output, so I'll evaluate replacing it with `gdb_continue_to_end` as part of the infra cleanup.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
spatrang: I like this. Mirroring the `hip` build option with an `omp` option (common `-fopenmp -fopenmp-targets=…` flags) plus a `rocm`/compiler-selection option would be cleaner than the bespoke `gdb_compile_omp_offload` wrapper, and it keeps the "which compiler" choice separate from "which language flags." I'll add this `omp`/`rocm` build option here in #134 on top of #62's reworked compiler discovery, so #62 stays unblocked.

spatrang:
@aktemur Agreed on settling the infrastructure. Per @lumachad's note about keeping #62 unblocked, I'll base #134 on top of #62's series and keep the OpenMP-offload infra + tests here so it's reviewed close-to-final, rather than holding it or pushing pieces into #62. The shared bits (common `omp`/`rocm` build option, `allow_*` reuse, search order) build on #62's foundation. The test-level cleanups (braces, `standard_testfile`, simplified teams block) are ready to apply.

gdb/testsuite/lib/rocm.exp
spatrang: Makes sense — I'll keep the OpenMP-offload series here in #134 and base it on top of #62's branch, so #62 stays independent (not blocked) and #134 is reviewed as close-to-final code. The OMP-specific bits (offload flags, the test helpers, the `allow_*` gate) stay here on top of #62's reworked compiler discovery/search-order foundation, rather than being pushed into #62.

gdb/testsuite/lib/rocm.exp
spatrang: Fixed — corrected the comment to use the proper `ROCm` spelling.

gdb/testsuite/gdb.rocm/omp-target-teams.c
spatrang: Done — added braces to the for-loop bodies in `omp-target-teams.c`.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
spatrang: Done — dropped the redundant `.c`; it is now just `standard_testfile`.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
spatrang: Agreed — removed the explicit thread switch. The breakpoint hit already puts us in the wave context, so we read `i` directly there.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
spatrang: You're right that a lane's `i` can be 0. The check is now just `print i >= 0`, which holds for any valid lane (0 included), so it stays correct without assuming a non-zero index.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
spatrang: Done — this is now a plain `gdb_test` for `print i >= 0`; the surrounding `gdb_test_multiple`/thread-switch is gone.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
spatrang: At least one — reworded both the comment and the assertion to say "at least one AMDGPU wave" to match the `>= 1` check.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
spatrang: Removed — dropped the `if {[llength $wave_list] >= 1}` guard; the `>= 1` assertion above already covers the no-wave case.

spatrang:
## Changelog — `6ce4ae405be` (gdb/testsuite: filter benign OpenMP-offload linker warnings)

Pushed a follow-up commit addressing the review feedback on the OpenMP-offload tests.

**Test infrastructure (`gdb/testsuite/lib/rocm.exp`)**
- Added `rocm_omp_prune_benign_warnings` to drop the known, harmless ROCm 7.14 `ld.lld` `__keep_alive` "local memory global used by non-kernel function" warning. The match is anchored to a single line so genuine compiler/linker errors are never swallowed.
- Added `rocm_omp_compile`, which compiles quietly, prunes the benign warning, and still logs any remaining output via `verbose -log` so real failures stay diagnosable.
- Routed `gdb_compile_omp_offload{,_cpp,_fortran}` through `rocm_omp_compile`.
- Dropped the `with_rocm_omp_gpu_lock` alias; tests call `with_rocm_gpu_lock` directly.
- Fixed `rocm` → `ROCm` spelling in comments.

**Tests (`gdb.rocm/omp-target-*`)**
- Switched the `.exp` files to `with_rocm_gpu_lock` and `standard_testfile` (no redundant `.c`).
- `omp-target-teams.c`: added braces to the for-loop bodies for stable breakpoint line numbers.
- `omp-target-teams.exp`: removed the explicit wave thread-switch (the breakpoint hit already selects the wave context), dropped the redundant `>= 1` guard, reworded the comments/assertion to "at least one AMDGPU wave", and kept a plain `gdb_test` for `print i >= 0` (valid for any lane, including lane 0).

**Why the warning filter**
The binary builds and runs correctly on ROCm 7.14; only the stderr warning tripped DejaGnu into marking `omp-target-teams.exp` `UNTESTED`. With the filter the test now runs.

**Verification**
- ROCm 7.14.0: all 6 `omp-target-*` tests pass (39 PASS, 0 FAIL, 0 UNTESTED).
- ROCm 7.13.0: all 6 pass (39 PASS), no regression.

> Note: the compiler-discovery and `omp`/`rocm` build-option items remain tracked against the #62-based rework, as discussed in the resolved threads.

lumachad:
Set to @aktemur as he left the earlier comments.

spatrang:
## Rebased on amd-staging + review fixes (`a5e145b6d9e`)

Rebased onto latest `amd-staging` (the hipcc→amdclang++ switch), squashed into one commit, and applied review feedback.

**Rebase**
- Clean rebase; renamed `allow_hipcc_tests` → `allow_hip_tests` reference.
- OpenMP tests use their own `amdclang` discovery, so they're unaffected by the HIP board change.

**Review fixes**
- `allow_omp_offload_tests` now reuses `allow_hip_tests` for the shared GPU/dbgapi gate instead of duplicating it.
- Dropped the dead `f77` clause in `gdb.exp`.
- Prune regex now also drops trailing `>>>` lld note lines (still anchored to `__keep_alive`).
- Reworded `rocm_omp_compile` log to "compile output".
- Removed unused `sq-line`/`cb-line` markers.
- Documented why the custom `rocm_omp_continue_*` helpers exist.

**Note on `\r\n` anchoring:** skipped — `gdb_test` here already appends `\r\n$gdb_prompt`, so adding it broke every value test. The implicit anchor already covers the `check-read1` concern.

**Testing:** ROCm 7.14 — all 6 tests pass (40 passes, 0 failures).

gdb/testsuite/lib/rocm.exp
aktemur: > dropping rocm_find_llvm_tool and the find_amd* helpers. Leaving them for now only so the suite is runnable standalone until #62 lands; I'll remove them in the rebase.

#62 is merged but we still have the "..find.." procs. Is this an oversight?

gdb/testsuite/gdb.rocm/omp-target-basic.c
```diff
@@ -0,0 +1,59 @@
...
+	      map(to:in_arr[0:N]) map(tofrom:out_arr[0:N])	\
+	      firstprivate(firstpriv_val)
+  {
+    int priv_val = 7;			/* first-target-line */
+    int step_two = priv_val + 1;
+    int step_three = step_two + 1;
+
+    for (int i = 0; i < N; i++)
+      out_arr[i] = in_arr[i] + priv_val + firstpriv_val
+		   + step_two + step_three;
```
aktemur: Nit: '+' should be indented two spaces inside wrt `out_arr`. A preferred alternative is to parenthesize the arithmetic expression, like this:
```
    for (int i = 0; i < N; i++)
      out_arr[i] = (in_arr[i] + priv_val + firstpriv_val
		    + step_two + step_three);
```

gdb/testsuite/gdb.rocm/omp-target-basic.exp
```diff
@@ -0,0 +1,81 @@
...
+
+with_rocm_gpu_lock {
+    if {![runto_main]} {
+	return
+    }
+
+    gdb_breakpoint "$srcfile:$first_line" allow-pending
+
+    with_test_prefix "breakpoint" {
+	rocm_omp_continue_to_breakpoint "continue to device breakpoint"
```
aktemur: I don't think using custom procs for these GDB commands will scale well. Can we please try to use the usual `gdb_breakpoint`, `gdb_test`, etc. procs? This applies to procs like `rocm_omp_check_on_wave`, `rocm_omp_next, `rocm_omp_continue_to_exit`, etc.

gdb/testsuite/gdb.rocm/omp-target-basic.exp
```diff
@@ -0,0 +1,81 @@
...
+	gdb_test "print in_arr\[0\]" "= 100" "mapped-to in_arr\[0\]"
+	gdb_test "print in_arr\[7\]" "= 107" "mapped-to in_arr\[7\]"
+	gdb_test "print out_arr\[0\]" "= 0" \
+	    "mapped-tofrom out_arr\[0\] before fill"
+    }
+
+    # Step forward through a few statements; each step must stay on a
+    # GPU wave.  Line numbers are not pinned because OpenMP-offload
+    # line tables vary across compiler versions.
+    with_test_prefix "step" {
```
aktemur: We don't need this prefix; there is `next-$which` below already.

gdb/testsuite/gdb.rocm/omp-target-basic.exp
```diff
@@ -0,0 +1,81 @@
...
+load_lib rocm.exp
+
+require allow_omp_offload_tests
+
+standard_testfile
+
+set first_line [gdb_get_line_number "first-target-line"]
+
+if {[gdb_compile_omp_offload [list $srcdir/$subdir/$srcfile] \
+	 [standard_output_file $testfile] executable {debug}] != ""} {
```
aktemur: My first expectation would be that we use the usual `build_executable`/`prepare_for_testing` by passing it the proper option (e.g. `omp rocm`), like we do for HIP tests. Then the testing framework finds the right compiler and sets the right flags, again similar to how HIP testing is done.

gdb/testsuite/lib/rocm.exp
```diff
@@ -679,3 +679,302 @@ gdb_caching_proc compiler_dwarf_language {} {
...
+    regsub -all \
+	{ld\.lld: warning:[^\r\n]*__keep_alive[^\r\n]*local memory global used by non-kernel function[^\r\n]*[\r\n]*(?:[ \t]*>>>[^\r\n]*[\r\n]*)*} \
+	$output "" output
+    regsub -all {^[\r\n]+|[\r\n]+$} $output "" output
+    return $output
+}
+
+# Compile via gdb_compile, prune known-harmless warnings, and log any
+# remaining output (gdb_compile is quiet so benign noise is suppressed).
+proc rocm_omp_compile {source dest type options} {
```
aktemur: By passing the proper omp and rocm options to `gdb_compile`, I think all can be handled in `gdb_compile`. So, this custom proc and the related ones below can be removed.

gdb/testsuite/lib/rocm.exp
```diff
@@ -679,3 +679,302 @@ gdb_caching_proc compiler_dwarf_language {} {
...
+	set F90_FOR_TARGET $amdflang
+	set F77_FOR_TARGET $amdflang
+	lappend options f90
+	lappend options "early_flags=[rocm_omp_offload_flags fortran]"
+	set result [rocm_omp_compile $source $dest $type $options]
+    }
+    return $result
+}
+
+# Helpers shared by the gdb.rocm/omp-target-*.exp tests.
```
aktemur: Ideally, the custom procs should all go away and we should be able to use the usual `gdb_...` procs in `.exp` files.

aktemur (changes requested):

spatrang:
@aktemur Thank you for the recent review.

On moving the OpenMP-offload tests to `build_executable`/`prepare_for_testing` with an `omp rocm`-style build option (mirroring how HIP works): agreed, I'll do that. One design question before I implement it.

Since amdclang isn't the only OpenMP compiler and the driver differs by language (C -> amdclang, C++ -> amdclang++, Fortran -> amdflang), would you prefer `omp` to set the common flags (`-fopenmp -fopenmp-targets=amdgcn-amd-amdhsa`, `--offload-arch=...`) and `rocm` to select the AMD compiler, with the language coming from the usual `c++` / `f90` options? I'll follow whatever option split you think reads best.

The remaining comments are clear and I'll address them in the next revision (fold the compile wrappers into `gdb_compile`, drop the now-redundant `find_*` helpers since #62 is merged, switch to the stock `gdb_breakpoint`/`gdb_test`/`gdb_continue_to_breakpoint`/`gdb_continue_to_termination` procs, remove the redundant `with_test_prefix`, and fix the `omp-target-basic.c` indentation).

aktemur:
> Since amdclang isn't the only OpenMP compiler and the driver differs by language (C -> amdclang, C++ -> amdclang++, Fortran -> amdflang), would you prefer `omp` to set the common flags (`-fopenmp -fopenmp-targets=amdgcn-amd-amdhsa`, `--offload-arch=...`) and `rocm` to select the AMD compiler, with the language coming from the usual `c++` / `f90` options? I'll follow whatever option split you think reads best.

Hi @spatrang, as the first approach, I'd suggest having two arguments: (1) language, (2) `omp-rocm`. Please take this as a mere suggestion and not as a "must". If you see a better split, we can certainly evaluate. I think we should avoid over-engineering and prefer simplicity. Splitting omp-rocm into two arguments, omp and rocm, may complicate the flow unnecessarily, because then we need to think about having "rocm" but no "omp" and vice-versa, too.

gdb/testsuite/gdb.rocm/omp-target-basic.c
spatrang: Done — parenthesized the expression as suggested.

gdb/testsuite/gdb.rocm/omp-target-basic.exp
spatrang: Done — switched to stock gdb_continue_to_breakpoint / gdb_test / gdb_continue_to_end and removed the custom procs.

gdb/testsuite/gdb.rocm/omp-target-basic.exp
spatrang: Done — removed the redundant prefix; next-$which already disambiguates.

gdb/testsuite/gdb.rocm/omp-target-basic.exp
spatrang: Done — now built via prepare_for_testing with a new omp-rocm build option, like HIP.

gdb/testsuite/lib/rocm.exp
spatrang: Done — folded into gdb_compile via the omp-rocm option; the custom wrappers are removed.

gdb/testsuite/lib/rocm.exp
spatrang: Done — all custom run/stop procs are removed; the tests use the stock gdb_* procs now.

spatrang:
## Reworked to use the standard testsuite framework

Addressed the latest review by moving the OpenMP-offload tests onto the usual framework instead of bespoke helpers.

**What changed**
- Added an `omp-rocm` build option to `gdb_compile`. It picks the AMD compiler by language (amdclang / amdclang++ / amdflang), adds the `-fopenmp` offload flags, and prunes a benign ROCm `ld.lld` warning so the build is not wrongly marked UNTESTED.
- Tests now build via `prepare_for_testing` / `build_executable` with `{... omp-rocm}`, like HIP.
- Removed the bespoke compile wrappers (`gdb_compile_omp_offload*`) and the `find_amd*` discovery; discovery reuses the framework (`#62`) helpers.
- Removed the custom run/stop procs; tests now use stock `gdb_continue_to_breakpoint`, `gdb_test`, `gdb_continue_to_end`, etc.
- Small nits: dropped a redundant `with_test_prefix`; parenthesized an expression in `omp-target-basic.c`.

**Testing**
- Built a fresh rocgdb tree and ran `.github/scripts/test_rocgdb.py` on ROCm 7.14: all six `omp-target-*` tests pass (GCC 51 PASS, LLVM 51 PASS, OVERALL PASS).

Will force-push the updated single commit shortly.

spatrang:
## Fixed CI pre-commit (codespell)

The pre-commit `codespell` hook flagged a pre-existing typo in `gdb/testsuite/lib/rocm.exp` (`successfuly` -> `successfully`); since this commit already modifies that file, codespell scanned it and failed. Fixed the typo and folded it into the commit.

Also noted the `gdb.exp` change in the commit log: it skips the GCC `-fno-stack-protector` workaround for non-GNU Fortran compilers (flang / amdflang / ifx), which reject it; gfortran is unaffected.

No functional change to the tests. Will force-push the updated commit.

gdb/testsuite/gdb.rocm/omp-target-basic.c
```diff
@@ -0,0 +1,59 @@
...
+	      map(to:in_arr[0:N]) map(tofrom:out_arr[0:N])	\
+	      firstprivate(firstpriv_val)
+  {
+    int priv_val = 7;			/* first-target-line */
+    int step_two = priv_val + 1;
+    int step_three = step_two + 1;
+
+    for (int i = 0; i < N; i++)
+      out_arr[i] = (in_arr[i] + priv_val + firstpriv_val
+		    + step_two + step_three);
```
aktemur: We stop at line 48 and then do 3 "next"s. Is there any reason to have this loop? It could as well be a simple statement, too.

gdb/testsuite/gdb.rocm/omp-target-cpp.exp
```diff
@@ -0,0 +1,67 @@
...
+    if {![runto_main]} {
+	return
+    }
+
+    with_test_prefix "templated add_t" {
+	gdb_breakpoint "$srcfile:$tmpl_line" allow-pending
+	gdb_continue_to_breakpoint "add_t<int>"
+	gdb_assert {[llength [info_thread_get_wave_list]] >= 1} \
+	    "AMDGPU wave at add_t"
+	delete_breakpoints
```
aktemur, line 50: We can add a caller frame between the kernel and "add_t" and then check "backtrace 2". This should show add_t and its caller.

gdb/testsuite/gdb.rocm/omp-target-cpp.exp
```diff
@@ -0,0 +1,67 @@
...
+	gdb_continue_to_breakpoint "multiplier::operator()"
+	gdb_assert {[llength [info_thread_get_wave_list]] >= 1} \
+	    "AMDGPU wave at functor"
+
+	# 'this->factor' should be the value we set on the host (3),
+	# transferred via the implicit 'map(to:m)'.
+	gdb_test "print this->factor" \
+	    "= 3" \
+	    "factor visible via 'this'"
+	delete_breakpoints
```
aktemur, line 66: Again, we can check the call stack.

gdb/testsuite/gdb.rocm/omp-target-data.exp
```diff
@@ -0,0 +1,71 @@
...
+    if {![runto_main]} {
+	return
+    }
+
+    gdb_breakpoint "$srcfile:$k1_line" allow-pending
+
+    with_test_prefix "first kernel" {
+	gdb_continue_to_breakpoint "first kernel"
+	gdb_assert {[llength [info_thread_get_wave_list]] >= 1} \
+	    "AMDGPU wave at first kernel"
```
aktemur: We did these checks in basic tests. I'm not sure what we gain by repeating the check at each breakpoint hit. This check only tells us there is at least one wave. It doesn't say anything about their state or location or weather the current thread is a wave. I'd just remove them.

gdb/testsuite/gdb.rocm/omp-target-data.exp
```diff
@@ -0,0 +1,71 @@
...
+	gdb_test "print a\[0\]" "= 0"   "a\[0\] in first kernel"
+	gdb_test "print b\[0\]" "= 100" "b\[0\] in first kernel"
+    }
+
+    # The kernel's loop body would fire the first breakpoint many
+    # more times.  Drop it and arm the next one only.
+    delete_breakpoints
```
aktemur: We can define the first breakpoint to be temporary and remove these comments and `delete_breakpoints`.

gdb/testsuite/gdb.rocm/omp-target-fortran.f90
```diff
@@ -0,0 +1,46 @@
...
+
+  do i = 1, n
+     a(i) = i
+     b(i) = 2 * i
+     c(i) = 0
+  end do
+
+  !$omp target map(to:a, b) map(tofrom:c)
+  do i = 1, n
+     c(i) = a(i) + b(i)             ! ftarget-line
```
aktemur: I wanted to see what happens at this breakpoint. I thought the loop would be vectorized, but it isn't. There is one lane active and it runs the loop. Did we use the right omp pragma? Furthermore, variables a, b and c are available, but i and n are not. We may need to report this.

gdb/testsuite/gdb.rocm/omp-target-multi-kernel.exp
```diff
@@ -0,0 +1,60 @@
...
+	return
+    }
+
+    with_test_prefix "square kernel" {
+	gdb_breakpoint "square_dev" allow-pending
+	gdb_continue_to_breakpoint "square_dev"
+	gdb_assert {[llength [info_thread_get_wave_list]] >= 1} \
+	    "AMDGPU wave at square_dev"
+	# Kernel iterates many times; clear the bp before moving on.
+	delete_breakpoints
```
aktemur: Here and below, we can define the BP to be temporary.

gdb/testsuite/gdb.rocm/omp-target-multi-kernel.exp
```diff
@@ -0,0 +1,60 @@
...
+}
+
+with_rocm_gpu_lock {
+    if {![runto_main]} {
+	return
+    }
+
+    with_test_prefix "square kernel" {
+	gdb_breakpoint "square_dev" allow-pending
+	gdb_continue_to_breakpoint "square_dev"
```
aktemur, line 46: Let's also print the argument x. Below in `cube_dev`, too.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
```diff
@@ -0,0 +1,62 @@
...
+    # The breakpoint hit already selects an AMDGPU wave; confirm the
+    # teams-distributed loop produced at least one wave at the stop
+    # (see the note at the top of the file about not asserting an
+    # exact count).
+    gdb_assert {[llength [info_thread_get_wave_list]] >= 1} \
+	"at least one AMDGPU wave at teams breakpoint"
+
+    gdb_test "print i >= 0" \
+	"= (true|1)" \
+	"loop index is non-negative"
```
aktemur: This test doesn't look strong enough. How about this: read the value of i here. Then switch to lane 7. Read again. The difference between the two i values should be 7. Also print `a[i] + b[i]`. It should be 256.

aktemur (commented):
I have minor feedback.

gdb/testsuite/gdb.rocm/omp-target-basic.c
spatrang: Good point — the loop wasn't needed; the checks don't rely on it. Replaced it with a simple statement that still uses the mapped and firstprivate values. Done.

gdb/testsuite/gdb.rocm/omp-target-cpp.exp
spatrang: Done — added a `sum_pair` caller and now check `backtrace 2` shows `add_t` and `sum_pair`.

gdb/testsuite/gdb.rocm/omp-target-cpp.exp
spatrang: Done — added a `scale` caller and check `backtrace 2` shows the functor and `scale`.

gdb/testsuite/gdb.rocm/omp-target-multi-kernel.exp
spatrang: Done — made the breakpoints temporary, so the manual `delete_breakpoints` are gone.

gdb/testsuite/gdb.rocm/omp-target-data.exp
spatrang: Done — made the breakpoints temporary and removed the comments and `delete_breakpoints`.

gdb/testsuite/gdb.rocm/omp-target-data.exp
spatrang: Agreed — removed the repeated wave checks here. The test now just verifies the buffer reuse across the two kernels (a[0], b[0], c[0]).

gdb/testsuite/gdb.rocm/omp-target-multi-kernel.exp
spatrang: Done — now `print x` at both `square_dev` and `cube_dev` (it's 1 for the first lane). Replaced the shallow wave check with this, since reading the device argument also confirms we're on a wave.

gdb/testsuite/gdb.rocm/omp-target-fortran.f90
spatrang: Thanks for digging into this. I investigated, and `i`'s invisibility isn't a fundamental amdflang DWARF gap — it's the optimization level:

- amdclang/amdclang++ default to `-O0` for device code, which is why the C/C++ tests show locals fine.
- amdflang optimizes device code by default, so the loop index `i` is optimized into a register without a debug location, and GDB reports "No symbol i".
- Compiling the offload region with `-O0` makes it visible: `print i` then works (`= 32`).

On the other points:
- Single lane / not parallelized — expected for a plain `!$omp target`: it offloads the region but has no worksharing, so the loop runs sequentially on one device thread (the Fortran analog of `omp-target-basic`). I tried `!$omp target teams distribute parallel do` — it does spread across lanes [0-31], but the stop then lands in an outlined `..omp_par` frame where even `a`/`b`/`c` aren't readable, so it's worse for a debugger test.
- `n` not available — `n` is a Fortran `parameter` (compile-time constant), so it's folded and not emitted as a runtime variable; that part is expected.

So the only real action is `i`'s visibility. How would you prefer to handle it?
1. Force `-O0` for the amdflang offload build in the test harness (only amdflang needs it; the clang drivers already default to `-O0`), so device locals are debuggable; or
2. Treat amdflang's default-optimization-dropping-locals as a compiler issue to report upstream, and keep the test asserting only what's reliably available.

I lean towards (1) for the test, but happy to follow your preference.

gdb/testsuite/gdb.rocm/omp-target-teams.exp
spatrang: Good call — strengthened it as you suggested. It now reads `i` on lane 0, switches to lane 7, reads `i` again, and asserts the difference is 7 (consecutive iterations land on consecutive lanes), plus `print a[i] + b[i]` is checked to be 256. Done.

gdb/testsuite/gdb.rocm/omp-target-fortran.f90
aktemur: I'd also favor option 1: Force -O0 for device code.

Also, when optimizations are applied, `i` should not disappear. It should still be recognized and listed in "info locals", but as `<optimized out>`. This would require the compiler to still emit DWARF for the variable but omit the location attribute.

spatrang: Thanks. I kept the f90 test basic and added a backtrace since it works. It now stops inside an `!$omp declare target` function, checks `backtrace 2` (the function called from the kernel) and the arguments, then reads the loop index `i` from the kernel frame. To make device locals visible, the test compiles with `-O0` (just this test, via `additional_flags`, not the shared flags) — the amdflang default-optimization debug-info gap is already tracked as ROCM-19898.

spatrang:
## Strengthened the tests per review (rebased on latest amd-staging)

Addressed the latest round of comments by making the assertions verify concrete debugger behavior rather than just "a wave exists".

**What changed (tests only; no shared-lib changes)**
- `omp-target-teams`: read the loop index on lane 0 and lane 7 and assert the difference is 7 (consecutive iterations on consecutive lanes), plus `a[i] + b[i] == 256`.
- `omp-target-cpp`: added `sum_pair`/`scale` device callers and assert `backtrace 2` shows the callee and its caller.
- `omp-target-multi-kernel`: print the device argument `x` at each kernel; breakpoints made temporary.
- `omp-target-data`: dropped the redundant wave checks (kept the buffer-reuse value checks); breakpoints made temporary.
- `omp-target-basic`: replaced the in-region loop with a simple statement (the stepping test didn't need it).
- `omp-target-fortran`: restructured to call a `!$omp declare target` function so it can check `backtrace 2` and the arguments, plus read the loop index. Built with `-O0` for this test only (amdflang optimizes device code by default and drops device-local debug info; tracked upstream as ROCM-19898 — amdclang already defaults to -O0).

**Rebase**
- Rebased onto the latest `amd-staging`. This surfaced a framework change: `gdb_breakpoint` now takes dashed flags (`parse_args`), so updated the calls to `-temporary -allow-pending` / `-allow-pending`.

**Testing**
- ROCm 7.14, freshly rebuilt tree: all six `omp-target-*` tests pass (51 PASS, 0 FAIL, 0 UNTESTED).

gdb/testsuite/gdb.rocm/omp-target-basic.exp
```diff
@@ -0,0 +1,83 @@
...
+	    # Require landing on a numbered source line (so a step that
+	    # errors out is not silently accepted); the exact line is
+	    # not pinned because OpenMP-offload line tables vary.
+	    gdb_test "next" "$::decimal\t\[^\r\n\]*" "next inside target region"
+	    gdb_assert {[llength [info_thread_get_wave_list]] >= 1} \
+		"still on AMDGPU wave"
+	}
+    }
+
+    delete_breakpoints
```
aktemur: I believe we don't need this.

gdb/testsuite/gdb.rocm/omp-target-fortran.exp
```diff
@@ -0,0 +1,72 @@
...
+with_rocm_gpu_lock {
+    if {![runto_main]} {
+	return
+    }
+
+    # Stop inside the '!$omp declare target' function and confirm the
+    # device call stack shows it called from the offloaded kernel.
+    gdb_breakpoint "$srcfile:$add_line" -temporary -allow-pending
+    gdb_continue_to_breakpoint "add_vals"
+    gdb_test "backtrace 2" \
```
aktemur: `backtrace 1` would suffice here. `__omp_offloading` is a generated name. I don't think we need to include it in the test.

gdb/testsuite/gdb.rocm/omp-target-multi-kernel.exp
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    # This is a serial 'omp target' loop, so the first iteration
+    # (i == 0) handles in[0] == 1; the device argument x is therefore 1
+    # in both kernels.  Reading it confirms we stopped in the right
+    # device function with a live argument (which also implies a wave).
+    with_test_prefix "square kernel" {
+	gdb_breakpoint "square_dev" -temporary -allow-pending
+	gdb_continue_to_breakpoint "square_dev"
+	gdb_test "print x" "= 1" "square_dev argument x"
```
aktemur: Nit: It'd be ok to not give a test name (there is already a prefix) so that the test name appears simply as "print x". Also in the "print x" test below.

gdb/testsuite/gdb.rocm/omp-target-teams.c
```diff
@@ -0,0 +1,56 @@
...
+      b[i] = N - i;
+      c[i] = 0;
+    }
+
+  #pragma omp target teams distribute parallel for	\
+	      map(to:a[0:N], b[0:N]) map(tofrom:c[0:N])
+  for (int i = 0; i < N; i++)
+    {
+      c[i] = a[i] + b[i];	/* teams-line */
+    }
```
aktemur: No braces needed here.

aktemur (approved):

gdb/testsuite/gdb.rocm/omp-target-basic.exp
spatrang: Right, removed it — the region runs once, so there's no re-hit before exit.

gdb/testsuite/gdb.rocm/omp-target-fortran.exp
spatrang: Done — switched to `backtrace 1` and dropped the generated `__omp_offloading` name; it now just checks we stopped in `add_vals`.

gdb/testsuite/gdb.rocm/omp-target-multi-kernel.exp
spatrang: Done — dropped the explicit names; they now read as "print x" under each kernel's prefix.

gdb/testsuite/gdb.rocm/omp-target-teams.c
spatrang: Done — removed the braces.

spatrang:
## Addressed the latest cosmetic comments (rebased on latest amd-staging)

- `omp-target-basic`: removed the unneeded `delete_breakpoints` before exit (the region runs once, so there's no re-hit).
- `omp-target-fortran`: simplified to `backtrace 1` and dropped the generated `__omp_offloading` name — it now just checks we stopped in `add_vals`.
- `omp-target-multi-kernel`: dropped the explicit `print x` test names (the kernel prefix already disambiguates).
- `omp-target-teams.c`: removed the braces around the single-statement kernel loop.

Rebased onto the latest `amd-staging`.

Testing (ROCm 7.14, freshly rebuilt tree): all six `omp-target-*` tests pass (51 PASS, 0 FAIL, 0 UNTESTED).

lumachad:
The tests are not running actually...

```
gdb compile failed, ld.lld: error: unable to find library -lompdevice
clang: error: ld.lld command failed with exit code 1 (use -v to see invocation)
/__w/ROCgdb/ROCgdb/build/lib/llvm/bin/clang-linker-wrapper: error: 'clang' failed
flang-23: error: linker command failed with exit code 1 (use -v to see invocation)
```

lumachad:
Blocked on:
- [ ] https://github.com/ROCm/TheRock/pull/6203

lumachad:
https://github.com/ROCm/TheRock/pull/6203 now merged, but we'll need to bump TheRock's ref in rocgdb to pick up the fix first.

lumachad:
Once this one lands, we should be good to rebase and try this: https://github.com/ROCm/ROCgdb/pull/194

lumachad:
Rebased. Let's see if the OMP tests run this time around, though I expect we will run into build issues due to upstream churn.

spatrang:
CI's "Build Linux Packages" is failing on an unrelated core-gdb compile error. After the rebase, the newer amd-llvm toolchain promotes `-Wpessimizing-move` to `-Werror` and breaks `gdb/record-full.c` (two `std::move` on temporaries around lines 2386/2393). "Run Tests" is skipped because the package build never completes.

This needs a fix on amd-staging (drop the pessimizing `std::move`s) or a toolchain/TheRock adjustment; I'll rebase once that lands.

spatrang:
Rebased onto the latest `amd-staging` to pick up the recent CI/container and testsuite fixes (clean rebase, no conflicts).

lumachad (approved):
LGTM.

PR #135, opened 2026-05-21
gdb/testsuite: make gdb.rocm/runtime-core.exp test lightweight cores

motokultivator:
The ROCr library has the ability to generate lightweight GPU core dumps. These core dumps omit some memory areas, such as whatever is allocated with hipMalloc.  Update gdb.rocm/runtime-core.exp to test them.

Modify gdb.rocm/runtime-core.cpp to allocate some memory on the device and fill it with known values.  Pass a pointer to that memory the kernels that generate the faults leading to the core dumps.  A full core dump will include this memory, while a lightweight core dump will not. Modify gdb.rocm/runtime-core.exp to try to access that memory.

Coauthored-By: Aleksandar Rikalo <arikalo@amd.com>


gdb/testsuite/gdb.rocm/runtime-core.cpp
```diff
@@ -128,6 +128,22 @@ main (int argc, char **argv)
...
+  {
+    constexpr int n_elements = 1000;
+    std::unique_ptr<int[]> host_data = std::make_unique<int[]> (n_elements);
+
+    for (int i = 0; i < n_elements; i++)
+      host_data[i] = i;
+
+    CHECK (hipMalloc (&device_data, n_elements * sizeof (int)));
+    CHECK (hipMemcpy (device_data, host_data.get (), n_elements * sizeof (int),
+		      hipMemcpyDefault));
```
lancesix: Could ad well use `hipMemcpyHostToDevice` as we statically know the copy direction.

gdb/testsuite/gdb.rocm/runtime-core.cpp
```diff
@@ -128,6 +128,22 @@ main (int argc, char **argv)
   hipStream_t st1;
   hipStream_t st2;
 
+  /* Allocate some data on the device.  This data is expected to not be
+     available in lightweight cores.  */
+  int *device_data;
+  {
+    constexpr int n_elements = 1000;
+    std::unique_ptr<int[]> host_data = std::make_unique<int[]> (n_elements);
```
lancesix: Just on a style perspective, it might be a bit personal, but this is one case where `auto` can be used to remove repeat while keeping things clear.

```c++
auto host_data = std::make_unique<int[]> (n_elements);
```

And I guess the canonical container here would probably be a vector:

```c++
    std::vector<int> host_data (n_elements);
    std::iota (host_data.begin (), host_data.end (), 0);
```

lancesix (approved):
Mostly stylistic remarks, but overall looks good to me, thanks.

I don't really think we can detect an old runtime which would not support lightweight core dumps, so not sure we can reliably do a a XFAIL in such configuration.

gdb/testsuite/gdb.rocm/runtime-core.cpp
motokultivator: done.

gdb/testsuite/gdb.rocm/runtime-core.cpp
motokultivator: done.

PR #138, opened 2026-05-21
[don't merge] rocgdb multi-arch CI

lumachad:
Draft for rocgdb's multi-arch CI.


.github/scripts/therock_matrix.py
geomin12, line 1: this will actually be determined in TheRock

geomin12 (commented):
thanks for doing this! i think we would eventually keep the legacy CI on workflow_dispatch and then add that multi-arch. we actually have it as a nightly run right now but once we have certain features in TheRock included, we'll slowly roll to postsubmit then gating presubmit

lumachad:
Needs to be reworked based on the latest developments.

PR #139, opened 2026-05-25
gdb/testsuite: Extend gdb.rocm/runtime-core.exp with piped GPU coredumps

motokultivator:
The ROCr runtime had recently introduced the HSA_COREDUMP_PATTERN to configure how/where to save a core dump.  It follows the kernel model for kernel.core_pattern.  The pattern can either describe a file name where the core dump should be created, or if it starts with |, it indicates that the core dump should be sent through a pipe to a custom command.

This change extends the runtime-core.exp test case to exercise the piped version of core dump.

### DIFF
--- a/gdb/testsuite/gdb.rocm/runtime-core.exp
+++ b/gdb/testsuite/gdb.rocm/runtime-core.exp
@@ -33,17 +33,22 @@ if { [build_executable "failed to prepare" ${testfile} ${srcfile} \
     return -1
 }
 
-# do_test FAULT CORE_TYPE
-#   Run the test exercising generation of a core dump on fault FAULT.
+# do_test FAULT CORE_TYPE OUTPUT_TYPE
+#   Run the test exercising generation of a core dump on fault FAULT
+#   using a file or a pipe.
 #
 # FAULT is one of "page", "abort",  or "assert", and selects which
 #   fault the application should generate.
 #
 # CORE_TYPE is one of "full" or "lightweight".
-proc do_test { fault core_type } {
+#
+# OUTPUT_TYPE is one of "file" of "pipe".
+#
+proc do_test { fault core_type output_type } {
+    set use_pipe [expr {$output_type eq "pipe"}]
     save_vars { ::env(HSA_ENABLE_LIGHTWEIGHT_COREDUMP) } {
 	set ::env(HSA_ENABLE_LIGHTWEIGHT_COREDUMP) [expr {$core_type == "lightweight"}]
-	set coredump [rocm_core_find $::binfile {} $fault]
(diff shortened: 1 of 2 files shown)


lumachad (commented):
Looks OK to me.

lumachad:
Just a note, the CI doesn't run any of the core file tests at the moment, so validation needs to take place manually.

I ran this test and it seems on the machine I tried it is taking almost 5 minutes to run. Whenever we decide to parallelize the gdb.rocm testsuite, this will be a bottleneck. Could we please split these variations of tests into distinct files so eventually concurrency can be applied to this test?

Also, the testsuite is detecting unexpected core files. This should be fixed:

```
Running /home/lumachad/tmp/therock/build/dist/rocm/tests/rocgdb/gdb/testsuite/gdb.rocm/runtime-core.exp ...

=== gdb Summary ===                                                                                                                                                                                                                                                     
# of unexpected core files      12
# of expected passes            148
```

lancesix:
> I ran this test and it seems on the machine I tried it is taking almost 5 minutes to run. 

Indeed, it might not be necessary to exercise all kind of fault for all combination of lightweight / full or pipe / regular core dump.

We could have

```tcl
with_rocm_gpu_lock {
    # Check the runtime can configure the core dump generation based
    # on env variables.
    foreach_with_prefix output_type { file pipe } {
	foreach_with_prefix core_type {full lightweight} {
            do_test pagefault $core_type $output_type
        }
    }


    # Check that various GPU events can cause a core dump.
    # "pagefault" was covered by previous cases already.
    foreach_with_prefix fault {abort assert} {
      do_test $fault lightweight file
    }
}
```

which would reduce the number of coredump by 3.

lumachad:
I think the unexpected core files are unrelated to this (or the other runtime-core) change.

lumachad:
Just an example of what we'd like to keep. The old tests ran for about 73 seconds. So a little over a minute.

gdb/testsuite/gdb.rocm/runtime-core.exp
lancesix: This should also be updated to refrect the core dump generation method

```
    set corename "${coredump}-${fault}-${core_type}-${output_type}"
```

lancesix (changes requested):

gdb/testsuite/gdb.rocm/runtime-core.exp
motokultivator: done.

motokultivator:
@lancesix @lumachad 
I reduced the number of core dumps as you suggested.
"unexpected core files" is another issue, independent of this change. I will prepare another PR related to that.
I tested manually, of course.

lancesix (approved):
This LGTM, thanks.

PR #141, opened 2026-05-26
gdb/dwarf: Fix dwarf_register::read overflow

lancesix:
The recent change f204e6a13ac "gdb/dwarf2/expr: Cache the register
reads" introduced a regression where the DWARF expression evaluator can
end-up reading past the end of a buffer.  This can happen if we try to
read past the end of a register.

Prior to this change, all register reads ended up calling into
dwarf2/expr.c:read_from_register which has this guard:

    /* If a register is wholly inside the OFFSET, skip it.  */
    if (frame == NULL || !regsize
        || offset + length > regsize || numregs < regnum)
      {
        *optimized = false;
        *unavailable = true;
        return;
      }

The refactored code assumes that the read is always within the bounds of
the register, but this does not has to be true.

This patch fixes this issue by doing the necessary bounds check.  If the
read tries to go past the end of the register, we consider the
inaccessible part has been optimized out (those bits do not exist).
Because we do not have a better granularity, we report the entire value
read from the register as optimized out, not just the bits past the end
of the register.

This patch also adds a test that exercises the issue.

### DIFF
--- a/gdb/dwarf2/expr.c
+++ b/gdb/dwarf2/expr.c
@@ -1518,7 +1518,7 @@ dwarf_register::to_gdb_value (const frame_info_ptr &initial_frame,
 	 return a generic optimized out value instead, so that we show
 	 <optimized out> instead of <not saved>.  */
       value *temp = value::allocate (subobj_type);
-      retval->contents_copy (temp, 0, 0, 0, subobj_type->length ());
+      retval->contents_copy (temp, 0, 0, subobj_type->length ());
       retval = temp;
     }
 
--- a/gdb/dwarf2/expr.c
+++ b/gdb/dwarf2/expr.c
@@ -1378,6 +1378,17 @@ dwarf_register::read (const frame_info_ptr &initial_frame, gdb_byte *buf,
   if (frame == NULL)
     internal_error (_("invalid frame information"));
 
+  /* Check that the read will not try to access past the end of the register.
+     If it does, consider that some bits of the read are optimized out.  Since
+     we do not have a better granularity during the register read, report the
+     entire read as optimized out even if we could read a few bits.  */
+  if (total_bits_to_skip + bit_size > reg_bits)
+    {
+      *optimized = true;
+      *unavailable = false;
+      return;
+    }
+
   /* Populate or refresh the cache if needed.  */
   if (m_storage->m_read_cache.has_value ()
       && m_storage->m_read_cache->outdated (frame))
--- a/gdb/dwarf2/expr.c
+++ b/gdb/dwarf2/expr.c
@@ -541,13 +541,16 @@ public:
 
      Note that some location types can be read without a FRAME context.
 
-     If the location is optimized out or unavailable, the OPTIMIZED and
-     UNAVAILABLE outputs are set accordingly.  */
+     If the location is optimized out or unavailable (entirely or
+     partially), the affected ranges will be appended to OPTIMIZED and
+     UNAVAILABLE as necessary.  OPTIMIZED or UNAVAILABLE are left unchanged
+     if the location is not optimized or unavailable.  The ranges will
+     reference regions relative to BITS_TO_SKIP.  */
   virtual void read (const frame_info_ptr &frame, gdb_byte *buf,
 		     int buf_bit_offset, size_t bit_size,
 		     LONGEST bits_to_skip, size_t location_bit_limit,
-		     bool big_endian, bool *optimized,
-		     bool *unavailable) const = 0;
+		     bool big_endian, std::vector<range> &optimized_ranges,
+		     std::vector<range> &unavailable) const = 0;
 
   /* Write contents to a described location.
 
@@ -710,20 +713,20 @@ dwarf_location::deref (const frame_info_ptr &frame,
      from the type length, we need to zero-extend it.  */
   gdb::byte_vector read_buf (type->length (), 0);
   gdb_byte *buf_ptr = read_buf.data ();
-  bool optimized, unavailable;
+  std::vector<range> optimized, unavailable;
 
   if (big_endian)
     buf_ptr += type->length () - actual_size;
 
   this->read (frame, buf_ptr, 0, actual_size * HOST_CHAR_BIT,
-	      0, 0, big_endian, &optimized, &unavailable);
+	      0, 0, big_endian, optimized, unavailable);
 
-  if (optimized)
+  if (!optimized.empty ())
     throw_error (OPTIMIZED_OUT_ERROR,
 		 _("Can't do read-modify-write to "
 		   "update bitfield; containing word "
 		   "has been optimized out"));
-  if (unavailable)
+  if (!unavailable.empty ())
     throw_error (NOT_AVAILABLE_ERROR,
 		 _("Can't dereference "
 		   "update bitfield; containing word "
@@ -764,17 +767,19 @@ dwarf_location::write_to_gdb_value (const frame_info_ptr &frame,
 				    LONGEST bits_to_skip, size_t bit_size,
 				    size_t location_bit_limit)
 {
-  bool optimized, unavailable;
+  std::vector<range> optimized, unavailable;
   bool big_endian = type_byte_order (value->type ()) == BFD_ENDIAN_BIG;
 
   this->read (frame, value->contents_raw ().data (), value_bit_offset,
 	      bit_size, bits_to_skip, location_bit_limit,
-	      big_endian, &optimized, &unavailable);
+	      big_endian, optimized, unavailable);
 
-  if (optimized)
-    value->mark_bits_optimized_out (value_bit_offset, bit_size);
-  if (unavailable)
-    value->mark_bits_unavailable (value_bit_offset, bit_size);
+  for (auto &&optimized_range : optimized)
+    value->mark_bits_optimized_out
+      (value_bit_offset + optimized_range.offset, optimized_range.length);
+  for (auto &&unavailable_range : unavailable)
+    value->mark_bits_unavailable
+      (value_bit_offset + unavailable_range.offset, unavailable_range.length);
 }
 
 /* Value entry found on a DWARF expression evaluation stack.  */
@@ -931,11 +936,10 @@ public:
 
   void read (const frame_info_ptr &frame, gdb_byte *buf, int buf_bit_offset,
 	     size_t bit_size, LONGEST bits_to_skip, size_t location_bit_limit,
-	     bool big_endian, bool *optimized,
-	     bool *unavailable) const override
+	     bool big_endian, std::vector<range> &optimized,
+	     std::vector<range> &unavailable) const override
   {
-    *unavailable = false;
-    *optimized = true;
+    optimized.push_back ({0, bit_size});
   }
 
   void write (const frame_info_ptr &frame, const gdb_byte *buf,
@@ -1005,7 +1009,8 @@ public:
   void read (const frame_info_ptr &frame, gdb_byte *buf, int buf_bit_offset,
 	     size_t bit_size, LONGEST bits_to_skip,
 	     size_t location_bit_limit, bool big_endian,
-	     bool *optimized, bool *unavailable) const override;
+	     std::vector<range> &optimized,
+	     std::vector<range> &unavailable) const override;
 
   void write (const frame_info_ptr &frame, const gdb_byte *buf,
 	      int buf_bit_offset, size_t bit_size, LONGEST bits_to_skip,
@@ -1054,7 +1059,8 @@ void
 dwarf_memory::read (const frame_info_ptr &frame, gdb_byte *buf,
 		    int buf_bit_offset, size_t bit_size,
 		    LONGEST bits_to_skip, size_t location_bit_limit,
-		    bool big_endian, bool *optimized, bool *unavailable) const
+		    bool big_endian, std::vector<range> &optimized,
+		    std::vector<range> &unavailable) const
 {
   LONGEST total_bits_to_skip = bits_to_skip;
   CORE_ADDR start_address
@@ -1065,7 +1071,6 @@ dwarf_memory::read (const frame_info_ptr &frame, gdb_byte *buf,
     = gdbarch_segment_address_to_core_address (m_arch, m_address_space,
 					       start_address);
 
-  *optimized = false;
   total_bits_to_skip += m_bit_suboffset;
 
   if (total_bits_to_skip % HOST_CHAR_BIT == 0
@@ -1073,9 +1078,12 @@ dwarf_memory::read (const frame_info_ptr &frame, gdb_byte *buf,
       && buf_bit_offset % HOST_CHAR_BIT == 0)
     {
       /* Everything is byte-aligned, no buffer needed.  */
+      bool is_unavailable;
       read_from_memory (start_address,
 			buf + buf_bit_offset / HOST_CHAR_BIT,
-			bit_size / HOST_CHAR_BIT, m_stack, unavailable);
+			bit_size / HOST_CHAR_BIT, m_stack, &is_unavailable);
+      if (is_unavailable)
+	unavailable.push_back ({0, bit_size});
     }
   else
     {
@@ -1084,13 +1092,16 @@ dwarf_memory::read (const frame_info_ptr &frame, gdb_byte *buf,
 
       /* Can only read from memory on byte granularity so an
 	 additional buffer is required.  */
+      bool is_unavailable;
       read_from_memory (start_address, temp_buf.data (), this_size,
-			m_stack, unavailable);
+			m_stack, &is_unavailable);
 
-      if (!*unavailable)
+      if (!is_unavailable)
 	copy_bitwise (buf, buf_bit_offset, temp_buf.data (),
 		      total_bits_to_skip % HOST_CHAR_BIT,
 		      bit_size, big_endian);
+      else
+	unavailable.push_back ({0, bit_size});
     }
 }
 
@@ -1198,17 +1209,17 @@ dwarf_memory::deref (const frame_info_ptr &frame,
     }
   else
     {
-      bool optimized, unavailable;
+      std::vector<range> optimized, unavailable;
 
       this->read (frame, buf_ptr, 0, size_in_bits, 0, 0,
-		  big_endian, &optimized, &unavailable);
+		  big_endian, optimized, unavailable);
 
-      if (optimized)
+      if (!optimized.empty ())
 	throw_error (OPTIMIZED_OUT_ERROR,
 		     _("Can't do read-modify-write to "
 		     "update bitfield; containing word "
 		     "has been optimized out"));
-      if (unavailable)
+      if (!unavailable.empty ())
 	throw_error (NOT_AVAILABLE_ERROR,
 		     _("Can't dereference "
 		     "update bitfield; containing word "
@@ -1268,8 +1279,8 @@ public:
 
   void read (const frame_info_ptr &frame, gdb_byte *buf, int buf_bit_offset,
 	     size_t bit_size, LONGEST bits_to_skip, size_t location_bit_limit,
-	     bool big_endian, bool *optimized,
-	     bool *unavailable) const override;
+	     bool big_endian, std::vector<range> &optimized,
+	     std::vector<range> &unavailable) const override;
 
   void write (const frame_info_ptr &frame, const gdb_byte *buf,
 	      int buf_bit_offset, size_t bit_size, LONGEST bits_to_skip,
@@ -1347,8 +1358,8 @@ void
 dwarf_register::read (const frame_info_ptr &initial_frame, gdb_byte *buf,
 		      int buf_bit_offset, size_t bit_size,
 		      LONGEST bits_to_skip, size_t location_bit_limit,
-		      bool big_endian, bool *optimized,
-		      bool *unavailable) const
+		      bool big_endian, std::vector<range> &optimized,
+		      std::vector<range> &unavailable) const
 {
   frame_info_ptr frame = initial_frame;
   LONGEST total_bits_to_skip = bits_to_skip;
@@ -1378,17 +1389,6 @@ dwarf_register::read (const frame_info_ptr &initial_frame, gdb_byte *buf,
   if (frame == NULL)
     internal_error (_("invalid frame information"));
 
-  /* Check that the read will not try to access past the end of the register.
-     If it does, consider that some bits of the read are optimized out.  Since
-     we do not have a better granularity during the register read, report the
-     entire read as optimized out even if we could read a few bits.  */
-  if (total_bits_to_skip + bit_size > reg_bits)
-    {
-      *optimized = true;
-      *unavailable = false;
-      return;
-    }
-
   /* Populate or refresh the cache if needed.  */
   if (m_storage->m_read_cache.has_value ()
       && m_storage->m_read_cache->outdated (frame))
@@ -1397,13 +1397,39 @@ dwarf_register::read (const frame_info_ptr &initial_frame, gdb_byte *buf,
   if (!m_storage->m_read_cache.has_value ())
     m_storage->m_read_cache.emplace (arch, frame, reg);
 
-  *optimized = m_storage->m_read_cache->optimized_out ();
-  *unavailable = m_storage->m_read_cache->unavailable ();
+  if (m_storage->m_read_cache->optimized_out ())
+    optimized.push_back ({0, bit_size});
 
-  if (!*optimized && !*unavailable)
-    copy_bitwise (buf, buf_bit_offset,
-		  m_storage->m_read_cache->data ().data (),
-		  total_bits_to_skip, bit_size, big_endian);
+  if (m_storage->m_read_cache->unavailable ())
+    unavailable.push_back ({0, bit_size});
+
+  if (!m_storage->m_read_cache->optimized_out ()
+      && !m_storage->m_read_cache->unavailable ())
+    {
+      if (total_bits_to_skip > reg_bits)
+	{
+	  /* The read is completely past the register, mark everything as
+	     optimized out.  */
+	  optimized.push_back ({0, bit_size});
+	}
+      else
+	{
+	  if (total_bits_to_skip + bit_size > reg_bits)
+	    {
+	      /* The access goes past the end of the register.  Mark this part
+		 as optimized out, and read the rest.  */
+	      optimized.push_back
+		({static_cast<LONGEST> (reg_bits) - total_bits_to_skip,
+		  (static_cast<LONGEST> (bit_size) + total_bits_to_skip
+		   - reg_bits)});
+	      bit_size = reg_bits - total_bits_to_skip;
+	    }
+
+	  copy_bitwise (buf, buf_bit_offset,
+			m_storage->m_read_cache->data ().data (),
+			total_bits_to_skip, bit_size, big_endian);
+	}
+    }
 
 }
 
@@ -1504,14 +1530,14 @@ dwarf_register::is_optimized_out (const frame_info_ptr &frame, bool big_endian,
 				  LONGEST bits_to_skip, size_t bit_size,
 				  size_t location_bit_limit) const
 {
-  bool optimized, unavailable;
+  std::vector<range> optimized, unavailable;
   gdb::byte_vector temp_buf (bit_size);
 
   this->read (frame, temp_buf.data (), 0, bit_size,
 	      bits_to_skip, location_bit_limit,
-	      big_endian, &optimized, &unavailable);
+	      big_endian, optimized, unavailable);
 
-  return optimized;
+  return !optimized.empty ();
 }
 
 /* Implicit location description entry.  Describes a location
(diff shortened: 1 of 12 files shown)


lumachad:
This feels to me like it would belong in gdb.dwarf2/var-access.exp rather than its own test. Unless this isn't upstreamable, then I'd go with an isolated test and I'd mark it as non-upstreamable.

lancesix:
> This feels to me like it would belong in gdb.dwarf2/var-access.exp rather than its own test. Unless this isn't upstreamable, then I'd go with an isolated test and I'd mark it as non-upstreamable.

The fix is not upstreamable yet, but there is no reason why the test should not (would just need to justify this test as "we had a downstream regression, add an upstream test that would have caught it").  Let me go and look at extending var-access.exp.

gdb/testsuite/gdb.dwarf2/access-past-end-reg.exp
```diff
@@ -0,0 +1,109 @@
...
+require dwarf2_support
+
+# This test will store a 128 bits variable in a register too small to
+# contain the entire value.  Select a suitable register for which is less
+# than 128 bits.
+if { [is_amd64_regs_target] } {
+    # Use RAX, a 64 bits register.
+    set dwarf_regnum 0
+} else {
+    verbose "Skipping ${gdb_test_file_name} for current architecture."
```
aktemur, line 33: Did you mean to `return`, too, so that we exit the test? It also makes sense to have `unsupported` before returning.

lancesix:
Actually, it seems that upstream would behave differently

```diff
diff --git a/gdb/testsuite/gdb.dwarf2/var-access.exp b/gdb/testsuite/gdb.dwarf2/var-access.exp
index 6719784f230..c72443e3445 100644
--- a/gdb/testsuite/gdb.dwarf2/var-access.exp
+++ b/gdb/testsuite/gdb.dwarf2/var-access.exp
@@ -243,6 +243,16 @@ Dwarf::assemble $asm_file {
                        DW_OP_bit_piece 10 42
                    } SPECIAL_expr
                }
+               # Packed struct in a register, by with the struct size
+               # exceeding the register size.
+               DW_TAG_variable {
+                   DW_AT_name "invalid"
+                   DW_AT_type :$struct_st_label
+                   DW_AT_location {
+                       DW_OP_regx [lindex $dwarf_regnum 0]
+                       DW_OP_piece 12
+                   } SPECIAL_expr
+               }
            }
        }
     }
@@ -340,3 +350,11 @@ gdb_test_no_output "set var t3.z = -614" \
 switch $endian {big {set val "0x59, 0xa2"} little {set val "0x6a, 0x56"}}
 gdb_test "print/x *(char \[2\] *) (a + 5)" " = \\{$val\\}" \
     "verify t3.z through mem"
+
+# Try to read the struct packed into a register which is too small.
+set reg_size [get_valueof "" "sizeof (\$[lindex $regname 0])" 0]
+if {$reg_size < 12} {
+    gdb_test "print invalid" " = <unavailable>"
+} else {
+    unsupported "print invalid"
+}
```

With this, `invalid` variable is read starting from reg 0 ($rax), but missing bits are read from reg 1 ($rbx):

```
(gdb) p $rax = 0
$1 = 0
(gdb) p $rbx = ~0ull
quit
$2 = -1
(gdb) p/x invalid
$3 = {s = {a = 0x0, b = 0x0, c = 0x0, d = 0x0}, t = {u = 0x0, x = 0xffffffff, y = 0xffffffff, z = 0xffffffff}}
```

I think this kind of rings a bell that @simark explained GDB can try to read across regnums, but I really do no think our downstream evaluator accepts this.  So as it is, that one testcase can't be upstreamed.  Should it remain downstream for now then?

gdb/testsuite/gdb.dwarf2/access-past-end-reg.exp
lancesix: Yes, I did intend to return…

lancesix:
Fixed the missing return.  Since the test is not applicable upstream, do I keep it in a separate file, or do you prefer it is added to var-access.exp and we have a downstream divergence?

lumachad:
> Fixed the missing return. Since the test is not applicable upstream, do I keep it in a separate file, or do you prefer it is added to var-access.exp and we have a downstream divergence?

If it's non-upstreamable, I think it's OK to keep it separate so we simplify future contributions.

gdb/dwarf2/expr.c
```diff
@@ -1378,6 +1378,17 @@ dwarf_register::read (const frame_info_ptr &initial_frame, gdb_byte *buf,
   if (frame == NULL)
     internal_error (_("invalid frame information"));
 
+  /* Check that the read will not try to access past the end of the register.
+     If it does, consider that some bits of the read are optimized out.  Since
+     we do not have a better granulirity during the register read, report the
```
palves: granulirity => granularity

It doesn't seem like it'd be hard to have the granularity though.  The caller is dwarf_location::write_to_gdb_value, which has a struct value handy.  So we would pass the value pointer down to the read function, which, if the value pointer is non-null, would mark the right bits in the value as either optimized out our unavailable.  Other callers could still pass null value if they don't care for the granularity.

gdb/testsuite/gdb.dwarf2/access-past-end-reg.exp
```diff
@@ -0,0 +1,111 @@
...
+		DW_AT_low_pc $main_start DW_FORM_addr
+		DW_AT_high_pc $main_end DW_FORM_addr
+		DW_AT_type :$int_label
+	    } {
+		DW_TAG_variable {
+		    DW_AT_name foo
+		    DW_AT_type :$foo_t_label
+		    DW_AT_location {
+			DW_OP_regx $::dwarf_regnum
+			DW_OP_piece 16
```
palves, line 131: Does this trigger only when a composite wraps a register?  Or would it trigger if you don't do the DW_OP_piece too, just have a pure register location?  

I ask because the commit log just talks about register locations, but then the actual location is a composite (which wraps a register).

gdb/testsuite/gdb.dwarf2/access-past-end-reg.exp
```diff
@@ -0,0 +1,111 @@
...
+# This test exercises the case where a DWARF expression leads the dwarf
+# machinery to try to read past a given register's end.  This should be
+# handled gracefully by GDB.
+
+load_lib "dwarf.exp"
+
+require dwarf2_support
+
+# This test will store a 128 bits variable in a register too small to
+# contain the entire value.  Select a suitable register for which is less
```
palves: I think the "for" in "register for which" shouldn't be there.  Just "register which".

gdb/dwarf2/expr.c
lancesix: I'll look at changing this, and map the actual optimized out bits.

gdb/testsuite/gdb.dwarf2/access-past-end-reg.exp
lancesix: Funnily enough, it seems that if we only have `DW_OP_piece X`, then ROCgdb read from reg X then X + 1 and so on.  but with `DW_OP_piece 16`, this forces read of 16 bytes from reg X which is the problematic case.

I still need to figure out exactly why that is.  But this feels inconsistent.

lancesix: Just having DW_OP_regx gets up in `read_frame_register_value` which reads past the boundaries of registers.  So currently, we have an inconsistency.

lancesix: I have added a new patch to the series regarding this.  Please let me know if that seems reasonable to you.

gdb/dwarf2/expr.c
```diff
@@ -1976,18 +2009,26 @@ void
...
   unsigned int i;
 
   if (!m_completed)
     ill_formed_expression ();
 
   total_bits_to_skip += m_offset * HOST_CHAR_BIT + m_bit_suboffset;
 
+  /* The range_bit_offset (bit offset that we need to add to any
+     optimized / unavailable range) can start negative if the current
+     location has an offset.  However, as we skep over pieces not
```
palves: skep => skip

gdb/dwarf2/expr.c
```diff
@@ -1976,18 +2009,26 @@ void
...
+     covered by the read / the begining of the first coverede biece,
```
palves: Several typos: begining, coverede, biece.

gdb/testsuite/gdb.dwarf2/access-past-end-reg.exp
```diff
@@ -108,4 +109,10 @@ if { ![runto_main] } {
     return
 }
 
-gdb_test "p/x foo" "= <optimized out>"
+# The test assumes a register size of 64bits, ensure we have this.
```
palves: "64bits" => "64 bits"

gdb/testsuite/gdb.dwarf2/access-past-end-reg.exp
```diff
@@ -108,4 +109,10 @@ if { ![runto_main] } {
...
+set regsize [get_sizeof "\$$regname" ""]
+if { $regsize != 8 } {
+    unsupported "Invalid register size"
```
palves: The register to use is hard coded in the test, so if we get here, it's an actual testcase bug, not unsupported.  Right?  I mean, should fail loudly.

gdb/dwarf2/expr.c
```diff
@@ -1065,17 +1071,20 @@ dwarf_memory::read (const frame_info_ptr &frame, gdb_byte *buf,
...
       && buf_bit_offset % HOST_CHAR_BIT == 0)
     {
       /* Everything is byte-aligned, no buffer needed.  */
+      bool is_unavailable;
       read_from_memory (start_address,
 			buf + buf_bit_offset / HOST_CHAR_BIT,
-			bit_size / HOST_CHAR_BIT, m_stack, unavailable);
+			bit_size / HOST_CHAR_BIT, m_stack, &is_unavailable);
+      if (is_unavailable)
+	unavailable.push_back ({bits_to_skip, bit_size});
```
palves: dwarf_location::read says "ranges will reference regions relative to BITS_TO_SKIP".  I.e. offset 0 is the start of the read.  So ISTM this is pushing the wrong offset.  Should be `{0, bit_size}`.  Curious this didn't trigger any testsuite problem, though?

gdb/dwarf2/expr.c
```diff
@@ -1084,13 +1093,16 @@ dwarf_memory::read (const frame_info_ptr &frame, gdb_byte *buf,
...
-			m_stack, unavailable);
+			m_stack, &is_unavailable);
 
-      if (!*unavailable)
+      if (!is_unavailable)
 	copy_bitwise (buf, buf_bit_offset, temp_buf.data (),
 		      total_bits_to_skip % HOST_CHAR_BIT,
 		      bit_size, big_endian);
+      else
+	unavailable.push_back ({bits_to_skip, bit_size});
```
palves: Ditto re. offset.

gdb/dwarf2/expr.c
```diff
@@ -1608,12 +1635,17 @@ dwarf_implicit::read (const frame_info_ptr &frame, gdb_byte *buf,
 
   if (total_bits_to_skip >= implicit_bit_size)
     {
-      (*unavailable) = 1;
+      optimized.push_back ({0, bit_size});
```
palves, line 1634: This changes unavailable => optimized-out.  This is correct, IMHO (as previously discussed), but it's a semantic change that is not mentioned in the commit log.  Could also consider splitting it to a separate preparatory commit.

gdb/dwarf2/expr.c
```diff
@@ -1583,15 +1610,15 @@ void
...
+		      std::vector<range> &unavailable) const
 {
   ULONGEST implicit_bit_size = HOST_CHAR_BIT * m_size;
   LONGEST total_bits_to_skip = bits_to_skip;
   size_t read_bit_limit = location_bit_limit;
 
-  *optimized = false;
-  *unavailable = false;
+  optimized = {};
+  unavailable = {};
```
palves: The dwarf_location::read says "affected ranges will be appended", but this clears.  Other location kinds truly only append.  It seems inconsistent.  Either everyone appends or nobody does, I'd say?

gdb/dwarf2/expr.c
```diff
@@ -1397,13 +1398,39 @@ dwarf_register::read (const frame_info_ptr &initial_frame, gdb_byte *buf,
...
-		  total_bits_to_skip, bit_size, big_endian);
+  if (m_storage->m_read_cache->unavailable ())
+    unavailable.push_back ({0, bit_size});
+
+  if (!m_storage->m_read_cache->optimized_out ()
+      && !m_storage->m_read_cache->unavailable ())
+    {
+      if (total_bits_to_skip > reg_bits)
+	{
+	  /* The read is completly past the register, mark everything as
```
palves: completly => completely

gdb/dwarf2/expr.c
```diff
@@ -1504,14 +1531,14 @@ dwarf_register::is_optimized_out (const frame_info_ptr &frame, bool big_endian,
...
+  std::vector<range> optimized, unavailable;
   gdb::byte_vector temp_buf (bit_size);
 
   this->read (frame, temp_buf.data (), 0, bit_size,
 	      bits_to_skip, location_bit_limit,
-	      big_endian, &optimized, &unavailable);
+	      big_endian, optimized, unavailable);
 
-  return optimized;
+  return optimized.size () != 0;
```
palves: `!optimized.empty ()` is the idiomatic form.  Appears multiple times.

gdb/testsuite/gdb.dwarf2/access-past-end-reg.exp
```diff
@@ -116,3 +124,4 @@ if { $regsize != 8 } {
     return
 }
 gdb_test "p/x foo" "= {a = $::hex, b = $::hex, c = <optimized out>, d = <optimized out>}"
+gdb_test "p/x bar" "= {a = $::hex, b = $::hex, c = <optimized out>, d = <optimized out>}"
```
palves: I'd suggest either naming the variables "var_piece", "var_reg" or some thing like that, or adding explicit messages like "p/x foo with DW_OP_piece" etc.

gdb/findvar.c
```diff
@@ -604,31 +604,29 @@ read_frame_register_value (value *value)
...
+    reg_len = len;
 
-      offset += reg_len;
-      len -= reg_len;
-      reg_offset = 0;
-      regnum++;
+  if (len > reg_len)
+    {
+      /* Mark bits past the end of the register as optimized out.  */
+      value->mark_bytes_optimized_out (reg_len, len - reg_len);
```
palves: Is bit_offset always 0 here?

palves (commented):
On the commit logs:

patch 1 - gdb/dwarf: Fix dwarf_register::read overflow

   - "can end-up" => "can end up"
   - "follow-up path" => "follow-up patch"
    - "Hover dwarf_register::to_gdb_value" => "However"
    - "does not has to be true" => s/has/have/

patch 3 - gdb: read_frame_register_value do not read past current register

  - "value's type is larger that regitser's X size"
        s/regitser/register/
        s/larger that/larger than/

   Suggest reworking slightly:

    "if a value is marked as residing in register X but its type is larger than register X's size"

gdb/testsuite/gdb.dwarf2/access-past-end-reg.exp
lancesix: Indeed, that would be a test issue (if someone ports this test to an arch different than x86_64 incorrectly). Chaging this to an assert, unless you see another mean we could have to say that the testcase is wrong (rather than a GDB issue).

gdb/dwarf2/expr.c
lancesix: Indeed, this was an error.

I do no think this is covered by any test, but TBH there are quite a few untested bits here.  I have a TODO to increase the test coverage, but I should make that a ticket to properly document this.

gdb/dwarf2/expr.c
lancesix: Thanks for catching this.  This is a leftover of the first implementation which did always reset the vector, but I found it to be quite heavy so ended up simplifying it to just appending.

gdb/testsuite/gdb.dwarf2/access-past-end-reg.exp
palves: gdb_assert seems good.

gdb/findvar.c
lancesix: I rewrote the function to think in term of bits, so this should capture the case where bit_offset pushes the read past the register as well.

gdb/dwarf2/expr.c
palves: Do you intend to keep it as two commits in the end?  If so, the typo still exists in the first commit.

gdb/testsuite/gdb.dwarf2/access-past-end-reg.exp
```diff
@@ -25,9 +25,10 @@ require dwarf2_support
 # This test will store a 128 bits variable in a register too small to
 # contain the entire value.  Select a suitable register which is less
 # than 128 bits in size.
-if { [is_amd64_regs_target] } {
+if { [istarget "x86_64-*-*"] } {
```
palves: Does the is_amd64_regs_target => `[istarget "x86_64-*-*"]` change belong in the previous patch?

And why isn't is_amd64_regs_target the correct choice?

gdb/dwarf2/expr.c
palves: This one is still not addressed.  Should it be a range with offset 0, or is this bits_to_skip really correct here?

gdb/dwarf2/expr.c
```diff
@@ -1065,17 +1071,20 @@ dwarf_memory::read (const frame_info_ptr &frame, gdb_byte *buf,
     = gdbarch_segment_address_to_core_address (m_arch, m_address_space,
 					       start_address);
 
-  *optimized = false;
+  optimized = {};
```
palves: Shouldn't this clear also be removed, per the "implementations append" contract?

gdb/testsuite/gdb.dwarf2/access-past-end-reg.exp
```diff
@@ -108,4 +109,7 @@ if { ![runto_main] } {
     return
 }
 
-gdb_test "p/x foo" "= <optimized out>"
+# The test assumes a register size of 64 bits, ensure we have this.
+set regsize [get_sizeof "\$$regname" ""]
+gdb_assert {$regsize == 8} "ensure register size"
+gdb_test "p/x var_piece" "= {a = $::hex, b = $::hex, c = <optimized out>, d = <optimized out>}"
```
palves, line 179: So the testcase no longer tests the non-piece case?  ISTR it was in a patch n3, but the series only has 2 now.  I couldn't find a comment in the github notifications explaining why that was dropped, so I'm left a bit confused.   :-/

gdb/testsuite/gdb.dwarf2/access-past-end-reg.exp
lancesix: no, is_amd64_regs_target is actually good. I dropped that bit of the change.

gdb/testsuite/gdb.dwarf2/access-past-end-reg.exp
lancesix: There **is** a patch… just when I last pushed it was not pushed on my stg stack :( fixing this now.

gdb/dwarf2/expr.c
lancesix: Fixed in the right commit.  Thanks.

gdb/findvar.c
```diff
@@ -601,34 +601,35 @@ read_frame_register_value (value *value)
...
-      reg_offset = 0;
-      regnum++;
+  if (len > reg_len)
+    {
+      /* Mark bits past the end of the register as optimized out.  */
+      value->mark_bits_optimized_out (reg_len, len - reg_len);
     }
+
+  regval->contents_copy (value, offset, reg_offset,
+			 bit_offset % HOST_CHAR_BIT, reg_len / HOST_CHAR_BIT);
```
palves: Should we gdb_assert that bit_offset / reg_len are byte aligned?

gdb/findvar.c
```diff
@@ -601,34 +601,35 @@ read_frame_register_value (value *value)
...
+    = type_length_units (check_typedef (value->type ())) * HOST_CHAR_BIT;
 
-  /* Skip registers wholly inside of REG_OFFSET.  */
-  while (reg_offset >= register_size (gdbarch, regnum))
+  /* We try to read past the end of the register.  */
+  if (reg_bit_offset >= register_size (gdbarch, regnum) * HOST_CHAR_BIT)
     {
-      reg_offset -= register_size (gdbarch, regnum);
-      regnum++;
+      value->mark_bits_optimized_out (reg_bit_offset, len - reg_bit_offset);
```
palves: I think the offset is wrong here.

`reg_bit_offset` comes from `value->offset()` which is the offset within the register.  

But the value's contents go from 0..LEN.  So if the value maps to a register read that is completely past the end of the register, it means the value is wholy optimized out.  And should be expressed with:

```
      value->mark_bits_optimized_out (0, len);
```

Right?

This matches what the code is already doing (correctly) further below, like:

```
      /* Mark bits past the end of the register as optimized out.  */
      value->mark_bits_optimized_out (reg_len, len - reg_len);
```

Here, reg_len is relative to position 0 of the value.

I'm not sure it's possible to write a testcase for this using standard DWARF.  Probably doable with DW_OP_offset, though, like:

```
 DW_OP_reg0
 DW_OP_LLVM_offset_constu 8
```

(untested)

The testcase is explicitly a downstream test, so that should be OK, and can be converted to DW_OP_offset when DWARF6 finally lands.

palves (commented):
In patch 3, duplicated "if a value is marked":

"GDB has had a behavior where if a value is marked if a value is marked"

gdb/findvar.c
lancesix: Probably not.  We can craft a testcase where reg_len is not byte aligned, I have added a testcase for this.

Fixing this correctly requires adjusting the interface of `value::contents_copy`, which I do in the new patch 3 of the series.

gdb/findvar.c
lancesix: indeed, I have fixed this.

I have also added a testcase.

gdb/value.h
```diff
@@ -628,8 +628,19 @@ struct value
...
+     bits into DST value's contents, starting at DST_OFFSET bits.
+     If unavailable contents are being copied from this value, the
+     corresponding DST contents are marked unavailable accordingly.
+     DST must not be lazy.  If this value is lazy, it will be fetched
+     now.
+
+     It is assumed the contents of DST in the [DST_OFFSET,
+     DST_OFFSET+LENGTH) range are wholly available.  */
+  void contents_copy_bitsize (struct value *dst, LONGEST dst_offset,
+			      LONGEST src_bit_offset, LONGEST length);
```
palves: "bitsize" gave me pause.  It's not a typical word we use to distinguish from byte versions.

We already have:

```
  void contents_copy_raw (struct value *dst, LONGEST dst_offset,
			  LONGEST src_offset, LONGEST length);
```

and (what I really want to point at):

```
  /* A helper for value_from_component_bitsize that copies bits from
     this value to DEST.  */
  void contents_copy_raw_bitwise (struct value *dst, LONGEST dst_bit_offset,
				  LONGEST src_bit_offset, LONGEST bit_length);
```

So I'd think the new function should be contents_copy_bitwise to match contents_copy_raw_bitwise.
(And also copy_bitwise from utils.h.)

(I'd guess your _bitsize came from value_from_component_bitsize, and now I wonder whether that was a typo.)

gdb/value.h
```diff
@@ -628,8 +628,19 @@ struct value
      It is assumed the contents of DST in the [DST_OFFSET,
      DST_OFFSET+LENGTH) range are wholly available.  */
   void contents_copy (struct value *dst, LONGEST dst_offset,
-		      LONGEST src_offset, LONGEST src_bit_offset,
-		      LONGEST length);
+		      LONGEST src_offset, LONGEST length);
+
+  /* Copy LENGTH bits of this value's content starting at SRC_OFFSET
```
palves: SRC_OFFSET => SRC_BIT_OFFSET

gdb/value.h
```diff
@@ -628,8 +628,19 @@ struct value
...
+     bits into DST value's contents, starting at DST_OFFSET bits.
+     If unavailable contents are being copied from this value, the
+     corresponding DST contents are marked unavailable accordingly.
+     DST must not be lazy.  If this value is lazy, it will be fetched
+     now.
+
+     It is assumed the contents of DST in the [DST_OFFSET,
+     DST_OFFSET+LENGTH) range are wholly available.  */
+  void contents_copy_bitsize (struct value *dst, LONGEST dst_offset,
```
palves: Probably dst_offset should be dst_bit_offset to match src_bit_offset (and be clear).  Also length => bit_length.

gdb/testsuite/gdb.dwarf2/access-past-end-reg.exp
```diff
@@ -113,3 +177,14 @@ if { ![runto_main] } {
...
 gdb_test "p/x var_piece" "= {a = $::hex, b = $::hex, c = <optimized out>, d = <optimized out>}"
+
+# Read a variable larger than the register holding it, with a simple register
+# location description.
+gdb_test "p/x var_reg" "= {a = $::hex, b = $::hex, c = <optimized out>, d = <optimized out>}"
+
+# Read a value fully past the end of the register.
+gdb_test "p/x var_reg_past_the_end" "= <optimized out>"
+
+# Read a 1 byte value 2 bits away from the end of the register.  GDB needs
```
palves: "a 1 byte value" => "a 1-byte value"

gdb/testsuite/gdb.dwarf2/access-past-end-reg.exp
```diff
@@ -113,3 +177,14 @@ if { ![runto_main] } {
...
+# to be able less than one byte (2 bits), bits 3-7 are optimized out.
```
palves: Some missing word in 
  "to be able less than one byte"
I'd guess 
  "to be able to read less than one byte"

I think that should be "bits 2-7 are optimized out.".  I.e., off by one.   The two readable bits are 0 and 1.

gdb/value.c
```diff
@@ -1335,12 +1325,24 @@ value::contents_copy_raw_bitwise (struct value *dst, LONGEST dst_bit_offset,
...
+    fetch_lazy ();
+
+  contents_copy_raw (dst, dst_offset, src_offset, length);
+}
+
+/* See value.h.  */
+
+void
+value::contents_copy_bitsize (struct value *dst, LONGEST dst_offset,
+			      LONGEST src_offset, LONGEST length)
```
palves: I'd be good to make the parameter names here match the names in the header.

gdb/value.c
```diff
@@ -1335,12 +1325,24 @@ value::contents_copy_raw_bitwise (struct value *dst, LONGEST dst_bit_offset,
...
 {
   if (m_lazy)
     fetch_lazy ();
 
-  contents_copy_raw (dst, dst_offset, src_offset, src_bit_offset, length);
+  contents_copy_raw_bitwise (dst, dst_offset, src_offset, length);
```
palves: Here we can see the contents_copy_bitsize => contents_copy_raw_bitwise naming oddity I mention elsewhere.

palves (commented):
This is now mostly OK to me, thanks.  Just nits now, nothing functional.

In patch 3 - "gdb/value: refactor value::copy_contents methods"

Both subject and body say "value::copy_contents", but it should be value::contents_copy.

gdb/value.h
lancesix: Ah… I guess this all boils down to dyslexia… I'll fix this.

gdb/value.h
```diff
@@ -628,8 +628,19 @@ struct value
...
+
+  /* Copy BIT_LENGTH bits of this value's content starting at
+     SRC_BIT_OFFSET bits into DST value's contents, starting at
+     DST_BIT_OFFSET bits.  If unavailable contents are being copied
+     from this value, the corresponding DST contents are marked
+     unavailable accordingly.  DST must not be lazy.  If this value
+     is lazy, it will be fetched now.
+
+     It is assumed the contents of DST in the [DST_OFFSET,
+     DST_OFFSET+LENGTH) range are wholly available.  */
```
palves: The variable names in `[DST_OFFSET, DST_OFFSET+LENGTH)` need to be updated too.

palves (commented):
Almost there, but still one nit.  Sorry!

gdb/value.h
lancesix: Fixed.

lancesix:
> Almost there, but still one nit. Sorry!

Don't be! Thanks a lot for catching those.

palves (approved):

PR #145, opened 2026-05-29
recognize address space operator in Fortran

aktemur:


gdb/eval.c
```diff
@@ -2549,6 +2550,39 @@ unop_extract_operation::evaluate (struct type *expect_type,
...
+
+  const std::string &name = std::get<1> (m_storage);
+  std::optional<arch_addr_space_id> address_space_id
+    = gdbarch_name_to_address_space_id (exp->gdbarch, name.c_str ());
+
+  if (!address_space_id.has_value ())
+    error (_("Address space %s not recognised by the architecture"),
+	name.c_str ());
+
+  CORE_ADDR adddress
```
lumachad: We might want to fix this typo while at it. adddress with 3 d's.

gdb/testsuite/gdb.rocm/aspace-user-input.exp
```diff
@@ -49,6 +49,11 @@ gdb_test "print private_lane#0xffffffff" " = \\(void \\*\\) private_lane#0xfffff
 gdb_test_no_output "set var \$addr = 0xffffffff"
 gdb_test "print private_lane#\$addr" " = \\(void \\*\\) private_lane#0xffffffff"
 
+# Fortran accepts the '#' operator.
```
lumachad, line 55: Could we also add a test that passes garbage as the address space so we test the failure modes?

lumachad (approved):
LGTM with the comments fixed / justified.
Manual validation of non-gdb.dwarf2 and non-gdb.rocm tests required.

gdb/testsuite/gdb.rocm/aspace-user-input.exp
aktemur: > Could we also add a test that passes garbage as the address space so we test the failure modes?

I added a new small patch that extends the existing testsuite with
```tcl
gdb_test "print gibberish#0x123" \
    "Address space gibberish not recognised by the architecture"
```

I then repeated that for Fortran in the last patch.

gdb/eval.c
aktemur: > We might want to fix this typo while at it. adddress with 3 d's.

Done.

aktemur (commented):
I'll run the whole testsuite separately before merging.

aktemur:
> I'll run the whole testsuite separately before merging.

Done. No regressions found.

PR #147, opened 2026-05-29
Switch to a different cluster label

lumachad:
The existing cluster label linux-gfx942-1gpu-ossci-rocm is going away EOD May 29th. We need to switch to the new label until end of june, when we'll need to switch again to some other label.


geomin12 (approved):
looks right: https://github.com/ROCm/TheRock/blob/f354ce1dace37fa9df26d78306d0ecc2131455ba/build_tools/github_actions/amdgpu_family_matrix.py#L153

lumachad:
@lancesix FYI, last minute change so we can keep our CI running.

lancesix:
> @lancesix FYI, last minute change so we can keep our CI running.

sure, looks good to me.

PR #148, opened 2026-05-29
[amd-staging-rocgdb-16] Switch to a different cluster label

lumachad:
The existing cluster label linux-gfx942-1gpu-ossci-rocm is going away EOD May 29th. We need to switch to the new label until end of june, when we'll need to switch again to some other label.

(cherry picked from commit c0a9c84102fc5e263a5adcd480b5911c8627b988)


lumachad:
@lancesix Likewise for amd-staging-rocgdb-16. This should keep our CI running.

PR #149, opened 2026-06-01
Run gdb.rocm tests in parallel across GPUs

lumachad:
## Summary                                                                                                                                                                                                                                                                   

  Add a per-GPU lock-pool layer so `gdb.rocm` tests can run in parallel across
  GPUs (and, optionally, share a GPU) under `make check-parallel`, instead of
  serialising the entire rocm suite behind a single machine-wide lock.

  Today every `gdb.rocm` test contends for the same `gpu-parallel.lock`, so on
  an N-GPU node the suite still runs one rocm test at a time. This series
  introduces a generic shared/exclusive file-lock primitive in
  `lib/gdb-utils.exp`, replaces the single lockfile with one pool per visible
  GPU, lets each test opt into one of three concurrency modes, and ships a
  self-test plus documentation for the new knobs.

  ## What's in each commit

  1. **`gdb/testsuite/rocm: allow parallel rocm tests via per-GPU lock pools`**
     New shared/exclusive lock primitives in `lib/gdb-utils.exp` (single-pool
     + multi-pool variants, writer-priority barriers, row-major fill so a
     multi-GPU node spreads before stacking). `lib/rocm.exp` replaces the
     single `gpu-parallel.lock` with one pool per visible GPU, picks
     per-test concurrency from `::rocm_gpu_concurrency`
     (`parallel`/`serial`/`machine`), and pins `ROCR_VISIBLE_DEVICES` to the
     scheduled GPU for the body. Honours a pre-set `ROCR_VISIBLE_DEVICES`
     (pool 0 → first listed device), `ROCM_TEST_ARCH` filtering, and a
     `ROCGDB_ROCM_GPU_MAX_PARALLEL` cap. Falls back to single-GPU on
     heterogeneous fleets; downgrades parallel→serial on devices without
     multi-process debug support. Extends the `check-parallel` cleanup glob
     to sweep the new per-pool files.

  2. **`gdb/testsuite: self-test for the slot-pool shared/exclusive lock layer`**
     `gdb.testsuite/slot-pool-lock.exp` exercises the new primitives across
     real tclsh child processes: shared capacity, shared/exclusive mutex,
     multi-pool spread, writer priority, exclusive-behind-exclusive,
     machine lock, randomised start offset, and error-releases-lock.

  3. **`gdb/testsuite/rocm: opt rocm tests into shared (parallel) GPU lock`**
     Marks 60 `gdb.rocm` tests as `parallel`-safe (small single-kernel
     tests that only inspect per-inferior state). Wraps
     `simple-outside-debugger.exp` and `mi-attach.exp` in
     `with_rocm_gpu_lock` (they previously ran with no coordination).
     Tests that touch device-wide state, spawn hog kernels, or use
     multiple concurrent GPU inferiors stay on the default `serial` mode.

  4. **`gdb/testsuite: document gdb.rocm parallel testing controls`**
     New "ROCm GPU Parallel Testing" section in `gdb/testsuite/README`
     covering `::rocm_gpu_concurrency`, the failsafe, and the
     `ROCGDB_ROCM_PARALLEL_SLOTS` / `ROCGDB_ROCM_GPU_MAX_PARALLEL` /
     `ROCR_VISIBLE_DEVICES` knobs, with the combined concurrency
     ceiling.

  5. **`gdb/testsuite/rocm: split runtime-core.exp into parallel shards`**
     Splits `runtime-core.exp`'s six serial `do_test` invocations into
     one shard per `(fault, core_type, output_type)` combination
     sourcing a shared `runtime-core.exp.tcl`, mirroring
     `all-architectures-*.exp` / `aarch64-sme-core-*.exp`. Each shard
     takes a whole GPU (default `serial` mode); other GPUs stay free.

  ## Defaults

  - `ROCGDB_ROCM_PARALLEL_SLOTS=1` — one rocm test per GPU; raise it on
    hardware/stacks known to tolerate higher kfd/dbgapi concurrency.
  - `ROCGDB_ROCM_GPU_MAX_PARALLEL` unset — use every eligible GPU.
  - Per-test default mode is `serial` (no behavioural change for unmarked
    tests, except they no longer block other GPUs).

  ## Test plan

  - [ ] `make check-parallel TESTS="gdb.testsuite/slot-pool-lock.exp"` passes
        (self-test of the lock layer).
  - [ ] `make check-parallel TESTS="gdb.rocm/*.exp"` passes on a multi-GPU
        node and finishes meaningfully faster than the pre-series baseline.
  - [ ] Same on a single-GPU node (parallel mode collapses to serial; no
        regressions).
  - [ ] With `ROCR_VISIBLE_DEVICES=<subset>` pre-set: rocm tests only touch
        the listed devices, regardless of pool indexing.
  - [ ] With `ROCGDB_ROCM_GPU_MAX_PARALLEL=1`: rocm suite shares a single
        GPU end-to-end (equivalent to the pre-series behaviour).
  - [ ] Heterogeneous-arch fleet without `ROCM_TEST_ARCH`: suite pins to
        a single GPU and logs the fallback reason.
  - [ ] Non-parallel `make check` run: unchanged behaviour.

### DIFF
--- a/gdb/testsuite/Makefile.in
+++ b/gdb/testsuite/Makefile.in
@@ -258,7 +258,7 @@ check-single-racy:
 check-parallel:
 	-rm -f *core*
 	-rm -rf cache outputs temp
-	-rm -f gpu-parallel.lock
+	-rm -f gpu-parallel*.lock
 	$(MAKE) -k do-check-parallel; \
 	result=$$?; \
 	if test -d outputs; then \
--- a/gdb/testsuite/README
+++ b/gdb/testsuite/README
@@ -354,6 +354,167 @@ by switching to a different user on the same machine.  These users
 will have random files copied into their $HOME directories, so it is a
 good idea to setup new users just for this purpose.
 
+ROCm GPU Parallel Testing
+*************************
+
+The gdb.rocm/ test suite runs GPU-bound tests under a file-based lock
+managed by with_rocm_gpu_lock (see lib/rocm.exp).  The lock layer keeps
+the suite safe to run with GDB_PARALLEL on multi-GPU nodes by gating
+how many tests may share a GPU at once and which physical GPU each
+test sees.
+
+Per-test concurrency marker
+---------------------------
+
+Each .exp file selects one of three modes by setting
+::rocm_gpu_concurrency near its load_lib rocm.exp, before any call to
+with_rocm_gpu_lock:
+
+	set rocm_gpu_concurrency parallel
+	set rocm_gpu_concurrency serial   ;# the default
+	set rocm_gpu_concurrency machine
+
+The three modes are:
+
+  parallel
+	Take one slot on any one GPU.  Multiple parallel-marked tests
+	may run simultaneously, spread across GPUs first and stacked on
+	the same GPU only once every GPU has at least one tenant.  Use
+	this for tests that launch a small kernel (typically one
+	workgroup with a handful of workitems) and only inspect
+	per-inferior state.  This is the only mode that allows more
+	than one test to share a single GPU.
+
+  serial (default)
+	Take every slot of one GPU.  No other test runs on that GPU
+	until this one releases it, but tests on other GPUs are
+	unaffected.  Use this for tests that touch device-wide state
+	(such as device-wide attach/interrupt or precise-memory
+	toggles) or that spawn multiple concurrent GPU inferiors
+	within a single test.  This is the historical "one rocm test
+	at a time" behaviour, except now scoped to a single GPU
+	instead of the whole machine.
+
+  machine
+	Take every slot of every active GPU.  No other rocm test runs
+	anywhere on the node until this one releases.  Use this only
+	for tests whose inferiors spawn sibling HIP processes whose
+	runtime initialisation does not tolerate concurrent ROCm
+	activity on neighbouring GPUs.  Note: the inferior under
+	machine mode still sees a single device via
+	ROCR_VISIBLE_DEVICES (the same first-pool pin as serial mode);
+	the exclusivity is about keeping other rocm tests off the node,
+	not about handing the body all GPUs.
+
+There is a failsafe: if any local device lacks multi-process debug
+support (per hip_devices_support_debug_multi_process), parallel mode
+is downgraded to serial regardless of the per-test marker.  This
+covers older gfx generations that cannot have two debugged processes
+on the GPU at once.
+
+ROCGDB_ROCM_PARALLEL_SLOTS
+
+Number of concurrent parallel-marked tests allowed on a single GPU.
+Must be a positive integer.  Default 1.
+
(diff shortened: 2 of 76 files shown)


palves:
Did you find the reason why currently "make check-parallel" fails?  I'd prefer seeing the fix for that in a separate change first, before all the sharding stuff.

lumachad:
> Did you find the reason why currently "make check-parallel" fails? I'd prefer seeing the fix for that in a separate change first, before all the sharding stuff.

I don't think it fails. Not deterministically anyway. It does fail from time to time due to non-determinism in what the usage of the GPU's look like and what GPU ROCR picks up for the job. The ultimate problem with the current code is wall clock time and resource utilization.

We do have a couple tests (mentioned in the description) that lacked the rocm lock, so those were running in parallel. But I don't think they would've caused issues.

palves:
> I don't think it fails. Not deterministically anyway. It does fail from time to time due to non-determinism in what the usage of the GPU's look like and what GPU ROCR picks up for the job. The ultimate problem with the current code is wall clock time and resource utilization.
> 
> We do have a couple tests (mentioned in the description) that lacked the rocm lock, so those were running in parallel. 

Oh, that's not my experience.  We have internal tickets about this.  My suspition (I think recorded in the jiras) is that we were missing some with_rocm_gpu_lock calls in some testcases.

> But I don't think they would've caused issues.

On the contrary -- that's exactly the sort of thing that causes problems, for which with_rocm_gpu_lock was invented in the first place.  May be you're running tests on a GPU that allows debugging multiple processes at the same time.  But on a GPU that does not, missing a with_rocm_gpu_lock lock in one testcase breaks any other that runs at the same time.

So I think that bit should be hoisted out into its own commit, and preferably, it's own PR.

lumachad:
> > I don't think it fails. Not deterministically anyway. It does fail from time to time due to non-determinism in what the usage of the GPU's look like and what GPU ROCR picks up for the job. The ultimate problem with the current code is wall clock time and resource utilization.
> > We do have a couple tests (mentioned in the description) that lacked the rocm lock, so those were running in parallel.
> 
> Oh, that's not my experience. We have internal tickets about this. My suspition (I think recorded in the jiras) is that we were missing some with_rocm_gpu_lock calls in some testcases.
> 
> > But I don't think they would've caused issues.
> 
> On the contrary -- that's exactly the sort of thing that causes problems, for which with_rocm_gpu_lock was invented in the first place. May be you're running tests on a GPU that allows debugging multiple processes at the same time. But on a GPU that does not, missing a with_rocm_gpu_lock lock in one testcase breaks any other that runs at the same time.
> 
> So I think that bit should be hoisted out into its own commit, and preferably, it's own PR.

Done now in https://github.com/ROCm/ROCgdb/pull/153.

lumachad:
pre-commit issue being handled in https://github.com/ROCm/ROCgdb/pull/188

lumachad:
> pre-commit issue being handled in #188

Now addressed.

PR #153, opened 2026-06-02
Wrap mi-attach in with_rocm_gpu_lock

lumachad:
mi-attach.exp exercises the GPU but did not call with_rocm_gpu_lock,
so it ran without any coordination against the rest of the gdb.rocm
suite under GDB_PARALLEL.  An unlocked GPU-using test can clash with
a concurrently running rocm test that holds the GPU lock, producing
spurious failures driven by contention on kfd/dbgapi rather than by
the code under test.

Wrap the GPU-using portion in with_rocm_gpu_lock so it participates
in the existing per-GPU serialisation like the other rocm tests.

### DIFF
--- a/gdb/testsuite/gdb.rocm/mi-attach.exp
+++ b/gdb/testsuite/gdb.rocm/mi-attach.exp
@@ -25,9 +25,11 @@ if {[build_executable "failed to prepare" $testfile $srcfile {debug hip}]} {
     return
 }
 
-set spawn_id [spawn_wait_for_attach $::binfile]
-set prog_pid [spawn_id_get_pid $spawn_id]
+with_rocm_gpu_lock {
+    set spawn_id [spawn_wait_for_attach $::binfile]
+    set prog_pid [spawn_id_get_pid $spawn_id]
 
-mi_clean_restart
+    mi_clean_restart
 
-mi_gdb_test "-target-attach $prog_pid" ".*\\^done.*" "attach \$PROG_PID"
+    mi_gdb_test "-target-attach $prog_pid" ".*\\^done.*" "attach \$PROG_PID"
+}


amd-shahab (approved):

lancesix (approved):

palves (changes requested):
Oh, hmm.  In the other discussion, I hadn't noticed that simple-outside-debugger.exp was one of the testscases affected.  

Currently, with_rocm_gpu_lock is meant to prevent _debugging_ more than one GPU process at a time, so that's why it didn't use with_rocm_gpu_lock.  So it was on purpose (I wrote the test not that long ago, I remember.)

Note that is how we document the mechanism:

```
 # The lock file used to ensure that only one GDB has access to the GPU
 # at a time.
 set gpu_lock_filename gpu-parallel.lock
```

However, thinking further, if we have one of the hog testcases running at the same time, it'll be helpful to avoid running anything else on the GPU at the same time, even if we don't debug it.  So I'll adjust my mental model a bit.  But that comment should be updated, at least.

lumachad:
> Oh, hmm. In the other discussion, I hadn't noticed that simple-outside-debugger.exp was one of the testscases affected.
> 
> Currently, with_rocm_gpu_lock is meant to prevent _debugging_ more than one GPU process at a time, so that's why it didn't use with_rocm_gpu_lock. So it was on purpose (I wrote the test not that long ago, I remember.)
> 
> Note that is how we document the mechanism:
> 
> ```
>  # The lock file used to ensure that only one GDB has access to the GPU
>  # at a time.
>  set gpu_lock_filename gpu-parallel.lock
> ```
> 
> However, thinking further, if we have one of the hog testcases running at the same time, it'll be helpful to avoid running anything else on the GPU at the same time, even if we don't debug it. So I'll adjust my mental model a bit. But that comment should be updated, at least.

@palves Would you like the comment you quoted to be updated on this PR?

lumachad:
> > Oh, hmm. In the other discussion, I hadn't noticed that simple-outside-debugger.exp was one of the testscases affected.
> > Currently, with_rocm_gpu_lock is meant to prevent _debugging_ more than one GPU process at a time, so that's why it didn't use with_rocm_gpu_lock. So it was on purpose (I wrote the test not that long ago, I remember.)
> > Note that is how we document the mechanism:
> > ```
> >  # The lock file used to ensure that only one GDB has access to the GPU
> >  # at a time.
> >  set gpu_lock_filename gpu-parallel.lock
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > However, thinking further, if we have one of the hog testcases running at the same time, it'll be helpful to avoid running anything else on the GPU at the same time, even if we don't debug it. So I'll adjust my mental model a bit. But that comment should be updated, at least.
> 
> @palves Would you like the comment you quoted to be updated on this PR?

Dropped the simple-outside-debugger.exp locking based on Pedro's feedback.

lancesix (approved):
LGTM, thanks.

amd-shahab (approved):

palves (approved):

PR #156, opened 2026-06-04
gdb.rocm/names.exp: Make test program exit once all kernels are running

palves:
Currently, if you run the 'names' program outside GDB, like so:

 GPU_MAX_HW_QUEUES=32 ./testsuite/outputs/gdb.rocm/names/names  all

... it hangs "forever", until it aborts due to the gdb_watchdog
timeout.

Make the kernels exit after every kernel is running, instead, using a
pair of globals and atomics.

Make gdb.rocm/names.exp test the manual recipe above itself, too.

## Motivation

Helps with diagnosing non-debugger problems.

## Test Plan

Run the testcase, try manual recipe in commit log.

## Test Result

Positive!

### DIFF
--- a/gdb/testsuite/gdb.rocm/names.cpp
+++ b/gdb/testsuite/gdb.rocm/names.cpp
@@ -35,21 +35,14 @@
 	}                                                                    \
     } while (0)
 
-/* All kernels call this.  */
-
-__device__ static void
-wait_forever ()
-{
-  while (1)
-    __builtin_amdgcn_s_sleep (1);
-}
+__device__ static void wait_all_kernels ();
 
 /* A kernel with a name longer than GDB's longest possible name.  */
 
 __global__ void
 long_kernel_name_abcdefghijklmnopqrstuvwxyz ()
 {
-  wait_forever ();
+  wait_all_kernels ();
 }
 
 /* A kernel with a name longer than GDB's longest possible name.  */
@@ -57,7 +50,7 @@ long_kernel_name_abcdefghijklmnopqrstuvwxyz ()
 __global__ void
 long_kernel_name_0123456789 ()
 {
-  wait_forever ();
+  wait_all_kernels ();
 }
 
 /* A kernel with a normal, short name.  */
@@ -65,7 +58,7 @@ long_kernel_name_0123456789 ()
 __global__ void
 kern ()
 {
-  wait_forever ();
+  wait_all_kernels ();
 }
 
 /* Same, but with some parameters.  */
@@ -73,7 +66,7 @@ kern ()
 __global__ void
 kern (int)
 {
-  wait_forever ();
+  wait_all_kernels ();
 }
 
 /* A kernel with extern "C" linkage, which doesn't include an argument
@@ -82,7 +75,7 @@ kern (int)
 extern "C" __global__ void
 c_kern ()
 {
-  wait_forever ();
+  wait_all_kernels ();
 }
 
 /* A kernel in a namespace.  */
@@ -92,7 +85,7 @@ namespace NS {
 __global__ void
 kern_ns ()
 {
-  wait_forever ();
+  wait_all_kernels ();
 }
 
 }
@@ -104,7 +97,7 @@ namespace {
 __global__ void
 anonymous ()
 {
-  wait_forever ();
+  wait_all_kernels ();
 }
 
 }
@@ -120,14 +113,14 @@ template<E val>
 __global__ void
 templ_non_type ()
 {
-  wait_forever ();
+  wait_all_kernels ();
 }
 
 template<typename... Args>
 __global__ void
 templ_type (Args... args)
 {
-  wait_forever ();
+  wait_all_kernels ();
 }
 
 /* List of kernel names and a callback that launches the named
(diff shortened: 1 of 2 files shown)


gdb/testsuite/gdb.rocm/names.cpp
```diff
@@ -218,13 +239,20 @@ main (int argc, char **argv)
       return EXIT_FAILURE;
     }
 
+  int h_number_kernels = 0;
+  for (auto &k : kernels)
```
lumachad, line 245: Nit: Braces around the multi-line block?

for (...)
{
...
}

lancesix (approved):
LGTM, thanks for looking into this.

gdb/testsuite/gdb.rocm/names.cpp
```diff
@@ -205,6 +198,34 @@ static struct
...
+}
+
+/* True if we want to run the kernel named KERNEL.  ARG is the
+   argument passed to the program.  Either a kernel name, or
+   "all".  */
+
+static bool
+should_run_kernel (const char *arg, const char *kernel)
+{
+  return (strcmp (arg, "all") == 0 || strcmp (arg, kernel) == 0);
```
lumachad: Nit: Parens might not be needed around the expression.

return strcmp (arg, "all") == 0 || strcmp (arg, kernel) == 0;?

gdb/testsuite/gdb.rocm/names.cpp
```diff
@@ -205,6 +198,34 @@ static struct
...
+__device__ static int kernels_seen = 0;
+__device__ static int number_kernels = 0;
+
+__device__ static void
+wait_all_kernels ()
+{
+  atomicAdd (&kernels_seen, 1);
+  while (atomicAdd (&kernels_seen, 0) < number_kernels)
+    {
+      /* Sleep a bit to avoid hogging too much.  */
```
lumachad: Clarification: /* Sleep a bit to avoid hogging the GPU.  */

gdb/testsuite/gdb.rocm/names.cpp
```diff
@@ -205,6 +198,34 @@ static struct
   },
 };
 
+/* All kernels call wait_all_kernels, which waits until every kernel
+   has reached that point, so that they are all running concurrently
+   before any of them exits.  */
+
```
lumachad: Nit: Move the comment above the function?

lumachad (approved):
Just nits. Thanks.

One thing that crossed my mind. Should we be careful about the number of kernels being expected to reach wait_all_kernels? What if a device can't spawn all of them at the same time? I'm not sure if that's a valid concern, but I thought I'd mention it.

lancesix:
> One thing that crossed my mind. Should we be careful about the number of kernels being expected to reach wait_all_kernels? What if a device can't spawn all of them at the same time? I'm not sure if that's a valid concern, but I thought I'd mention it.

In such case (for example if `GPU_MAX_HW_QUEUES` is not big enough), we do expect the hang.  But having this change makes it possible for us to prove that any hang is not a GDB bug but an issue somewhere else in the stack.

amd-shahab:
> [...] if you run the 'names' program outside GDB [...]
> [...] it hangs "forever", until it aborts due to the gdb_watchdog timeout.

I don't understand how there can be a `gdb_watchdog` when the program is run standalone? Are the test binaries linked against some wrappers?

UPDATE: seems there's a watch dog in the final binary
```
$ ./gdb/testsuite/outputs/gdb.rocm/names/names all
  gdb_watchdog: timeout expired - aborting test
  [1]    15235 IOT instruction (core dumped)  ./gdb/testsuite/outputs/gdb.rocm/names/names all
```

lancesix:
@amd-shahab, we have a helper header (https://github.com/ROCm/ROCgdb/blob/amd-staging/gdb/testsuite/lib/gdb_watchdog.h) which provides a helper to call `alarm (timeout)`.

amd-shahab:
> @amd-shahab, we have a helper header (https://github.com/ROCm/ROCgdb/blob/amd-staging/gdb/testsuite/lib/gdb_watchdog.h) which provides a helper to call `alarm (timeout)`.

right! there's a direct call to it in the program code

```
int
main (...)
{
  /* So that the GPU threads don't spin forever.  */
  gdb_watchdog (30);
  ...
}
```

amd-shahab (approved):

gdb/testsuite/gdb.rocm/names.cpp
palves: I did this, but IMHO, that's just added noise in this case.  The for block is a single statement, and there's no if/else ambiguity.

gdb/testsuite/gdb.rocm/names.cpp
palves: Done.

gdb/testsuite/gdb.rocm/names.cpp
palves: Done.

gdb/testsuite/gdb.rocm/names.cpp
palves: That was intended as a leading comment to the block of variables + function, but I can see how it didn't come out great.  I did that now, and also added intro comments to the variables.

palves:
I realized that I could also make the testcase run the program outside the debugger itself.  I did that change.

Please take another look.

palves:
Found myself wasting time debating what to do with the assigned field in this case, so I just ended up assigning to everyone who had originally approved.  :-D

lumachad:
Approved from my end.

amd-shahab (approved):

lancesix (approved):
LGTM, thanks.

PR #161, opened 2026-06-05
gdb.rocm/update-thread-list.exp: Avoid "update" in exe name

palves:
Running gdb.rocm/update-thread-list.exp on Windows 11 fails because GDB cannot spawn the inferior:

 (gdb) run
 Starting program: .../gdb.rocm/update-thread-list/update-thread-list
 Error creating process .../update-thread-list (error 740): The requested operation requires elevation.

Error 740 is ERROR_ELEVATION_REQUIRED.

Windows has an "installer detection" heuristic that refuses to launch executables whose filename contains keywords like "update", "setup", "install" or "patch" without elevation (admin rights), unless the PE embeds a manifest declaring requestedExecutionLevel="asInvoker".

I found some older Microsoft documentation claiming that the heuristic only applies to 32-bit binaries, but what I observe is that:

 - It still triggers with 64-bit PEs on current Windows.

 - Matches "update" as a substring anywhere in the filename (not just as a prefix).

 - Is not drive-dependent.  I thought moving the process to a dev drive might suppress it, but it does not.

Fix it by passing an explicit executable name to prepare_for_testing, so the on-disk binary is "test" instead of "update-thread-list".

Change-Id: I974616de4c08d702671a8316d822360398dfdf85

## Motivation

One fewer testcase failing on Windows.

## Technical Details

See commit log.

## Test Plan

Tested on Linux & Windows.

## Test Result

Happiness.

## Submission Checklist

- [x] Onions.

### DIFF
--- a/gdb/testsuite/gdb.rocm/update-thread-list.exp
+++ b/gdb/testsuite/gdb.rocm/update-thread-list.exp
@@ -25,7 +25,14 @@ require allow_hipcc_tests
 
 standard_testfile .cpp
 
-if {[prepare_for_testing "failed to prepare ${testfile}" $testfile $srcfile {debug hip}]} {
+# Don't use $testfile as the executable name, because Windows'
+# installer detection heuristic refuses to launch executables whose
+# name contains "update" (and a few other keywords like "setup",
+# "install", "patch") without elevation, failing with
+# ERROR_ELEVATION_REQUIRED (740).
+set executable "test"
+
+if {[prepare_for_testing "failed to prepare ${testfile}" $executable $srcfile {debug hip}]} {
     return -1
 }
 


lancesix (approved):
LGTM, thanks.

aktemur (approved):

gdb/testsuite/gdb.rocm/update-thread-list.exp
```diff
@@ -25,7 +25,14 @@ require allow_hipcc_tests
...
 
-if {[prepare_for_testing "failed to prepare ${testfile}" $testfile $srcfile {debug hip}]} {
+# Don't use $testfile as the executable name, because Windows'
+# installer detection heuristic refuses to launch executables whose
+# name contains "update" (and a few other keywords like "setup",
+# "install", "patch") without elevation, failing with
+# ERROR_ELEVATION_REQUIRED (740).
+set executable "test"
+
+if {[prepare_for_testing "failed to prepare ${testfile}" $executable $srcfile {debug hip}]} {
```
lumachad, line 35: Sounds like this should be a pre-commit hook that prevents introduction of anything named in a way that doesn't work.

lumachad (approved):
LGTM.

lancesix (approved):

gdb/testsuite/gdb.rocm/update-thread-list.exp
palves: If we were to have a check, I think it would be better in the testsuite itself, like in gdb_compile or somewhere that calls it.  Seems hard to check this from a pre-commit.
I suspect this will be quite rare, though and it isn't worth the effort.  There's only one other instance in the testsuite (gdb.base/execl-update-breakpoints.exp) and that one doesn't work on Windows (no fork/exec on native Windows; and on Cygwin, no support for catching such events).

lumachad: > If we were to have a check, I think it would be better in the testsuite itself, like in gdb_compile or somewhere that calls it. Seems hard to check this from a pre-commit. I suspect this will be quite rare, though and it isn't worth the effort. There's only one other instance in the testsuite (gdb.base/execl-update-breakpoints.exp) and that one doesn't work on Windows (no fork/exec on native Windows; and on Cygwin, no support for catching such events).

Ok. But we do have gdb.base/patch.[c|exp]. And according to your comment on the test, it would cause issues.

lumachad: Maybe also gdb.rocm/info-dispatches.[cpp|exp]? Or is it smart enough to dodge that?

palves: Looks like the information floating around that says that "patch" is also blacklisted is wrong.  gdb.rocm/info-dispatches.exp passes cleanly on Windows.  I wondered whether it's because it is looking for a whole word (like, dispatch good, patch bad), but no:

```
$ ./outputs/gdb.rocm/info-dispatches/info-dispatches
info: launch 'single_wave_kernel1' and 'single_wave_kernel2'  ## ran, so it's good.
$ cp ./outputs/gdb.rocm/info-dispatches/info-dispatches ./patch
$ ./patch
info: launch 'single_wave_kernel1' and 'single_wave_kernel2'  ## ran, so it's good.
$ mv patch update
$ ./update
-bash: ./update: Permission denied
$ mv update setup
$ ./setup
-bash: ./setup: Permission denied
$ mv setup setup1
$ ./setup1
-bash: ./setup1: Permission denied
$ mv setup1 1setup
$ ./1setup
-bash: ./1setup: Permission denied
$ mv 1setup dissetup
$ ./dissetup
-bash: ./dissetup: Permission denied
```

gdb.base/patch.exp doesn't work with our msvc-abi clang, but testing with gcc, I see that it runs fine.

I'll send a patch to update the comments to not mention "patch".  Thanks.

palves: > I'll send a patch to update the comments to not mention "patch".

https://github.com/ROCm/ROCgdb/pull/167

PR #162, opened 2026-06-08
Update TheRock dependencies and container images

lumachad:
TheRock commit:
  a9fd5ab0f531 -> 25083b016c47

Container images:
  therock_build_manylinux_x86_64:
    sha256:484925405916... -> sha256:f9fa413c3484...

PR #163, opened 2026-06-08
.github/workflows: use installed test_rocgdb.py

lumachad:
TheRock now installs test_rocgdb.py to tests/rocgdb (see ROCm/TheRock#4960), so invoke the installed copy from OUTPUT_ARTIFACTS_DIR instead of the TheRock-hosted script.

Also drop the unused THEROCK_BIN_DIR env var.

### DIFF
--- a/.github/workflows/therock-test-packages.yml
+++ b/.github/workflows/therock-test-packages.yml
@@ -29,7 +29,6 @@ env:
   THEROCK_COMMIT_REF: 25083b016c475a65cbea46ae7326104680bff1d5 # 2026-06-08
   VENV_DIR: ${{ github.workspace }}/.venv
   OUTPUT_ARTIFACTS_DIR: "./build"
-  THEROCK_BIN_DIR: "./build/bin"
 
 jobs:
   rocgdb-tests:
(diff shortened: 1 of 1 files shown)


.github/workflows/therock-test-packages.yml
```diff
@@ -68,4 +67,4 @@ jobs:
 
       - name: Run rocgdb tests
         run: |
-          python build_tools/github_actions/test_executable_scripts/test_rocgdb.py
+          python ${{ env.OUTPUT_ARTIFACTS_DIR }}/tests/rocgdb/test_rocgdb.py
```
lancesix, line 70: EDIT: seems I did not hit send on that comment yesterday.  Now that the base is good, things pass, so this is probably not needed anymore.

---

This causes failures.   Should this select the file based on what is available?

```
if [[ -e ${{ env.OUTPUT_ARTIFACTS_DIR }}/tests/rocgdb/test_rocgdb.py ]]
then
  python ${{ env.OUTPUT_ARTIFACTS_DIR }}/tests/rocgdb/test_rocgdb.py
else
  python build_tools/github_actions/test_executable_scripts/test_rocgdb.py
fi
```

This, or we need to bump the rock version used in the same batch.

lumachad:
Needs https://github.com/ROCm/ROCgdb/pull/162 first.

lancesix (approved):
LGTM, thanks.

PR #164, opened 2026-06-08
Test: Relax load corefile test in runtime-core.exp

amd-bfilipov:
We have rewritten coredump logic in runtime (see https://github.com/ROCm/rocm-systems/pull/6811). As the consequence of that, we can't rely on AMDGPU thread being selected when the dump is loaded (we have it's data but it's not printed right away on load). This is not a problem, if there is something wrong with dumps, we'll catch it in the other tests.


lancesix (changes requested):
I do not think this the right approach.

The core dump must retain the per wave state.  I did send you a patch  in dbgapi which would fix this issue.

lancesix:
FYI, I have this patch in dbgapi (still pending some discussion) which should address your issue: https://github.com/ROCm/rocm-systems/pull/6947

PR #165, opened 2026-06-09
Backport Windows UTF-8 fixes

palves:
## Motivation

Fix problems like:

```
 Thread 7 "kernel" hit Hardware watchpoint 3: array
 Old value = \<error reading variable: Converting character sets: Invalid argument.\>
 New value = \<error reading variable: Converting character sets: Invalid argument.\>
```

... when the Windows codepage is UTF-8 (65001).

## Technical Details

Pick fixes from upstream.  See commit logs for details.

The main fix is in commit "gdb: Support UTF-8 output on MS-Windows terminal".
The other commits are dependencies of that one.

(And yes, one of the commits does not have a proper subject, it was already like that upstream.)

## Test Plan

Run gdb.rocm/ tests on Windows.

## Test Result

Codepage problems went away.  Results look as good as with a non-UTF-8 codepage.

### DIFF
--- a/gdb/NEWS
+++ b/gdb/NEWS
@@ -35,6 +35,15 @@
 
 * Support for the binary file format dbx has been removed.
 
+* When running on the Windows Terminal console, GDB now supports
+  true-color 24-bit colors.
+
+* When running on the Windows Terminal console whose output codepage
+  is 65001, GDB now supports emoji styling and display of UTF-8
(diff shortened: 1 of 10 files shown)


palves:
Forgot to mention: the patches applied cleanly, no adjustment was necessary.

lancesix (approved):
LGTM, thanks.

PR #166, opened 2026-06-10
gdb/testsuite/gdb.rocm: extract reusable multi-inferior driver helpers

spatrang:
## Why this PR

This is preparatory refactoring split out of #131 at reviewer request.
While reviewing #131 (which adds a new multi-inferior stress test), it
was noted that the new test shares most of its driver logic with the
existing `gdb.rocm/multi-inferior-gpu.exp`. Rather than duplicate that
logic, the common parts are extracted here into shared helpers first, so
that #131 can reuse them and its diff reduces to just what is genuinely
new.

The tests are kept separate (only the driver logic is shared).

## Dependent PR

- #131 ("gdb/testsuite/gdb.rocm: add multi-inferior stress test")
  depends on this PR and will be rebased on top of it once this merges.

## Summary

Extract the shared non-stop multi-inferior driver logic out of
`gdb.rocm/multi-inferior-gpu.exp` into two reusable helper procs in
`gdb/testsuite/lib/rocm.exp`, and convert the existing test to use them.

## Helpers added (`lib/rocm.exp`)

- `rocm_multi_inferior_run_to_kernels {args_list expected}` — load the
  program, enable non-stop with `detach-on-fork off` / `follow-fork
  parent`, plant the breakpoints, run the parent to its pre-fork
  breakpoint, resume in the background, and collect one kernel
  breakpoint stop per child inferior. Returns the list of stopped GPU
  thread ids. The child count can be passed explicitly or discovered at
  runtime.
- `rocm_multi_inferior_drain {threads}` — continue each stopped GPU
  inferior to a clean exit, wait for the parent to reach its
  post-`waitpid` breakpoint, and run the parent to completion.

## Behavior note

The extracted driver is intentionally stricter than the inlined
original: it deduplicates GPU stops by inferior, uses literal-matched
regexes, and fails loudly on timeout or a non-zero child exit instead of
hanging. Coverage of the converted test is otherwise unchanged.

## Files changed

- `gdb/testsuite/lib/rocm.exp` — add the two helpers.
- `gdb/testsuite/gdb.rocm/multi-inferior-gpu.exp` — convert to use them.

### DIFF
--- a/gdb/testsuite/gdb.rocm/multi-inferior-gpu.cpp
+++ b/gdb/testsuite/gdb.rocm/multi-inferior-gpu.cpp
@@ -15,12 +15,30 @@
    You should have received a copy of the GNU General Public License
    along with this program.  If not, see <http://www.gnu.org/licenses/>.  */
 
+/* Shared driver program for the gdb.rocm multi-inferior tests.
+
+   The parent process forks N children; each child re-execs itself so
+   that its GPU runtime is initialized in a clean address space, then
+   launches its own GPU kernel.  Re-exec'ing makes each child a separate
+   process from the operating system's point of view, which is what the
+   multi-inferior tests debug concurrently.
+
+   N comes from argv[1] when given; otherwise it defaults to the number
+   of GPU devices found at runtime (one child per device).  The
+   companion helpers in multi-inferior.exp.tcl plant breakpoints on the
+   pre-fork and post-waitpid source markers and on the kernel.  */
+
 #include <sys/types.h>
 #include <sys/wait.h>
-#include <unistd.h>
+
 #include <hip/hip_runtime.h>
+
+#include <unistd.h>
+#include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
+#include <string>
 
 #include "rocm-test-utils.h"
 
@@ -30,77 +48,99 @@ kern ()
   asm ("s_sleep 1");
 }
 
-/* Spawn one child process per detected GPU.  */
+static int
+child (int argc, char **argv)
+{
+  if (argc < 4)
+    {
+      fprintf (stderr, "%s: expected: child <idx> <num_devices>\n",
+	       argv[0]);
+      return -1;
+    }
+
+  int idx = atoi (argv[2]);
+  int num_devices = atoi (argv[3]);
+  if (num_devices <= 0)
+    {
+      fprintf (stderr, "child %d: invalid num_devices %d\n", idx,
+	       num_devices);
+      return -1;
+    }
+
+  CHECK (hipSetDevice (idx % num_devices));
+  kern<<<1, 1>>> ();
+  CHECK (hipDeviceSynchronize ());
+  return 0;
+}
+
+/* Spawn NUM_CHILDREN child processes.  When NUM_CHILDREN is not
+   positive, spawn one child per detected GPU device.  */
 
 static int
-parent (int argc, char **argv)
+parent (const char *argv0, int num_children)
 {
-  /* Identify how many GPUs we have, and spawn one child for each.  */
   int num_devices;
   CHECK (hipGetDeviceCount (&num_devices));
+  if (num_devices <= 0)
+    {
+      fprintf (stderr, "no GPU devices available\n");
+      return -1;
+    }
+
+  if (num_children <= 0)
+    num_children = num_devices;
 
   /* Break here.  */
 
-  for (int i = 0; i < num_devices; i++)
+  for (int i = 0; i < num_children; i++)
     {
-      char n[32] = {};
-      snprintf (n, sizeof (n), "%d", i);
+      std::string idx_str = std::to_string (i);
+      std::string ndev_str = std::to_string (num_devices);
+
       pid_t pid = fork ();
       if (pid == -1)
 	{
-	  perror ("Fork failed");
+	  perror ("fork");
 	  return -1;
 	}
 
       if (pid == 0)
 	{
-	  /* Exec to force the child to re-initialize the ROCm runtime.  */
-	  if (execl (argv[0], argv[0], n, nullptr) == -1)
+	  if (execl (argv0, argv0, "child", idx_str.c_str (),
+		     ndev_str.c_str (), (char *) nullptr) == -1)
 	    {
-	      perror ("Failed to exec");
-	      return -1;
+	      perror ("execl");
+	      _exit (127);
 	    }
 	}
     }
 
-  /* Wait for all children.  */
+  /* Reap every child.  Any non-zero exit from a child is a failure
+     (e.g. a runtime initialization failure under contention).  */
+  int failed = 0;
   while (true)
     {
       int ws;
       pid_t ret = waitpid (-1, &ws, 0);
       if (ret == -1 && errno == ECHILD)
 	break;
+      if (ret > 0 && (!WIFEXITED (ws) || WEXITSTATUS (ws) != 0))
+	failed++;
     }
 
   /* Last break here.  */
-  return 0;
+  return failed == 0 ? 0 : 1;
 }
-
-static int
-child (int argc, char **argv)
-{
-  int dev_number;
-  if (sscanf (argv[1], "%d", &dev_number) != 1)
-    {
-      fprintf (stderr, "Invalid argument \"%s\"\n", argv[1]);
-      return -1;
-    }
-
-  CHECK (hipSetDevice (dev_number));
-  kern<<<1, 1>>> ();
-  CHECK (hipDeviceSynchronize ());
-  return 0;
-}
-
-/* When called with no argument, identify how many AMDGPU devices are
-   available on the system and spawn one worker process per GPU.  If a
-   command-line argument is provided, it is the index of the GPU to use.  */
 
 int
 main (int argc, char **argv)
 {
-  if (argc <= 1)
-    return parent (argc, argv);
-  else
+  if (argc >= 2 && strcmp (argv[1], "child") == 0)
     return child (argc, argv);
+
+  int num_children = 0;
+  if (argc >= 2)
+    num_children = atoi (argv[1]);
+
+  return parent (argv[0], num_children);
 }
--- a/gdb/testsuite/gdb.rocm/multi-inferior-gpu.exp
+++ b/gdb/testsuite/gdb.rocm/multi-inferior-gpu.exp
@@ -20,6 +20,8 @@ load_lib rocm.exp
 
 standard_testfile .cpp
 
+source $srcdir/$subdir/multi-inferior.exp.tcl
+
 require allow_hip_tests
 require hip_devices_support_debug_multi_process
 require allow_fork_tests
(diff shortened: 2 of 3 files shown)


gdb/testsuite/lib/rocm.exp
```diff
@@ -640,3 +640,111 @@ gdb_caching_proc compiler_dwarf_language {} {
...
+# run the parent to the pre-fork breakpoint, resume everything in the
+# background, and wait for one "kern" breakpoint stop per child.
+#
+#   args_list  - arguments passed via "set args" ("" for none).
+#   expected   - number of child stops to wait for.  If "", it is read
+#                from the "num_devices" value at the pre-fork stop.
+#
+# Return the list of stopped GPU thread ids (one per child).
+
+proc rocm_multi_inferior_run_to_kernels { args_list expected } {
```
lancesix: Could use `{arg_list ""} {expected ""}` to set default values for those arguments.

lancesix (commented):
I have not really thought deeply about it, but the thing which tickles me here is that this helper implicitly relies on properties of the source file (the markers where to insert breakpoints).

If the functions built around those source assumptions are common, I'd expect the source file to be common as well.  If we get to a point where we have multiple tests using those helpers, it will get harder to keep the source / tcl bits in sync.

It really feels like the .cpp of multi-inferior-gpu should also be made generic if we go this way.

gdb/testsuite/lib/rocm.exp
spatrang: Done — the helper now takes `{ {arg_list ""} {expected ""} }`, and callers that want runtime discovery (multi-inferior-gpu.exp) just call it with no arguments.

spatrang:
> It really feels like the .cpp of multi-inferior-gpu should also be made generic if we go this way.

Done. Rather than add a separate program, I generalized `multi-inferior-gpu.cpp` in place so the breakpoint markers the helpers rely on live in a single source: the child count now comes from `argv` when given and otherwise defaults to the detected GPU device count, and each child re-execs through a `child` argv dispatch. The follow-up stress test (#131) will point `standard_testfile` at this same source instead of carrying its own copy, so the `.cpp` and the `.exp` helpers stay in sync.

gdb/testsuite/gdb.rocm/multi-inferior-gpu.cpp
```diff
@@ -15,12 +15,28 @@
...
+   The parent process forks N children; each child re-execs itself so
+   that its GPU runtime is initialized in a clean address space, then
+   launches its own GPU kernel.  Re-exec'ing makes each child a separate
+   process from the operating system's point of view, which is what the
+   multi-inferior tests debug concurrently.
+
+   N comes from argv[1] when given; otherwise it defaults to the number
+   of GPU devices found at runtime (one child per device).  The
+   companion .exp helpers in lib/rocm.exp plant breakpoints on the
+   pre-fork and post-join source markers and on the kernel.  */
```
lumachad: s/post-join/post-waitpid.

gdb/testsuite/lib/rocm.exp
```diff
@@ -640,3 +640,111 @@ gdb_caching_proc compiler_dwarf_language {} {
 	return "others"
     }
 }
+
```
lumachad: This entire rocm.exp change/code should be in gdb.rocm/ rather than lib/rocm.exp. You just need the common part as a TCL file and then include the file in whatever test that wants to use it. lib/rocm.exp changes are supposed to be for infrastructure purposes or very basic testsuite functionality.

gdb/testsuite/gdb.rocm/multi-inferior-gpu.cpp
```diff
@@ -15,12 +15,28 @@
...
+
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <unistd.h>
```
lumachad: Cosmetic: Group the headers. First sys/, then hip/ then the rest. Or another order, but keep it organized since we're touching this anyway.

spatrang:
Rebased on the latest `amd-staging` and force-pushed (`f0c9558` → `5a0b205`). No functional changes; this only updates the base.

gdb/testsuite/gdb.rocm/multi-inferior-gpu.cpp
spatrang: Fixed — now reads "post-waitpid".

spatrang:
Updated to address the review and squashed the branch into a single commit (force-pushed).

Changes since the last push:
- Moved the two driver helpers out of `lib/rocm.exp` into a new `gdb.rocm/multi-inferior.exp.tcl`, sourced via `$srcdir/$subdir/multi-inferior.exp.tcl`; `lib/rocm.exp` is now unchanged versus the base.
- Fixed the `post-join` -> `post-waitpid` wording in the program header comment.
- Regrouped the includes (`sys/` first, then `hip/`, then the rest).

`gdb.rocm/multi-inferior-gpu.exp` still passes (26/0).

gdb/testsuite/lib/rocm.exp
spatrang: Done. Moved both helpers into a new `gdb.rocm/multi-inferior.exp.tcl` and source it from the test via `$srcdir/$subdir/multi-inferior.exp.tcl` (same pattern as gdb.arch/aarch64-sme-core.exp.tcl). `lib/rocm.exp` is now unchanged versus the base.

gdb/testsuite/gdb.rocm/multi-inferior-gpu.cpp
spatrang: Done — regrouped as sys/ then hip/ then the rest.

lumachad:
Any reason why this wasn't kept with @czidev-amd?

spatrang:
> Any reason why this wasn't kept with @czidev-amd?

Since you’ve already reviewed this PR and have the context, I assumed it would be easier for you to continue as the reviewer.
Happy to switch to auto-assignment if that works better.

palves:
I'm worried that this is being pushed as general reusable infrastructure for testing multi-inferior things when we shouldn't be relying on fork for that.  See:

https://github.com/ROCm/ROCgdb/pull/131#issuecomment-4731383072

Usage of fork in tests should be restricted to testing fork-specific things.

spatrang:
@palves @lumachad — thanks, that's a fair concern and I'd rather settle the mechanism before building more on top of it.

To make sure I take the right direction, which of these do you prefer for the multi-inferior driver?

1. **`add-inferior` within a single GDB** — no `fork` in the test program; portable to Windows once GDB there supports multi-process. The helper would set up N inferiors with `add-inferior` / `file` / `run` instead of forking.
2. **One GDB process per inferior**, coordinated with `spawn_id`s — also fork-free and portable, and maps naturally onto the "many independent GPU processes" stress goal.
3. **Both** — exercise (1) and (2), if the extra coverage is worth the complexity.

My read of the thread is that (1) is the minimal portable replacement and (2) gives the most realistic stress coverage; happy to start with (1) only if (2) is too much for a first cut.

A couple of things I'd like your steer on:
- Should the shared helper (this PR, #166) be reworked to the chosen non-fork mechanism before it lands, or land a minimal version first and evolve it with #131?
- @lumachad, does keeping it as a sourced `gdb.rocm/*.exp.tcl` still match your intent if the mechanism switches from `fork` to `add-inferior`/per-GDB?

I'll hold further changes until we agree on the approach.

lumachad:
I still think this should be sourced as a separate generic/template .tcl file instead of things living in lib/*.exp.

> I'm worried that this is being pushed as general reusable infrastructure for testing multi-inferior things when we shouldn't be relying on fork for that. See:
> 
> [#131 (comment)](https://github.com/ROCm/ROCgdb/pull/131#issuecomment-4731383072)
> 
> Usage of fork in tests should be restricted to testing fork-specific things.

The goal is to expand test coverage for multi-inferior debugging. Right now these tests rely on fork. So it is merely refactoring that code. I think the Windows testsuite run (if/when they run for this case) will still skip these tests due to the use of fork, right?

We should document this is fork-specific/*nix-specific though, if there is a worry someone might misunderstand the restriction.

lumachad:
If we do want to refactor these fork-based tests into something else not using fork, I think that's fair. But then we need to check that against our extra test coverage efforts.

lumachad:
@palves Does the above sound reasonable? I feel we could refactor this some (still using fork) and then refactor it again to not use fork at a later stage.

palves:
> The goal is to expand test coverage for multi-inferior debugging. Right now these tests rely on fork. So it is merely refactoring that code. I think the Windows testsuite run (if/when they run for this case) will still skip these tests due to the use of fork, right?

To recap where this is all coming from:

At the hardware level, there's a number of simultaneous processes that can use the GPU at a time, and that can be debugged simultaneously, somewhere around a dozen or so.  If we debug more than that, things should still work, with the firmware doing transparent context switching.  The goal is to have a testcase that debugs a large-ish number of GPU programs at the same time, to stress test the support across the whole stack, from the hardware to the firmware, to the kernel driver, to dbgapi, and all the way up to gdb.

Looking at the code in this PR, the current "shared driver" code is left in a state that is doing the wrong thing for what we want the new test to do.  We will want to exercise multiple processes _in the same GPU_.   While the refactored "test driver" is baking in the assumption that we run each kernel on a specific GPU with hipSetDevice, even when an explicit children number is passed.  This really all looks like premature refactoring, which complicates more than helps.

> @palves Does the above sound reasonable? I feel we could refactor this some (still using fork) and then refactor it again to not use fork at a later stage.

Refactoring at a later stage from fork to separate programs launched with "run" is largely going to be a rewrite -- the .cpp program is going to look very different, no parent / child in the same .cpp file, no waitpid, etc.  Probabaly will just be a very simple program that launches a kernel, and then GDB launches that program many times and runs to a breakpoint.  It doesn't sound like there's all that much that is interesting to share with the current code.

So to me, trying to reuse fork code from multi-inferior-gpu.cpp, is going the wrong direction.  Afterwards, it might be possible to refactor/rewrite the existing multi-inferior-gpu.exp test to do the same as the new test we add now, and reuse code there.  But until we have the new test in the final form we want, testing what we think it should be testing, it won't be easy to see what that refactor should look like, what should really be moved to shared code.

spatrang:
Closing this PR.

The discussion here and on the JIRA has made clear that this refactoring is premature. The test we actually want (#131) needs to stress **multiple processes on the same GPU** to cross the hardware/firmware multi-debug threshold, whereas the driver extracted here distributes kernels across GPUs via `hipSetDevice(idx % num_devices)` — the wrong behavior for that goal. And the direction we're being asked to move toward is fork-free (`add-inferior` / `start`, so it can also work on Windows), which makes the current fork-based `multi-inferior-gpu.cpp` largely a rewrite rather than something to share.

As Pedro noted, there isn't much worth extracting from the current fork code, so I'll drop this shared-helper refactor. Plan:
- Get #131 into its final intended form first — fork-free, and actually saturating a single GPU beyond the threshold.
- Revisit extracting/refactoring any common driver code (including possibly reworking `multi-inferior-gpu.exp`) only once it's clear what's genuinely reusable.

Thanks all for the review and input.

PR #167, opened 2026-06-11
gdb.rocm/update-thread-list.exp: Drop "patch" from elevation keyword list

palves:
## Motivation

Commit 5003a3aee79 ("gdb.rocm/update-thread-list.exp: Avoid "update" in exe name") listed "patch" among the keywords that trigger Windows' installer-detection heuristic.  That was based on documentation floating around the web, but it turns out to be wrong: an executable whose name contains "patch" runs fine.

Remove "patch" from the comment to avoid confusion.

## Technical Details

Just a comment change.

## Test Plan

Make sure gdb.rocm/update-thread-list.exp still passes.

## Test Result

It still passes.


lancesix (approved):
LGTM, thanks.

PR #168, opened 2026-06-11
gdb, amdgpu: use segment address in amdgpu_get_watchable_aliases

aktemur:
When converting an address to an alias, pass the segment address to dbgapi.  The original core address potentially contains address space bits and it should not be used verbatim.

A regression test could not be included with this patch because the bug is exposed when using a generic address that points to the global address space, and another bug in dbgapi prevents converting such addresses correctly.  However, gdb.rocm/aspace-watchpoint.exp contains tests that will cover the change introduced by this patch when address space information at the type level is handled by GDB.  So, we will have coverage.

## Motivation

Fixing a potential bug.

## Technical Details

See the description.

## Test Plan

There is future coverage.

### DIFF
--- a/gdb/amdgpu-tdep.c
+++ b/gdb/amdgpu-tdep.c
@@ -1832,9 +1832,10 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
 			      ptid_t ptid, int simd_lane,
 			      addr_range range)
 {
-  CORE_ADDR addr = range.addr;
+  CORE_ADDR addr
+    = amdgpu_segment_address_from_core_address (range.addr);
   arch_addr_space_id addr_space_id
-    = amdgpu_address_space_id_from_core_address (addr);
+    = amdgpu_address_space_id_from_core_address (range.addr);
   size_t size = range.size;
 
   /* Addresses from the default address space are always watchable.  */


lancesix (commented):
I am not sure I'd say "and another bug in dbgapi prevents converting such
addresses correctly".  I would call this a limitation, which we are seeing here because we end-up trying to do the address space conversion with a wave_id = AMD_DBGAPI_WAVE_NONE.

dbgapi's interface states in the interface for `amd_dbgapi_convert_address_space`:
 
```
\param[in] wave_id The wave that is using the address.  If the \p
address_space is ::AMD_DBGAPI_ADDRESS_SPACE_GLOBAL then \p wave_id may be
::AMD_DBGAPI_WAVE_NONE, as the address space does not depend on the active
wave[…]
```

so the behaviour you see is expected based on that contract.

That being said, the end of this comment states:

```
[…], in which case \p process_id is used.
```

but amd_dbgapi_convert_address_space does not accept a `process_id` parameter.

We need to change dbgapi to accept a `process_id`.  From the process ID, we **should** be able to figure out the generic aperture configuration, and if address_space is generic, and the address falls in the global aperture, then we indeed can do the conversion without the `wave_id`.

That comment apart, the content of the patch looks good to me.

aktemur:
I updated the commit message.

lancesix (approved):
LGTM, thanks.

PR #170, opened 2026-06-11
[DRAFT][rocgdb] Test filter standardization

dileepr1:
> **Draft / WIP** — opened to exercise CI against the paired TheRock branch. Do **not** merge as-is: `THEROCK_COMMIT_REF` in the workflows points at a feature branch and must be reverted to a pinned `ROCm/TheRock` SHA before merge.

## Summary

Adopts TheRock's test-filter standardization (RFC0010) for ROCgdb, keeping **all** component-specific logic in this repo. TheRock stays component-agnostic — it only invokes its generic `test_runner.py`.

New sibling files under `gdb/testsuite/` (installed verbatim by TheRock's existing testsuite install rule to `tests/rocgdb/gdb/testsuite/`):

- `test_categories.yaml` — source of truth for the `quick`/`standard`/`comprehensive`/`full` tiers (DejaGnu `.exp` selection, labels, timeouts).
- `test_rocgdb.py` — the DejaGnu launcher with a `--tier` flag (honours `TEST_TYPE`); self-locates `rocgdb` + testsuite from `__file__` when run by ctest from the install tree.
- `gen_ctestfile.py` — dev tool that regenerates `CTestTestfile.cmake` from the YAML.
- `CTestTestfile.cmake` — pre-generated + committed; one `add_test()` per tier; each invokes `test_rocgdb.py --tier <name>` via a relative path (ctest sets the working directory to the test dir).

Workflows now invoke `test_runner.py` with `TEST_COMPONENT=rocgdb` / `TEST_TYPE=standard` instead of a per-component script.

## Pairing

Paired with TheRock branch `users/dravindr/tr_rocgdb`, whose net footprint is just:
- `fetch_test_configurations.py`: rocgdb job → `test_runner.py`
- `test_runner.py`: a single `COMPONENT_OVERRIDES["rocgdb"]` entry → `tests/rocgdb/gdb/testsuite`

`THEROCK_COMMIT_REF` in `therock-ci-linux.yml` and `therock-test-packages.yml` is temporarily set to `users/dravindr/tr_rocgdb` so this draft builds/tests against that paired state.

## Flow

```
workflow → test_runner.py (TheRock, generic)
        → ctest -L <tier> --test-dir tests/rocgdb/gdb/testsuite
            → CTestTestfile.cmake (ROCgdb-owned)
                → ./test_rocgdb.py --tier <name> (ROCgdb-owned)
                    → reads ./test_categories.yaml → make check TESTS=...
```

## Test plan

- [ ] TheRock CI Linux build (`gfx94X-dcgpu`) succeeds against `tr_rocgdb`
- [ ] `Run Tests` job exercises `ctest -L standard` → `test_rocgdb.py --tier standard` (GCC + LLVM matrix)
- [ ] Revert `THEROCK_COMMIT_REF` to a pinned `ROCm/TheRock` SHA before un-drafting

🤖 Generated with [Cursor](https://cursor.com)

### DIFF
--- a/.github/test-runner/test_categories.yaml
+++ b/.github/test-runner/test_categories.yaml
@@ -1,0 +1,115 @@
+# ROCgdb Test Categories Configuration
+#
+# Aligned with TheRock's Test Filter Standardization (RFC0010).
+#
+# Source of truth for the tiered (quick/standard/comprehensive/full)
+# DejaGnu `.exp` test selection. This file lives under .github/ (not next
+# to gdb's upstream sources) to keep ROCgdb close to upstream; TheRock's
+# ROCgdb cmake wrapper installs it when present. Consumed by:
+#
+#   - .github/scripts/test_rocgdb.py
+#     Reads --tier <name> [--domain cpu|gpu|corefile|all], expands the
+#     matching test_patterns relative to the installed testsuite dir, and
+#     invokes `make check TESTS=...` for each compiler. TheRock installs this
+#     launcher to tests/rocgdb/test_rocgdb.py and installs this YAML next
+#     to the testsuite at tests/rocgdb/gdb/testsuite/test_categories.yaml.
+#
+#   - TheRock's CI matrix (build_tools/github_actions/fetch_test_configurations.py)
+#     Drives the installed launcher once per runner category, all sharing the
+#     same tier definition:
+#       `test_rocgdb.py --tier <tier> --domain cpu`      (rocgdb-cpu job)
+#       `test_rocgdb.py --tier <tier> --domain gpu`      (rocgdb-gpu job)
+#       `test_rocgdb.py --tier <tier> --domain corefile` (rocgdb-corefile job)
+#
+# Patterns are matched by Python `fnmatch` against paths relative to the
+# testsuite dir (e.g. "gdb.rocm/simple.exp" or "gdb.dwarf2/*.exp").
+# --domain splits a tier into three disjoint runner categories (their union
+# is the whole tier, with no test running twice):
+#   cpu      - everything except gdb.rocm/*        (CPU-only runners)
+#   gpu      - gdb.rocm/* minus the corefile tests (standard GPU runners)
+#   corefile - the corefile_tests listed below     (core-dump GPU runners)
+#   all      - every test (the default)
+#
+# Note: per-compiler XFAILED lists (GCC- and LLVM-only known failures) are
+# kept inline in test_rocgdb.py's XFAILED_TESTS dict; they are orthogonal
+# to tier selection and only suppress reporting of expected failures.
+
+# Corefile tests: the gdb.rocm/* tests marked `require allow_rocm_core_tests`.
+# They exercise host core-dump handling and therefore need a GPU runner whose
+# host kernel core_pattern / ulimit are provisioned to write plain core files
+# (plus the coremerge tool). --domain corefile selects exactly this set, and
+# --domain gpu excludes it, so the two GPU shards never overlap. Kept as a
+# top-level list (not per tier) since membership is a property of the test.
+corefile_tests: &corefile_tests
+  - "gdb.rocm/corefile.exp"
+  - "gdb.rocm/core-no-read-special-files.exp"
+  - "gdb.rocm/gcore-after-attach.exp"
+  - "gdb.rocm/load-core-remote-system.exp"
+  - "gdb.rocm/runtime-core.exp"
+
+# Shared exclusion list — keep in sync with the "Generic" entries of
+# test_rocgdb.py's XFAILED_TESTS["Generic"]. These tests have known
+# infrastructure-level failures that affect every tier and every compiler.
+# (corefile.exp / load-core-remote-system.exp are NOT excluded here: they are
+# routed to the dedicated corefile runner category above instead.)
+_common_excludes: &common_excludes
+  - "gdb.rocm/device-interrupt.exp"
+
+test_categories:
+  quick:
+    description: "Smoke checks (target: < 5 min) - tiny representative subset"
+    test_patterns:
+      - "gdb.rocm/simple.exp"
+      - "gdb.rocm/generic-address.exp"
+      - "gdb.rocm/show-info.exp"
+      - "gdb.rocm/hip-lang-detect.exp"
+      - "gdb.dwarf2/atomic-type.exp"
+      - "gdb.dwarf2/count.exp"
+      - "gdb.dwarf2/comp-unit-lang.exp"
+      - "gdb.dwarf2/bad-regnum.exp"
+    exclude: *common_excludes
+    labels:
+      - "quick"
+      - "pre-commit"
+      - "smoke"
+
+  standard:
+    description: "Pre-checkin validation (target: < 45 min) - all GPU + DWARF tests"
+    test_patterns:
+      - "gdb.rocm/*.exp"
+      - "gdb.dwarf2/*.exp"
+    exclude: *common_excludes
+    labels:
+      - "standard"
+      - "pr"
+      - "precheckin"
+
+  comprehensive:
+    description: "Nightly comprehensive (target: < 4 hours) - same surface as standard, currently no extra tier"
+    test_patterns:
+      - "gdb.rocm/*.exp"
+      - "gdb.dwarf2/*.exp"
+    exclude: *common_excludes
+    labels:
+      - "comprehensive"
+      - "nightly"
+      - "extended"
+
+  full:
+    description: "Complete - GPU + DWARF (placeholder for adding extended suites later)"
+    test_patterns:
+      - "gdb.rocm/*.exp"
+      - "gdb.dwarf2/*.exp"
+    exclude: *common_excludes
+    labels:
+      - "full"
+      - "all"
+
+execution_settings:
+  default_timeout: 300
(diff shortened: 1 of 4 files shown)


.github/workflows/therock-ci-linux.yml
```diff
@@ -14,7 +14,10 @@ permissions:
   contents: read
 
 env:
-  THEROCK_COMMIT_REF: 25083b016c475a65cbea46ae7326104680bff1d5 # 2026-06-08
```
lumachad: These workflows will disappear in favor of a multi-arch CI cross-trigger implementation. I'm not sure how this affects the test filter standardization plans, but I thought I'd mention it. We don't want to patch it now only to have to discard it and do it some other way later.

.github/test-runner/CTestTestfile.cmake
```diff
@@ -0,0 +1,33 @@
...
+
+add_test(rocgdb_comprehensive "python3" "../../test_rocgdb.py" "--tier" "comprehensive" "--try-rocm-path")
+set_tests_properties(rocgdb_comprehensive PROPERTIES
+    LABELS "comprehensive;nightly;extended"
+    TIMEOUT 14400)
+
+add_test(rocgdb_full "python3" "../../test_rocgdb.py" "--tier" "full" "--try-rocm-path")
+set_tests_properties(rocgdb_full PROPERTIES
+    LABELS "full;all"
+    TIMEOUT 28800)
```
lumachad: Given ROCgdb is upstream-based and we would like to keep it as close to upstream as possible, we don't want to add extraneous files to the sources, which means these files can't be put next to gdb's files. The exception here is we have .github/ storing workflows and some CI scripts.

So I think we could store these files in .github/test-runner and install them into the ROCm tree via TheRock's ROCgdb cmake wrapper in <therock>/debug-tools/rocgdb/CMakeLists.txt. So this would be a 2-part change.

For instance, see https://github.com/ROCm/TheRock/pull/5720. We always want to make the installation optional, as the wrapper is also supposed to work with upstream gdb sources that don't contain these files.

.github/workflows/therock-test-packages.yml
```diff
@@ -66,5 +73,18 @@ jobs:
...
+          # from test_categories.yaml and drives the GCC/LLVM DejaGnu matrix.
+          # All rocgdb-specific logic lives in this repo; TheRock stays generic.
+          TEST_COMPONENT: rocgdb
+          # Use the curated `quick` smoke tier for PR CI: it finishes within the
+          # job timeout. The broader standard/comprehensive/full tiers run the
+          # full DejaGnu matrix and are intended for nightly/extended pipelines.
+          TEST_TYPE: quick
         run: |
-          python ${{ env.OUTPUT_ARTIFACTS_DIR }}/tests/rocgdb/test_rocgdb.py
+          python build_tools/github_actions/test_executable_scripts/test_runner.py
```
lumachad: We have our own ROCgdb CI running via the workflows, and I'm not sure we want to run test_runner.py here. I think what we might want is to trigger test_runner.py in TheRock. But I think this might depend on the upcoming multi-arch CI changes being done by @geomin12, myself and others.

.github/test-runner/test_categories.yaml
```diff
@@ -0,0 +1,96 @@
...
+#   python3 gdb/testsuite/gen_ctestfile.py
+#
+# Note: per-compiler XFAILED lists (GCC- and LLVM-only known failures) are
+# kept inline in test_rocgdb.py's XFAILED_TESTS dict; they are orthogonal
+# to tier selection and only suppress reporting of expected failures.
+
+# Shared exclusion list — keep in sync with the "Generic" entries of
+# test_rocgdb.py's XFAILED_TESTS["Generic"]. These tests have known
+# infrastructure-level failures that affect every tier and every compiler.
+_common_excludes: &common_excludes
```
lumachad, line 55: I suppose the ROCgdb team will have to come up with suitable lists of tests to fit these categories, right?

lumachad (commented):
Thanks for the PR. Some preliminary comments. I think it's mostly OK, pending adjusting location of files, lists of tests and validation.

.github/test-runner/CTestTestfile.cmake
dileepr1: Done. 
Moved them out of gdb/testsuite/ into .github/test-runner/ (next to .github/scripts/test_rocgdb.py). 
TheRock's ROCgdb wrapper now installs test_categories.yaml and CTestTestfile.cmake into tests/rocgdb/gdb/testsuite/ via EXISTS-guarded rules (paired branch users/dravindr/tr_rocgdb), so absence is a no-op and it stays compatible with upstream gdb — matching the pattern in #5720.

.github/test-runner/test_categories.yaml
dileepr1: Correct.
The tests present here are added as a sample.
rocgdb owns this file and can add/remove tests as needed.

.github/workflows/therock-test-packages.yml
dileepr1: The therock-ci-linux.yml / therock-test-packages.yml edits (and the THEROCK_COMMIT_REF pin) are intentionally temporary — just enough to validate test_runner.py driving the tiered tests end-to-end. They'll be removed before merge, and a follow-up TheRock-side PR will own triggering the tests from TheRock CI.

.github/workflows/therock-ci-linux.yml
dileepr1: This change is temporary for validation and will be removed before merge.

lumachad:
Thanks for the updates. Let me go through this again.

lumachad:
Let's wait until https://github.com/ROCm/ROCgdb/pull/117 gets merged, as that one has a large change to test_rocgdb.py. Since this one touches that same script, we should avoid needless conflict resolution cycles.

lumachad:
https://github.com/ROCm/ROCgdb/pull/117 has been merged. I think we can rebase the changes in this PR on top of current test_rocgdb.py.

lumachad:
@dileepr1 Thanks. Let me go through this again and see what we should put into the test categories.

lumachad:
@dileepr1 With rocgdb testing we have both CPU and GPU tests that must be routed to the appropriate runners to make the best use of resources. Can these filters handle that?

dileepr1:
> @dileepr1 With rocgdb testing we have both CPU and GPU tests that must be routed to the appropriate runners to make the best use of resources. Can these filters handle that?

I'll make the changes for this.

lumachad:
Hi @dileepr1, I don't think we want to push the gpu/cpu part into test_rocgdb.py. It already knows about that via --gpu-tests and --cpu-tests. What we want to support is sharding different types of tests to their required runners.

For instance, for rocgdb we currently have the cpu tests (runs only gdb.dwarf2), gpu tests (runs only gdb.rocm) and soon we will also have the corefile tests. All 3 categories need specific runners. Can we split those at a different level that is not test_rocgdb.py itself?

If so, we wouldn't even need to modify test_rocgdb.py, and we would just modify test_categories.yml to contain that information, like TheRock's test matrix/configuration.

dileepr1:
Thanks @lumachad, that makes sense.

You're right that the cpu/gpu split already lives at a level above the launcher — it's in TheRock's fetch_test_configurations.py (the rocgdb-cpu / rocgdb-gpu components with their --tests gdb.dwarf2 / --tests gdb.rocm selection and runner assignment). Our --domain addition just duplicated that split down into test_rocgdb.py, which is the wrong layer.

So we'll revert the --domain change and keep test_rocgdb.py generic (using the existing --tests / --gpu-tests / --cpu-tests), leaving the category → runner sharding entirely at the matrix level.

For the standardization part —  could you share a sample test_categories.yml in the shape you envision (how you'd express the cpu / gpu / corefile categories, their test sets, runner requirements, and any tiering)? We'll build the framework around the structure you prefer so it lines up with how you'd want TheRock's matrix/config to consume it.

lumachad:
Hi @dileepr1. I was thinking about this. Given we have our test script (test_rocgdb.py), we can carry on from the approach you've taken of teaching the script about the tiers (quick/standard/comprehensive/full). But we could make the tiers generic and based on the yaml file, such that we could pass --tier <tier name>.

The distinction here is we will have 3 categories of tests that need to be executed each in distinct runners.

- (1) GPU tests run on GPU runners
- (2) GPU corefile tests run on special GPU runners given their corefile-specific settings
- (3) CPU tests run on GPU-less runners

I think that needs to be taken care of by upper layers. Maybe the workflow or TheRock. So each tier would fire up the tier's tests for each one of the 1/2/3 runners.

We would run something like this for each tier:

- test_rocgdb.py --tier <tier name> --gpu (case 1, GPU runners)
- test_rocgdb.py --tier <tier name> --corefile (case 2, GPU corefile runners)
- test_rocgdb.py --tier <tier name> --cpu (case 3, CPU runners).

Do you think that could work? As for the individual tests for each tier, we would be in charge of determining those. But your initial list is fine for us.

dileepr1:
@lumachad I've updated the PR as per above, could you re-review please?

lumachad:
> @lumachad I've updated the PR as per above, could you re-review please?

I have this on my TODO, but have a few things I need to get out of the way first, sorry.

lumachad:
Still on my list.

PR #173, opened 2026-06-12
gdb: Fix gdb.rocm/mi-attach.exp FAILs on Windows

czidev-amd:
gdb.rocm/mi-attach.exp FAILs on Windows because -target-attach <pid> emits ^running as its result-record instead of ^done. The test matcher .\^done. never fires, so it fails as "unexpected output". The bug is not in the test — ^done is what the MI protocol requires for a synchronous command with set mi-async off, and Linux has been emitting it correctly since Feb 2025. Windows still emits ^running because windows-nat.c::do_initial_windows_stuff has a stale clear_proceed_status call that, in combination with the MI interp's ^running/^done back-compat hack, latches mi_proceeded=1 before the dispatcher gets to print ^done, which then gets suppressed.

This patch removes the stale clear_proceed_status call.

### DIFF
--- a/gdb/testsuite/gdb.mi/mi-attach.c
+++ b/gdb/testsuite/gdb.mi/mi-attach.c
@@ -1,0 +1,37 @@
+/* Copyright 2026 Free Software Foundation, Inc.
+
+   This file is part of GDB.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.  */
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#ifdef _WIN32
+#include <windows.h>
+#else
+#include <unistd.h>
+#endif
+
+int
+main (void)
+{
+  /* Sleep long enough (5 min) to be attached to.  */
+#ifdef _WIN32
+  Sleep (300000);
+#else
+  sleep (300);
+#endif
+  return 0;
+}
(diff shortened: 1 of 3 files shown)


gdb/windows-nat.c
```diff
@@ -1944,7 +1944,6 @@ windows_nat_target::do_initial_windows_stuff (DWORD pid, bool attaching)
   if (!inf->target_is_pushed (this))
     inf->push_target (this);
   windows_clear_solib ();
-  clear_proceed_status (0);
   init_wait_for_inferior ();
```
aktemur, line 1947: In `init_wait_for_inferior` we call `clear_proceed_status (0, false);`.:

```
/* Clear out all variables saying what to do when inferior is
   continued or stepped.  First do this, then set the ones you want,
   then call `proceed'.  STEP indicates whether we're preparing for a
   step/stepi command.  Set ABOUT_TO_PROCEED to false if we're not
   calling `proceeed` yet.  */
extern void clear_proceed_status (int step, bool about_to_proceed = true);
```
Based on this, the patch LGTM.

aktemur (approved):

palves (commented):
This LGTM indeed.  There is no reason to hold this downstream though.  This should be sent upstream (please CC me).  Ideally, there'd be a non-gdb.rocm/ testcase what would exercise this, too.  Maybe there's one already.

Also, it's fine to use claude to help with writing commit logs, but I'd be preferable to make sure that comes out is a bit more humanized.

- em-dash, "latches" are dead giveaways.
- I don't know what the "Feb 2025" is supposed to be telling.
- please show a snippet of the FAIL
- use paragraphs.
- where's that test matcher?

Also, "patche" typo.

amd-shahab:
from the PR description, which is hard-to-grasp and different from the commit message, I can't see what problem this "redundant" call cause. Could you explain how this leads to a failure on Windows?

gdb/testsuite/gdb.mi/mi-attach.c
```diff
@@ -0,0 +1,19 @@
+#include <stdio.h>
```
lancesix, line 18: Just FYI, this file is missing the copyright header.

lumachad:
If this is supposed to go upstream, let's not have a PR here.

Also gdb.mi never gets exercised in CI, so it is a bit of a waste of resources. Let's mark it ci:skip.

czidev-amd:
Committed upstream: https://sourceware.org/pipermail/gdb-patches/2026-June/228225.html

PR #174, opened 2026-06-16
A navigation sidebar for the HTML manuals

palves:
The HTML build of texinfo manuals emits one bare page per node, with
only the texinfo prev/next/up links to get around.  For ROCgdb, we
want something closer to a modern documentation site: a navigation
tree that stays visible on the left of every page.

Unfortunately, texinfo has no built-in support for a sidebar, but OTOH
fortunately it can be driven by a texi2any init file, which is enough
to build a sidebar ourselves, with custom CSS and JavaScript.  The
init file we're adding uses a texi2any init API introduced in version
7.0, and falls back to the default contents-only manual on older
texi2any.  The texinfo version that comes with Ubuntu 22.04 is too old
(6.8), but the version in Ubuntu 24.04 (7.1) is good.

This series adds the sidebar, built from the manual's own table of
contents so there is no separate data to keep in sync.  The sidebar
needs JavaScript, and with JavaScript off the page falls back to the
plain single-column layout the manual had before.  The sidebar
remembers its position and set of open nodes across page navigation.
On top of the base sidebar, the series adds a filter box that narrows
the tree as you type, and a draggable divider to customize the sidebar
width.

One of the new files is a Perl file, but that's not by choice.  The
sidebar is driven by a new texi2any init file, toc-sidebar.init.
texi2any is itself a Perl program and loads its --init-file into its
own interpreter, so the init file has to be Perl.  This is the same
way texinfo's own bundled HTML customization is done.

The series targets the split HTML build (one page per node), which is
what the AMD documentation site publishes.  The single-page manual
(makeinfo --no-split, the gdb.html linked from
https://sourceware.org/gdb/documentation/) is built separately and
does not use the init file, so it stays as it is today.  The sidebar's
design assumes per-page navigation and does not carry over to a single
page -- that mode would require a different (and probably simpler)
implementation.

The sidebar comes with a theme/style that works well with the default
texinfo style.

For ROCgdb, we'd like to use the same theme used by the rest of the
ROCm documentation website.  For that, the sidebar is theme aware, and
the series also adds support for theming, via a MANUAL_CSS Makefile
variable.  For the theme/stylesheet to use, the series adds a
rocm-theme.css file that mimics the ROCm documentation website's
style.  Part of it was automatically generated by a new
rocm-theme-sync.py script, also included in the series.

# Preview URLs

## No sidebar (as currently)

https://palves.github.io/gdb-manual-preview/toc-pane/bare/gdb/
https://palves.github.io/gdb-manual-preview/toc-pane/bare/binutils/
https://palves.github.io/gdb-manual-preview/toc-pane/bare/gas/
https://palves.github.io/gdb-manual-preview/toc-pane/bare/ld/
https://palves.github.io/gdb-manual-preview/toc-pane/bare/bfd/

## Sidebar + default theme

https://palves.github.io/gdb-manual-preview/toc-pane/default/gdb/
https://palves.github.io/gdb-manual-preview/toc-pane/default/binutils/
https://palves.github.io/gdb-manual-preview/toc-pane/default/gas/
https://palves.github.io/gdb-manual-preview/toc-pane/default/ld/
https://palves.github.io/gdb-manual-preview/toc-pane/default/bfd/

## Sidebar + ROCm theme

https://palves.github.io/gdb-manual-preview/toc-pane/rocm/gdb/
https://palves.github.io/gdb-manual-preview/toc-pane/rocm/binutils/
https://palves.github.io/gdb-manual-preview/toc-pane/rocm/gas/
https://palves.github.io/gdb-manual-preview/toc-pane/rocm/ld/
https://palves.github.io/gdb-manual-preview/toc-pane/rocm/bfd/


texinfo/toc-sidebar/toc-sidebar.js
```diff
@@ -0,0 +1,388 @@
+/* Copyright (C) 2026 Free Software Foundation, Inc.
```
lancesix, line 1: I have not tested, but is such header correclty detected by LibreJS?

lancesix (commented):
I only did a very brief read through, and 😍 !

Thank for doing this.

Just a simple question, and overall, given that we only ship GDB (+ coremerge) and do not ship LD / GAS / ..., do we need to theme those?

palves:
Added notes about texinfo API and HTML stability, after digging through the texinfo sources and NEWS in some more detail.  gnu.org seems to be down a lot nowadays...  https://gnu.googlesource.com/texinfo seems to be the next best thing.

palves:
> I only did a very brief read through, and 😍 !
> 
> Thank for doing this.
> 
> Just a simple question, and overall, given that we only ship GDB (+ coremerge) and do not ship LD / GAS / ..., do we need to theme those?

The current docs generation process does generate the html of all of those, e.g., 

 https://rocm.docs.amd.com/projects/ROCgdb/en/latest/ROCgdb/ld/doc/ld/index.html .

There are some xrefs into binutils in the gdb manual, but now that I check, they link to sourceware.org.
E.g., see the link to objdump here:
https://rocm.docs.amd.com/projects/ROCgdb/en/latest/ROCgdb/gdb/doc/gdb/Machine-Code.html

Do we have a link to the coremerge docs in the gdb manual?  I couldn't find it.  If we do, it'd make sense to carry the binutils docs too.  Otherwise, probably not.  The non-gdb manuals we host may be unreachable, but not 100% sure.

In any case, at first, I made this work for GDB (and used gdb-prefixed symbols in the css, javascript, etc, which was a pain to generalize later :-P) , and the new files all lived inside the gdb/doc/ dir.  Then I realized that none of what I had was GDB-specific and should work with any texinfo manual, so I tried wiring it into the ld, gas, etc manuals .  
That paid off as it exposed a couple bugs.  

Also, keeping the idea of upstreaming all this except the rocm theme, I thought I'd try to see how hard it'd be to have a single MANUAL_CSS Makefile variable that works on all manuals in the tree.  If people like that, I should explore making sure the top level Makefile passes down MANUAL_CSS to the submakes, and then we wouldn't have to hack in the rocm-theme.css file in each of the tool's makefiles, and if we wanted, in none, we could pass MANUAL_CSS at build time.  So that's mainly why I did that.  Just to see that is all worked, and to decide whether this would be a GDB-only thing, or for all the manuals in the tree.

PR #175, opened 2026-06-16
[2026-06-16] Merge remote-tracking branch 'origin/master' into amd-staging

lancesix:
Handle conflicts introduced by recent upstream changes.  Conflicting
upstream changes are:
- c4aa08decda gdb/amd-dbgapi-target: Move Linux code to amd-dbagapi-posix-hdep.c
- f67ddc1be6e gdb/amd-dbgapi-target: Update xfer_partial to always use dbgapi
- c04f1f54f6c gdb/testsuite: Add mechanism to compile Windows native programs on Cygwin
- 2b5689d3792 [gdb/contrib] Describe words in codespell-ignore-words.txt
    
Most changes were trivial.  The main notable changes are in
gdb/testsuite/lib/gdb.exp, becauso of downstream change 27f8b77806e "HIP
and set_unbuffered_mode_saved.o".

This merge commit also corrects one typo as codespell have been enabled
for the gdb directory by 664cc2deff7 [pre-commit] "Include gdb in
codespell check".


palves:
The set_unbuffered_mode_saved.o changes have a problem.  

To see it easily, you can force that code path to run on linux too with this hack:

```
diff --git i/gdb/testsuite/lib/gdb.exp w/gdb/testsuite/lib/gdb.exp
index a38178a24ed..0b6405717bf 100644
--- i/gdb/testsuite/lib/gdb.exp
+++ w/gdb/testsuite/lib/gdb.exp
@@ -6923,9 +6923,9 @@ proc gdb_compile {source dest type options} {
   }
 
     if { $type == "executable" } {
-       if { ([istarget "*-*-mingw*"]
+       if {1 || ( ([istarget "*-*-mingw*"]
              || [istarget "*-*-*djgpp"]
-             || [istarget "*-*-cygwin*"])} {
+             || [istarget "*-*-cygwin*"]))} {
            # Force output to unbuffered mode, by linking in an object file
            # with a global constructor that calls setvbuf.
            #
```

and with that, we get:

```
Running /home/pedro/rocm/gdb/build/gdb/testsuite/../../../src/gdb/testsuite/gdb.rocm/simple.exp ...
ERROR: tcl error sourcing /home/pedro/rocm/gdb/build/gdb/testsuite/../../../src/gdb/testsuite/gdb.rocm/simple.exp.
ERROR: tcl error code TCL LOOKUP SUBCOMMAND basename
ERROR: unknown or ambiguous subcommand "basename": must be atime, attributes, channels, copy, delete, dirname, executable, exists, extension, isdirectory, isfile, join, link, lstat, mkdir, mtime, nativename, normalize, owned, pathtype, readable, readlink, rename, rootname, separator, size, split, stat, system, tail, tempfile, type, volumes, or writable
    while executing
"file basename  [file rootname $::gdb_saved_set_unbuffered_mode_obj_name  ]  "
    (procedure "gdb_compile" line 514)
    invoked from within
```

"file basename" doesn't exist.  It's spelled "file tail" instead.

$ rlwrap tclsh 
file tail a/b/c.o
c.o

But then I don't understand why we'd need to extract the basename, since gdb_saved_set_unbuffered_mode_obj_name is always a basename, never a full path.  So I think not calling "file tail" or anything at all would work.  Like:

		set unbuf_obj_basename \
		    [file rootname $::gdb_saved_set_unbuffered_mode_obj_name]-$unbuf_obj_lang.o

That works for me on Linux with the hack above.  Not sure if I missed something, though.

While at it, typo in commit log: becauso => because

lancesix:
Fixed, applied the change as suggested.

lancesix:
I had to rebase the merge as well, here is the delta between rebase and fixed patch:

```
diff --git a/gdb/amd-dbgapi-target.c b/gdb/amd-dbgapi-target.c
index 0d905e93eca..0e89953c283 100644
--- a/gdb/amd-dbgapi-target.c
+++ b/gdb/amd-dbgapi-target.c
@@ -992,7 +992,7 @@ amd_dbgapi_target::thread_name (thread_info *tp)
 
      Some kernel entry point functions have very long names, so we
      have a hardcoded length limit.  Very long names are not really
-     useful to users, and would make "info threads" output unuseable
+     useful to users, and would make "info threads" output unusable
      otherwise.  If the function name is longer than the limit, we
      trim it and append a 4-characters hash of the trimmed function
      name so that two trimmed function names with the same prefix
diff --git a/gdb/testsuite/lib/gdb.exp b/gdb/testsuite/lib/gdb.exp
index d1b6328cee8..ae009aeaf50 100644
--- a/gdb/testsuite/lib/gdb.exp
+++ b/gdb/testsuite/lib/gdb.exp
@@ -6972,9 +6972,7 @@ proc gdb_compile {source dest type options} {
                # which is important for GDB_PARALLEL.  See comments
                # above.
                set unbuf_obj_basename \
-                   [file basename \
-                     [file rootname $::gdb_saved_set_unbuffered_mode_obj_name \
-                     ] \
+                   [file rootname $::gdb_saved_set_unbuffered_mode_obj_name \
                    ]-$unbuf_obj_lang.o
                set unbuf_obj [standard_temp_file $unbuf_obj_basename]
```

palves (approved):

PR #177, opened 2026-06-18
gdb/testsuite/rocm/runtime-core: ensure all or no test run

lancesix:
When working on coredump related issue, it was noticed that some runtime changes could work in some conditions but fail in others.  The gdb.rocm/runtime-core.exp testcase was not really complaining about that, it would validate the core dumps when produced and skip the test if the core was absent.

It is expected that if the system configuration is not adequate, none of the runtime-core cases would run, but if at least one can run, it means all must also run.

This test ensures that we run either all or none of the testcase, and issue a FAIL if this property is not satisfied.

Change-Id: Ic79b958a0991b21dc1dfeb6bb13b3f292c53261f

### DIFF
--- a/gdb/testsuite/gdb.rocm/runtime-core.exp
+++ b/gdb/testsuite/gdb.rocm/runtime-core.exp
@@ -33,6 +33,15 @@ if { [build_executable "failed to prepare" ${testfile} ${srcfile} \
     return -1
 }
 
+# The "do_test" function is called multiple times, and any may exit early if
+# it turns out the current configuration does not allow the creation of core
+# dumps.  However, if at least one core dump can be produced, all must be.
+# Keep track of the number of attempts and their status to enforce this
+# property.
+set total_count 0
+set untested_count 0
+set success_count 0
+
 # do_test FAULT CORE_TYPE OUTPUT_TYPE
 #   Run the test exercising generation of a core dump on fault FAULT
 #   using a file or a pipe.
@@ -45,6 +54,7 @@ if { [build_executable "failed to prepare" ${testfile} ${srcfile} \
 # OUTPUT_TYPE is one of "file" of "pipe".
 #
 proc do_test { fault core_type output_type } {
+    incr ::total_count
     set use_pipe [expr {$output_type eq "pipe"}]
     save_vars { ::env(HSA_ENABLE_LIGHTWEIGHT_COREDUMP) } {
 	set ::env(HSA_ENABLE_LIGHTWEIGHT_COREDUMP) [expr {$core_type == "lightweight"}]
@@ -53,6 +63,7 @@ proc do_test { fault core_type output_type } {
 
     if {$coredump eq ""} {
 	untested "Could not generate a core file"
+	incr ::untested_count
 	return
     }
 
@@ -134,6 +145,7 @@ proc do_test { fault core_type output_type } {
     }
 
     gdb_exit
+    incr ::success_count
 }
 
 with_rocm_gpu_lock {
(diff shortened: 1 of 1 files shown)


gdb/testsuite/gdb.rocm/runtime-core.exp
```diff
@@ -134,6 +145,7 @@ proc do_test { fault core_type output_type } {
     }
 
     gdb_exit
+    uncr ::success_count
```
lumachad: s/uncr/incr

lancesix: Fixed (sorry, did not wait for assignee to change)

lumachad:
With the fix I now see the following:

```
Running /develop/work/repos/TheRock/build/dist/rocm/tests/rocgdb/gdb/testsuite/gdb.rocm/runtime-core.exp ...
FAIL: gdb.rocm/runtime-core.exp: coredump consistently produced

                === gdb Summary ===

# of expected passes            53
# of unexpected failures        1
# of untested testcases         2
```

Probably need an extra fix to rocr-runtime?

lumachad (approved):
The change itself is OK for me.

lancesix:
> With the fix I now see the following:
> 
> ```
> Running /develop/work/repos/TheRock/build/dist/rocm/tests/rocgdb/gdb/testsuite/gdb.rocm/runtime-core.exp ...
> FAIL: gdb.rocm/runtime-core.exp: coredump consistently produced
> 
>                 === gdb Summary ===
> 
> # of expected passes            53
> # of unexpected failures        1
> # of untested testcases         2
> ```
> 
> Probably need an extra fix to rocr-runtime?

would you have the gdb.log with those 2 untested entries I could look at?

gdb/testsuite/gdb.rocm/runtime-core.exp
```diff
@@ -33,6 +33,15 @@ if { [build_executable "failed to prepare" ${testfile} ${srcfile} \
     return -1
 }
 
+# The "do_test" function is called for multiple times, but for each can exit
+# early if it turns out the current configuration does not allow the creation
+# of core dump.  However, if at least one core dump can be produced, all must
```
palves: I think the "for" word shouldn't be in the first sentence, in the two spots it appears.  

"core dump" => "core dumps".

I'd suggest:

```
# The "do_test" function is called multiple times, and any may exit
# early if it turns out the current configuration does not allow the
# creation of core dumps.
```

gdb/testsuite/gdb.rocm/runtime-core.exp
```diff
@@ -33,6 +33,15 @@ if { [build_executable "failed to prepare" ${testfile} ${srcfile} \
...
+# be.  Keep track of the number of attempts and status to enforce this
+# property.
```
palves, line 40: Suggest adding "their":

```
# Keep track of the number of attempts and their status to enforce this property.
```

lancesix:
Applied suggestions to comments.

aktemur (approved):
Looks reasonable to me.

palves (approved):

aktemur:
@lancesix, can this be merged?

lumachad:
We need to re-trigger CI and see if the tests pass. The previous run skipped rocgdb testing.

lancesix:
> @lancesix, can this be merged?

I think so. If I emember correctly, it was help as it could start to issues fails due to some other bug, but should be ok now (otherwise, CI will eventually let me know).

lumachad:
> We need to re-trigger CI and see if the tests pass. The previous run skipped rocgdb testing.

Actually, this is a corefile test. Se we should validate it locally.

PR #129, opened 2026-05-15
gdb: add "catch hiperr" to break on HIP API errors

amd-shahab:
This change relies on "Add __hipOnError debugger hook for HIP API failures" [PR](https://github.com/ROCm/rocm-systems/pull/4755) to be functional. Although, without it, ROCgdb won't be broken. Please read the commit message for the description of this change itself.

--

What
----
This change adds a new feature to GDB, when it is targeted for
amdgcn-amd-amdhsa, that enables it to stop whenever HIP API
errors are returned.  An abridged sample of execution looks
like this:

(gdb) catch hiperr
(gdb) run
Thread 1 "hip" hit Catchpoint 1 (HIP error)
hipErrorInvalidDevice (101): invalid device ordinal

Moreover, a $_hiperr convenience variable is added that holds
the error code.  For further details, please see the updated
documentation in this change.

Design
------
Most of the implementation is in the newly added module called
break-catch-hiperr.c.  The naming and its context follows the
break-catch-throw.c module.

The whole mechanism work by setting a breakpoint (catchpoint)
on a __hipOnError() symbol, that must be provided by the CLR
library.  Enabling the catchpoint when there's no __hipOnError
symbol, simply has no effect.

There's a architect dependent section that is responsible for
extracting the error parameters (code, name, description) from
a single argument passed to the __hipOnError() function as a
structure pointer.  The expected format of this struct is:

  uint32_t    version;   // at this point, only 1 is supported
  uint32_t    code;      // the numerical error code
  const char *name;      // name of the error
  const char *desc;      // a phrase about the error

The code is devised in such a way that if the __hipOnError()
exists, but for whatever reason the parameters cannot be deduced,
it won't interfere with the catchpoint being triggered.  It will
affect the reported data of the catchpoint though.

Test
----
A new test, gdb.rocm/hip-catch-errors, is added to the mix to
ensure that:

1. (Pending) catchpoints work.
2. Multiple catchpoints in one execution are reported accordingly.
3. Errors of indirect HIP calls like "kenel<<<...>>>()" are caught.
4. The convenience variable can be used for conditional situations.
5. Conditional and temporary catchpoints work as expected.

ROCM-1548

PR #180, opened 2026-06-19
gdb: add -catch-hiperr to the Machine Interface (MI)

amd-shahab:
Introduce a new MI command to catch HIP runtime errors that takes the form of:
```
  -catch-hiperr [ -c <condition> ] [ -t ]
```

Along with it the documentation is updated and new tests are added.

## Motivation

Add Machine Interface for "catch hiperr".

## Technical Details

Machine Interface allows other tools control GDB by sending it commands and reading its feedback. This was missing for the "catch hiperr" support that we added [earlier](https://github.com/ROCm/ROCgdb/pull/129).

## Test Plan

New MI tests are added `hip-catch-errors.exp`.

## Test Result

All the new tests are and "catch hiperr" tests are passing.
```
Running /data/work/src/dev/rocgdb/gdb/testsuite/gdb.rocm/hip-catch-errors.exp ...
PASS: gdb.rocm/hip-catch-errors.exp: mi_parsing: trailing -t in condition: mi-catch-hiperr
PASS: gdb.rocm/hip-catch-errors.exp: mi_parsing: trailing -t in condition: breakpoint info
PASS: gdb.rocm/hip-catch-errors.exp: mi_parsing: dangling -c: mi-catch-hiperr
PASS: gdb.rocm/hip-catch-errors.exp: mi_parsing: unknown option: mi-catch-hiperr
PASS: gdb.rocm/hip-catch-errors.exp: mi_parsing: dangling texts: mi-catch-hiperr
PASS: gdb.rocm/hip-catch-errors.exp: mi_parsing: dangling texts: breakpoint info
[ CLI tests in here ]
PASS: gdb.rocm/hip-catch-errors.exp: mi_test: simple catch: mi-catch-hiperr
PASS: gdb.rocm/hip-catch-errors.exp: mi_test: simple catch: breakpoint info
PASS: gdb.rocm/hip-catch-errors.exp: mi_test: simple catch: set argument
PASS: gdb.rocm/hip-catch-errors.exp: mi_test: simple catch
PASS: gdb.rocm/hip-catch-errors.exp: mi_test: temporary catch: mi-catch-hiperr
PASS: gdb.rocm/hip-catch-errors.exp: mi_test: temporary catch: breakpoint info
PASS: gdb.rocm/hip-catch-errors.exp: mi_test: temporary catch: set argument
PASS: gdb.rocm/hip-catch-errors.exp: mi_test: temporary catch
PASS: gdb.rocm/hip-catch-errors.exp: mi_test: conditional catch: mi-catch-hiperr
PASS: gdb.rocm/hip-catch-errors.exp: mi_test: conditional catch: breakpoint info
PASS: gdb.rocm/hip-catch-errors.exp: mi_test: conditional catch: set argument
PASS: gdb.rocm/hip-catch-errors.exp: mi_test: conditional catch
PASS: gdb.rocm/hip-catch-errors.exp: mi_test: temporary conditional catch: mi-catch-hiperr
PASS: gdb.rocm/hip-catch-errors.exp: mi_test: temporary conditional catch: breakpoint info
PASS: gdb.rocm/hip-catch-errors.exp: mi_test: temporary conditional catch: set argument
PASS: gdb.rocm/hip-catch-errors.exp: mi_test: temporary conditional catch
```

### DIFF
--- a/CHANGELOG_AMD.md
+++ b/CHANGELOG_AMD.md
@@ -5,6 +5,13 @@ Full documentation for ROCgdb is available at
 
 ## ROCgdb-X for ROCm-next
 
+### Added
+
+- The "catch hiperr" feature is now exposed to MI too, with a new
+  `-catch-hiperr` command and related fields in `*stopped` records.
+  See the "HIP Runtime Error" subsection of the "GDB/MI Catchpoint
+  Commands" section in the ROCgdb manual.
+
 ## ROCgdb-16-3 for ROCm-7.14
 
 ### Added
--- a/gdb/break-catch-hiperr.c
+++ b/gdb/break-catch-hiperr.c
@@ -148,26 +148,26 @@ hiperr_frame_p (frame_info_ptr frame)
 }
 
 /* Format the HIP error info as a string (with newline) for printing.
-   Returns an empty string if the parameters cannot be fetched.  */
+   Returns an empty string if the information cannot be fetched.  */
 
 static std::string
 hiperr_info_string (struct gdbarch *gdbarch)
 {
   frame_info_ptr frame = get_selected_frame (_("No frame selected"));
 
-  if (!hiperr_frame_p (frame) || !gdbarch_fetch_hiperr_parameters_p (gdbarch))
+  if (!hiperr_frame_p (frame) || !gdbarch_fetch_hiperr_info_p (gdbarch))
     return std::string ();
 
-  std::optional<hiperr_parameters> err_params
-    = gdbarch_fetch_hiperr_parameters (gdbarch, frame);
+  std::optional<hiperr_info> err_info
+    = gdbarch_fetch_hiperr_info (gdbarch, frame);
 
-  if (!err_params.has_value ())
+  if (!err_info.has_value ())
     return std::string ();
 
   return string_printf (_("HIP API call failed with error %s (%u): %s\n"),
-			err_params->err_name.get (),
-			err_params->err_no,
-			err_params->err_str.get ());
+			err_info->err_name.get (),
+			err_info->err_no,
+			err_info->err_str.get ());
 }
 
 /* Message printed when the breakpoint is hit.  */
@@ -188,11 +188,11 @@ hiperr_catchpoint::print_it (const bpstat *bs) const
   print_num_locno (bs, uiout);
   uiout->text (" (HIP error)\n");
 
-  const std::string hiperr_info
+  const std::string hiperr_info_str
     = hiperr_info_string (bs->bp_location_at->gdbarch);
-  if (!hiperr_info.empty ())
+  if (!hiperr_info_str.empty ())
     {
-      uiout->text (hiperr_info.c_str ());
+      uiout->text (hiperr_info_str.c_str ());
       uiout->text
 	("\nThe $_hiperr convenience variable holds the error number.\n");
     }
@@ -244,13 +244,13 @@ compute_hiperr (struct gdbarch *gdbarch,
 {
   frame_info_ptr frame = get_selected_frame (_("No frame selected"));
 
-  if (!hiperr_frame_p (frame) || !gdbarch_fetch_hiperr_parameters_p (gdbarch))
+  if (!hiperr_frame_p (frame) || !gdbarch_fetch_hiperr_info_p (gdbarch))
     return value::allocate (builtin_type (gdbarch)->builtin_void);
 
-  std::optional<hiperr_parameters> err_params
-    = gdbarch_fetch_hiperr_parameters (gdbarch, frame);
+  std::optional<hiperr_info> err_info
+    = gdbarch_fetch_hiperr_info (gdbarch, frame);
 
-  if (!err_params.has_value ())
+  if (!err_info.has_value ())
     return value::allocate (builtin_type (gdbarch)->builtin_void);
 
   /* If there's a "hipError_t" type in the debug symbols that is
@@ -263,13 +263,13 @@ compute_hiperr (struct gdbarch *gdbarch,
     {
       gdb_byte err_no[4];
       const bfd_endian byte_order = gdbarch_byte_order (gdbarch);
-      store_unsigned_integer (err_no, 4, byte_order, err_params->err_no);
+      store_unsigned_integer (err_no, 4, byte_order, err_info->err_no);
       return value_from_contents (bs.symbol->type (), err_no);
     }
 
   /* Return hiperr as an unsigned int.  */
   return value_from_ulongest (builtin_type (gdbarch)->builtin_uint32,
-			      err_params->err_no);
+			      err_info->err_no);
 }
 
 /* Implementation of the '$_hiperr' variable.  */
--- a/gdb/break-catch-hiperr.c
+++ b/gdb/break-catch-hiperr.c
@@ -36,11 +36,6 @@ struct hiperr_catchpoint : public code_breakpoint
 		     const char *cond_string)
     : code_breakpoint (gdbarch, bp_catchpoint, temp, cond_string)
   {
-    /* Make sure "locspec" is initialized, irrespective of the
-       "__hipOnError ()" symbol being found.  This allows listing the
-       breakpoint as a pending one when the binary is not loaded yet.
-       Otherwise, "info breakpoints" is going to trigger an assert.  */
-    locspec = new_explicit_location_spec_function ("__hipOnError");
     pspace = current_program_space;
     re_set (nullptr);
   }
@@ -71,7 +66,9 @@ hiperr_catchpoint::re_set (program_space * /*ps*/)
 
   try
     {
-      sals = this->decode_location_spec (this->locspec.get (),
+      location_spec_up locspec
+	= new_explicit_location_spec_function ("__hipOnError");
+      sals = this->decode_location_spec (locspec.get (),
 					 current_program_space);
     }
   catch (const gdb_exception_error &ex)
--- a/gdb/break-catch-hiperr.c
+++ b/gdb/break-catch-hiperr.c
@@ -144,29 +144,6 @@ hiperr_frame_p (frame_info_ptr frame)
   return true;
 }
 
-/* Format the HIP error info as a string (with newline) for printing.
-   Returns an empty string if the information cannot be fetched.  */
-
-static std::string
-hiperr_info_string (struct gdbarch *gdbarch)
-{
-  frame_info_ptr frame = get_selected_frame (_("No frame selected"));
-
-  if (!hiperr_frame_p (frame) || !gdbarch_fetch_hiperr_info_p (gdbarch))
-    return std::string ();
-
-  std::optional<hiperr_info> err_info
-    = gdbarch_fetch_hiperr_info (gdbarch, frame);
-
-  if (!err_info.has_value ())
-    return std::string ();
-
-  return string_printf (_("HIP API call failed with error %s (%u): %s\n"),
-			err_info->err_name.get (),
-			err_info->err_no,
-			err_info->err_str.get ());
-}
-
 /* Message printed when the breakpoint is hit.  */
 
 enum print_stop_action
@@ -185,10 +162,19 @@ hiperr_catchpoint::print_it (const bpstat *bs) const
   print_num_locno (bs, uiout);
   uiout->text (" (HIP error)\n");
 
-  const std::string hiperr_info_str
-    = hiperr_info_string (bs->bp_location_at->gdbarch);
-  if (!hiperr_info_str.empty ())
+  std::optional<hiperr_info> err_info;
+  frame_info_ptr frame = get_selected_frame (_("No frame selected"));
+  struct gdbarch *barch = bs->bp_location_at->gdbarch;
+  if (hiperr_frame_p (frame) && gdbarch_fetch_hiperr_info_p (barch))
+    err_info = gdbarch_fetch_hiperr_info (barch, frame);
+
+  if (err_info.has_value ())
     {
+      const std::string hiperr_info_str
+	= string_printf (_("HIP API call failed with error %s (%u): %s\n"),
+			 err_info->err_name.get (),
+			 err_info->err_no,
+			 err_info->err_str.get ());
       uiout->text (hiperr_info_str.c_str ());
       uiout->text
 	("\nThe $_hiperr convenience variable holds the error number.\n");
@@ -201,9 +187,26 @@ hiperr_catchpoint::print_it (const bpstat *bs) const
       uiout->field_string ("reason",
 			   async_reason_lookup (EXEC_ASYNC_BREAKPOINT_HIT));
       uiout->field_string ("disp", bpdisp_text (disposition));
+
+      if (err_info.has_value ())
+	{
+	  uiout->field_unsigned ("hiperr-code", err_info->err_no);
+	  uiout->field_string ("hiperr-name", err_info->err_name.get ());
+	  uiout->field_string ("hiperr-str", err_info->err_str.get ());
+	}
     }
 
   return PRINT_NOTHING;
+}
+
+/* Instantiate a hiperr_catchpoint.  */
+
+void
+add_catch_hiperr (const char *condition, bool tempflag)
+{
+  auto hcp = std::make_unique<hiperr_catchpoint> (get_current_arch (),
+						  tempflag, condition);
+  install_breakpoint (0, std::move (hcp), 1);
 }
 
 /* Implementation of "catch hiperr" command.  */
@@ -224,10 +227,7 @@ catch_hiperr_command (const char *arg,
   if ((*arg != '\0') && !c_isspace (*arg))
     error (_("Junk at the end of arguments."));
 
-  auto hcp = std::make_unique<hiperr_catchpoint> (get_current_arch (),
-						  tempflag, cond_string);
-
-  install_breakpoint (0, std::move (hcp), 1);
+  add_catch_hiperr (cond_string, tempflag);
 }
 
 
--- a/gdb/breakpoint.h
+++ b/gdb/breakpoint.h
@@ -2099,6 +2099,13 @@ extern void catch_exception_event (enum exception_event_kind ex_event,
 				   const char *regex, bool tempflag,
 				   int from_tty);
 
+/* Catch a HIP error according to the input arguments.  If CONDITION is
+   provided, only failures that satisfy it will trigger the catchpoint.
+   If TEMPFLAG is set, the catchpoint is triggered once and then it
+   is removed.  */
+
+extern void add_catch_hiperr (const char *condition, bool tempflag);
+
 /* Return true if BPT is of any hardware watchpoint kind.  */
 
 bool is_hardware_watchpoint (const struct breakpoint *bpt);
--- a/gdb/doc/gdb.texinfo
+++ b/gdb/doc/gdb.texinfo
@@ -36745,6 +36745,7 @@ catchpoints.
 * Shared Library GDB/MI Catchpoint Commands::
 * Ada Exception GDB/MI Catchpoint Commands::
 * C++ Exception GDB/MI Catchpoint Commands::
+* HIP Runtime Error GDB/MI Catchpoint Commands::
 @end menu
 
 @node Shared Library GDB/MI Catchpoint Commands
@@ -37082,6 +37083,74 @@ and @samp{tcatch catch} (@pxref{Set Catchpoints}).
   thread-id="1",stopped-threads="all",core="6"
 (gdb)
 @end smallexample
+
+@node HIP Runtime Error GDB/MI Catchpoint Commands
+@subsection HIP Runtime Error @sc{gdb/mi} Catchpoints
+
+The following @sc{gdb/mi} commands can be used to create catchpoints
+that stop the execution when HIP runtime errors are being returned.
+
+@findex -catch-hiperr
+@subheading The @code{-catch-hiperr} Command
+
+@subsubheading Synopsis
+
+@smallexample
+ -catch-hiperr [ -c @var{condition} ] [ -t ]
+@end smallexample
+
+Stop when the debuggee catches a HIP runtime error.
+
+@table @samp
+@item -c @var{condition}
+Make the catchpoint conditional on @var{condition}.
+@item -t
+Create a temporary catchpoint.
+@end table
+
+@subsubheading @value{GDBN} Command
+
+The corresponding @value{GDBN} commands are @samp{catch hiperr} and
+@samp{tcatch hiperr} (@pxref{catch hiperr}).
+
+@subsubheading Example
+
+@smallexample
+-catch-hiperr -c $_hiperr==hipErrorInvalidDevice -t
+^done,bkpt=@{number="1",type="catchpoint",disp="del",enabled="y",
+  what="catch hiperr",catch-type="hiperr",
+  cond="$_hiperr==hipErrorInvalidDevice",times="0"@}
+(gdb)
+-exec-run
+^running
+(gdb)
+~"\n"
+~"Thread 1 \"hip\" hit Temporary catchpoint 1 (HIP error)\n"
+~"HIP API call failed with error hipErrorInvalidDevice (101):
+  invalid device ordinal\n"
+~"\n"
+~"The $_hiperr convenience variable holds the error number.\n"
+*stopped,bkptno="1",reason="breakpoint-hit",disp="del",
+  hiperr-code="101",hiperr-name="hipErrorInvalidDevice",
+  hiperr-str="invalid device ordinal",thread-id="1",
+  stopped-threads="all",core="0"
+=breakpoint-deleted,id="1"
+@end smallexample
+
+The @samp{*stopped} record is extended with three fields:
+
+@table @samp
+
+@item hiperr-code
+The error number.
+
+@item hiperr-name
+The name of the error.
+
+@item hiperr-str
+The descriptive text about the error.
+
+@end table
 
 @c %%%%%%%%%%%%%%%%%%%%%%%%%%%% SECTION %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 @node GDB/MI Program Context
--- a/gdb/testsuite/gdb.rocm/hip-catch-errors.exp
+++ b/gdb/testsuite/gdb.rocm/hip-catch-errors.exp
@@ -16,20 +16,18 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-# Test catching HIP's API errors:
+# Test catching HIP's API errors.  Test goals are:
 #
-# 1. The executable is loaded and started.
-# 2. The executable is loaded but not started.
-# 3. There's no executable set yet.
-# 4. Two API errors are returned and caught.
-# 5. Implicit (stub) execution in the form of "kernel<<<...>>>();"
-# 6. Conditional catch of 1 error, using $_hiperr, while there are 2 errors.
-# 7. Temporary catch of the first error, while there are 2 errors.
-# 8. $_hiperr is printed correctly when debug symbols are available.
-# 9. $_hiperr is printed correctly without debug symbols.
-#    This one checks if "catch hiperr" works without debug symbols as well.
+# - CLI: check catchpoints, either simple or complex (conditional or
+# temporary) are getting triggered correctly.  Also try without debug
+# symbols.
+#
+# - MI: parsing for the "-catch-hiperr ..." command.
+# - MI: catchpoints are set correctly and reported correctly when
+# triggered.
 
 load_lib rocm.exp
+load_lib mi-support.exp
 
 require allow_hip_tests
 
@@ -77,21 +75,29 @@ proc get_string_valueof {exp default} {
 }
 
 # Get the reference values from the program (which is run with "ref" argument)
+# Return true when all references are initialized; false, otherwise.
 
 proc init_reference_hip_errors {} {
     foreach func {set_device get_device launch_kernel} {
+	set sum 0
+
 	set ::hip_${func}_err_no [get_integer_valueof "hip_${func}_err.no" bad]
-	gdb_assert {[set ::hip_${func}_err_no] != "bad"} \
-	    "hip_${func}_err_no initialized"
+	incr sum [gdb_assert {[set ::hip_${func}_err_no] != "bad"} \
+		      "hip_${func}_err_no initialized"]
 
 	set ::hip_${func}_err_name [get_string_valueof "hip_${func}_err.name" bad]
-	gdb_assert {[set ::hip_${func}_err_name] != "bad"} \
-	    "hip_${func}_err_name initialized"
+	incr sum [gdb_assert {[set ::hip_${func}_err_name] != "bad"} \
+		      "hip_${func}_err_name initialized"]
 
 	set ::hip_${func}_err_str [get_string_valueof "hip_${func}_err.str" bad]
-	gdb_assert {[set ::hip_${func}_err_str] != "bad"} \
-	    "hip_${func}_err_str initialized"
+	incr sum [gdb_assert {[set ::hip_${func}_err_str] != "bad"} \
+		      "hip_${func}_err_str initialized"]
+
+	if {$sum != 3} {
+	    return false
+	}
     }
+    return true
 }
 
 # Set a hiperr catchpoint and check if it set accordingly.  If TEMP is
@@ -187,12 +193,14 @@ proc check_hip_launch_kernel_err {cmd} {
 	$::hip_launch_kernel_err_str
 }
 
-proc do_test {} {
-    with_rocm_gpu_lock {
+# Check for __hipOnError symbol existence and initialize references
+# using the test program.  Return false if anything goes wrong.
 
+proc_with_prefix preparation_test {} {
+    with_rocm_gpu_lock {
 	clean_restart ${::testfile}.dbg
 	if {![runto_main]} {
-	    return
+	    return false
 	}
 	gdb_test_multiple "info symbol __hipOnError" \
 	    "check if __hipOnError symbol exists" {
@@ -201,21 +209,39 @@ proc do_test {} {
 	    }
 	    -re -wrap ".*" {
 		unsupported "no __hipOnError symbol was found"
-		return
+		return false
 	    }
 	}
 
 	# Let all the errors occur and their values be recorded by the program.
-	with_test_prefix "getting reference values" {
+	with_test_prefix "get ref vals" {
 	    clean_restart ${::testfile}.dbg
 	    gdb_test_no_output "set args ref"
 	    if {![runto [gdb_get_line_number \
 			    "Break after reference initialization"]]} {
-		return
+		return false
 	    }
-	    init_reference_hip_errors
+	    set res [init_reference_hip_errors]
 	}
+    }
+    return $res
+}
 
+# Test catching HIP's API errors (through CLI interface):
+#
+# 1. The executable is loaded and started.
+# 2. The executable is loaded but not started.
+# 3. There's no executable set yet.
+# 4. Two API errors are returned and caught.
+# 5. Implicit (stub) execution in the form of "kernel<<<...>>>();"
+# 6. Conditional catch of 1 error, using $_hiperr, while there are 2 errors.
+# 7. Temporary catch of the first error, while there are 2 errors.
+# 8. $_hiperr is printed correctly when debug symbols are available.
+# 9. $_hiperr is printed correctly without debug symbols.
+#    This one checks if "catch hiperr" works without debug symbols as well.
+
+proc_with_prefix cli_test {} {
+    with_rocm_gpu_lock {
 	with_test_prefix "after starting" {
 	    clean_restart ${::testfile}.dbg
 	    gdb_test_no_output "set args one"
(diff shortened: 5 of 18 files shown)


gdb/doc/gdb.texinfo
```diff
@@ -37262,6 +37262,54 @@ and @samp{tcatch catch} (@pxref{Set Catchpoints}).
...
+
+@subsubheading @value{GDBN} Command
+
+The corresponding @value{GDBN} commands are @samp{catch hiperr} and
+@samp{tcatch hiperr} (@pxref{catch hiperr}).
+
+@subsubheading Example
+
+@smallexample
+-catch-hiperr -c $_hiperr==hipErrorInvalidDevice -t
```
lancesix, line 37119: usual parsing question…  How do you differentiate `-t` as a flag after the condition v.s. `-t` being the unari `-` operator, i.e. `hipErrorInvalidDevice - t`?

gdb/testsuite/gdb.rocm/mi-catch-hiperr.exp
```diff
@@ -0,0 +1,159 @@
...
+#
+# Only check the MI output and if the breakpoints are inserted correctly.
+# For checking the functionality of "catch hiperr" there's another test
+# called hip-catch-errors.exp.
+
+load_lib rocm.exp
+
+require allow_hip_tests
+
+standard_testfile hip-catch-errors.cpp
```
lancesix: I think we usually try to avoid having tests share the same source.  Things are easier to evolve if we have one source per test.

That being said, the test can have two phases, one testing the cli interpreter, and one testing the mi interpreter.

gdb/testsuite/gdb.rocm/mi-catch-hiperr.exp
```diff
@@ -0,0 +1,159 @@
...
+# 1. -catch-hiperr
+# 2. -catch-hiperr -t
+# 3. -catch-hiperr -c $_hiperr==123
+# 4. -catch-hiperr -c $_hiperr==321 -t
+#
+# Only check the MI output and if the breakpoints are inserted correctly.
+# For checking the functionality of "catch hiperr" there's another test
+# called hip-catch-errors.exp.
+
+load_lib rocm.exp
```
lancesix: To test MI, you most probably want to have

```
load_lib mi-support.exp
```

and use `mi_gdb_test`

gdb/breakpoint.h
```diff
@@ -2099,6 +2099,12 @@ extern void catch_exception_event (enum exception_event_kind ex_event,
 				   const char *regex, bool tempflag,
 				   int from_tty);
 
+/* Handle "catch hiperr" MI command.  If CONDITION is provided, only
+   failures that satisfy it will trigger the catchpoint.  If TEMPFLAG
+   is set, the catchpoint is triggered once and then it is removed.  */
+
+extern void catch_hiperr_mi (const char *condition, bool tempflag);
```
lancesix: is `breakpoint.h` the best place to define this?  Feels out of place.

lancesix (commented):
First very quick remarks.

gdb/breakpoint.h
amd-shahab: I've followed what other `catch`es (eg. `(un)load`,  `(re)throw|catch`) have done:

```
gdb/breakpoint.h
----------------
extern void add_solib_catchpoint (...);   /* called by mi_cmd_catch_(un)load ()  */
extern void catch_exception_event (...);  /* called by mi_cmd_catch_exception_event ()  */
```

It is worth noting that these interfaces are only externally used by the `mi_cmd`s. So they're solely added in `breakpoint.h` to be used by the MI.

I'm going to mark this as "resolved", feel free to change it back.

gdb/testsuite/gdb.rocm/mi-catch-hiperr.exp
amd-shahab: This test does not use any source program. It only exercises the part of MI that turns the command into a breakpoint. That's why the test is not guarded by `with_rocm_gpu_lock { ... }`.

At first I removed the
```
require allow_hip_tests

standard_testfile hip-catch-errors.cpp

if {[prepare_for_testing "failed to prepare ${testfile}" \
	$testfile $srcfile {debug hip}]} {
    return -1
}
```

but then I was getting errors probably due to some global variables not being set. I did not look into that, but maybe it's time I take another stab at it.

gdb/doc/gdb.texinfo
amd-shahab: Debugging the code around the option parsing [1], reveals that the valid ways of passing arguments to options (in this example, _condition_ to `-c`) is to **not use any spaces in the argument** or **have them double-quoted**. For instance, see the two alternatives below and how they end up setting the parameters:

```
# alt.1: no spaces
interpreter-exec mi '-catch-hiperr -c $_hiperr=hipErrorInvalidDevice-t -t'

mi_cmd_catch_hiperr (..., const char *const argv, int argc)
  argc: 3
  argv[0]: "-c"
  argv[1]: "$_hiperr==hipErrorInvalidDevice-t"
  argv[2]: "-t"
  .
  .
  .
  condition = "$_hiperr==hipErrorInvalidDevice-t"
  temp = true

---------------------------------------------------
# alt.2: double-quoted
interpreter-exec mi '-catch-hiperr -c "$_hiperr == hipErrorInvalidDevice - t" -t'

mi_cmd_catch_hiperr (..., const char *const argv, int argc)
  argc: 3
  argv[0]: "-c"
  argv[1]: "$_hiperr == hipErrorInvalidDevice - t"
  argv[2]: "-t"
  .
  .
  .
  condition = "$_hiperr == hipErrorInvalidDevice - t"
  temp = true
```

Any deviation from this and we'll end up with the wrong `argc/argv` to begin with:
```
interpreter-exec mi '-catch-hiperr -c $_hiperr == hipErrorInvalidDevice - t -t'

mi_cmd_catch_hiperr (..., const char *const argv, int argc)
  argc: 7
  argv[0]: "-c"
  argv[1]: "$_hiperr"
  argv[2]: "=="
  argv[3]: "hipErrorInvalidDevice"
  argv[4]: "-"
  argv[5]: "t"
  argv[6]: "-t"
  .
  .
  .
  condition = "$_hiperr"
  then "opt" in the for-loop [1] becomes negative and the loop is broken
  and we'll end up with a conditional catchpoint with $_hiperr as its condition.
```

[1] `gdb/mi/mi-cmd-catch.c`
```c
void mi_cmd_catch_hiperr (const char *cmd, const char *const *argv, int argc)
{
  ...
  for (;;)
    {
      int opt = mi_getopt (cmd, argc, argv, opts, &oind, &oarg);

      if (opt < 0)
        break;

      switch ((enum opt) opt)
        {
        case OPT_CONDITION:
          condition = oarg;
          break;
        case OPT_TEMP:
          temp = true;
          break;
        }
    }
  ...
}
```
For the record, I've added `-catch-hiperr -c "$_hiperr == hipErrorInvalidDevice -t" -t` as a test now.

gdb/testsuite/gdb.rocm/mi-catch-hiperr.exp
amd-shahab: This test is a mixture of MI commands and normal commands. For `mi_gdb_test` to work, I have to do a `mi_clean_restart`. That will put GDB in MI mode and normal tests like
```
gdb_test "info break" "regex_output" "reason"
```
must adjust their `regex_output` to reflect that.

This is `info break` output without `mi_clean_restart`
```
info break
Num     Type           Disp Enb Address    What
1       catchpoint     keep y              catch hiperr
```

and this `info break` output with `mi_clean_restart`:
```
info break
&"info break\n"
~"Num     Type           Disp Enb Address    What\n"
~"1       catchpoint     keep y              catch hiperr\n"
^done
```

I don't immediately see the benefit of switching to `mi_gdb_test` then. One could argue that "maybe the MI version of 'info break' should be used", but the goal here is not to go full fledged MI.

amd-shahab:
- Use `-catch-hiperr -c "$_hiperr == hipErrorInvalidDevice -t" -t` as temporary conditional test
- Use `mi_gdb_test`s instead of `gdb_test` and `gdb_test_multiple`
- get rid of the _rocm boiler plates_ (`load_lib rocm.exp`, `require allow_hip_tests`, and the `standard_testfile ...`/`prepare_for_testing ...`)

gdb/testsuite/gdb.rocm/mi-catch-hiperr.exp
amd-shahab: I've decided to turn the whole test into an MI based test. I think now the test has better structure due to reusable code that evolved around it.

gdb/testsuite/gdb.rocm/mi-catch-hiperr.exp
amd-shahab: Now that I [load mi-support.exp](https://github.com/ROCm/ROCgdb/pull/180#discussion_r3454862328), I could easily get rid of the boilerplates:

```
-load_lib rocm.exp
+load_lib mi-support.exp

-require allow_hip_tests
-
-standard_testfile hip-catch-errors.cpp

-if {[prepare_for_testing "failed to prepare ${testfile}" \
-       $testfile $srcfile {debug hip}]} {
-    return -1
```
The test is much faster too.

gdb/break-catch-hiperr.c
```diff
@@ -209,6 +209,17 @@ hiperr_catchpoint::print_it (const bpstat *bs) const
   return PRINT_NOTHING;
 }
 
+/* Instantiate a hiperr_catchpoint as the base of the two fuctions:
```
palves: fuctions => functions

gdb/break-catch-hiperr.c
```diff
@@ -227,10 +238,15 @@ catch_hiperr_command (const char *arg,
...
+  catch_hiperr_core (cond_string, tempflag);
+}
 
-  install_breakpoint (0, std::move (hcp), 1);
+/* A wrapper around "catch_hiperr_core ()" for handling MI.  */
+
+void
+catch_hiperr_mi (const char *condition, bool tempflag)
+{
+  catch_hiperr_core (condition, tempflag);
```
palves: catch_hiperr_mi adds nothing to catch_hiperr_core, it's just a pure wrapper with the same interface.  Why have both?

Better delete catch_hiperr_mi and rename catch_hiperr_core to install_hiperr_catchpoint or set_catch_hiperr or add_catch_hiperr or something like that, and use that from both CLI and MI.

gdb/testsuite/gdb.rocm/mi-catch-hiperr.exp
```diff
@@ -0,0 +1,146 @@
...
+# -break-info BP_NUM
+#
+# ...bkpt={number="BP_NUM",type="catchpoint",disp="DISP",enabled="y",
+#	   what="catch hiperr",catch-type="hiperr",cond="COND",
+#	   times="[0-9]+",original-location="-function __hipOnError"}...
+#
+# The 'cond=...' field is added to the expected output only if the
+# provided COND is not an empty string.
+proc chk_mi_bkpt {bp_num disp cond msg} {
+    set bpkt ""
```
palves: Typo: "set bpkt" => "set bkpt"

gdb/testsuite/gdb.rocm/mi-catch-hiperr.exp
```diff
@@ -0,0 +1,146 @@
...
+    set disp "keep"
+    set cond "\$_hiperr == 123"
+    chk_mi_catch_hiperr "-catch-hiperr -c \"$cond\"" \
+	$bp_num $disp [string_to_regexp "$cond"] "$test"
+    chk_mi_bkpt $bp_num $disp [string_to_regexp $cond] "breakpoint info of $test"
+
+    # (4)
+    # -catch-hiperr -c "$_hiperr == hipErrorInvalidDevice -t" -t
+    # ^done,bkpt={number="4",...,disp="del",...,
+    #		  cond="$_hiperr == hipErrorInvalidDeivce -t",...}
```
palves: Typo: hipErrorInvalidDeivce => hipErrorInvalidDevice

gdb/doc/gdb.texinfo
```diff
@@ -37262,6 +37262,54 @@ and @samp{tcatch catch} (@pxref{Set Catchpoints}).
 (gdb)
 @end smallexample
 
+@findex -catch-hiperr
+@subheading The @code{-catch-hiperr} Command
+
+@subsubheading Synopsis
+
+@smallexample
+ -catch-hiperr [ -c @var{condition}] [ -t ]
```
palves: Missing rhs space in `"[ -c @var{condition}]"`

gdb/testsuite/gdb.rocm/mi-catch-hiperr.exp
```diff
@@ -0,0 +1,146 @@
...
+# Test different combinations of "-catch-hiperr [ -c <condition>] [ -t ]"
+#
+# 1. -catch-hiperr
+# 2. -catch-hiperr -t
+# 3. -catch-hiperr -c $_hiperr==123
+# 4. -catch-hiperr -c $_hiperr==321 -t
+#
+# Only check the MI output and if the breakpoints are inserted correctly.
+# For checking the functionality of "catch hiperr" there's another test
+# called hip-catch-errors.exp.
```
palves: We should really test the async "*stopped" notification too (mi_expect_stop), at least.

Negative tests for stuff like "-catch-hiperr -c" with no arg, bogus option, junk, etc. I'd think useful too.

gdb/testsuite/gdb.rocm/mi-catch-hiperr.exp
```diff
@@ -0,0 +1,146 @@
...
+#
+# -catch-hiperr [-c COND] [-t]
+#
+# ^done,bkpt={number="BP_NUM",type="catchpoint",disp="DISP",enabled="y",
+#	      what="catch hiperr",catch-type="hiperr",cond="COND",
+#	      times="[0-9]+",original-location="-function __hipOnError"}
+#
+# The 'cond=...' field is added to the expected output only if the
+# provided COND is not an empty string.
+proc chk_mi_catch_hiperr {cmd bp_num disp cond msg} {
```
palves: This and chk_mi_bkpt build the same attribute list.  You should factor that to a hiperr_bkpt_re or some such proc, then used by both chk_mi_catch_hiperr and chk_mi_bkpt.  Or see about reusing mi_create_breakpoint and mi_make_breakpoint.

gdb/testsuite/gdb.rocm/mi-catch-hiperr.exp
```diff
@@ -0,0 +1,146 @@
...
+    # -break-info 4
+    # ...bkpt={number="4",...,disp="del",...,
+    #	       cond="$_hiperr == hipErrorInvalidDevice -t",...}
+    set bp_num "4"
+    set test "temporary conditional catch"
+    set disp "del"
+    set cond "\$_hiperr == hipErrorInvalidDevice -t"
+    chk_mi_catch_hiperr "-catch-hiperr -c \"$cond\" -t" \
+	$bp_num $disp [string_to_regexp "$cond"] "$test"
+    chk_mi_bkpt $bp_num $disp [string_to_regexp $cond] "breakpoint info of $test"
```
palves: The 4 cases are very similar and could be driven by a table.  Like:

```
    # Each element is {NAME OPTIONS DISP COND}, where OPTIONS are the
    # extra "-catch-hiperr" options to test, DISP is the expected
    # disposition ("keep" or "del"), and COND is the expected
    # condition ("" for none).  Note that in the last case the "-t"
    # embedded in the -c argument is part of the condition, while the
    # trailing -t is what makes the catchpoint temporary.
    set tests {
	{ "simple catch"
	    ""
	    keep
	    "" }

	{ "temporary catch"
	    "-t"
	    del
	    "" }

	{ "conditional catch"
	    {-c "$_hiperr == 123"}
	    keep
	    {$_hiperr == 123} }

	{ "temporary conditional catch"
	    {-c "$_hiperr == hipErrorInvalidDevice -t" -t}
	    del
	    {$_hiperr == hipErrorInvalidDevice -t} }
    }

    set bp_num 0
    foreach t $tests {
	lassign $t name options disp cond
	incr bp_num

	set cond_re [expr {$cond eq "" ? "" : [string_to_regexp $cond]}]
	chk_mi_catch_hiperr "-catch-hiperr $options" \
	    $bp_num $disp $cond_re $name
	chk_mi_bkpt $bp_num $disp $cond_re "breakpoint info of $name"
    }
```

gdb/doc/gdb.texinfo
```diff
@@ -37262,6 +37262,54 @@ and @samp{tcatch catch} (@pxref{Set Catchpoints}).
...
+^running
+(gdb)
+~"\n"
+~"Thread 1 \"hip\" hit Temporary catchpoint 1 (HIP error)\n"
+~"HIP API call failed with error hipErrorInvalidDevice (101):
+  invalid device ordinal\n"
+~"\n"
+~"The $_hiperr convenience variable holds the error number.\n"
+*stopped,bkptno="1",reason="breakpoint-hit",disp="del",thread-id="1",
+  stopped-threads="all",core="0"
```
palves, line 37136: If this is really what MI outputs when the catchpoint triggers, then I'm surprised it doesn't contain attributes for the error caught.  We shouldn't expect that the MI frontend is parsing the CLI ~"HIP API call failed with error ..." string.

palves: To be clear, what I'm thinking of is something like in print_it, when is_mi_like_p(), emit structured fields, something like:

```
  uiout->field_unsigned ("hiperr-code", err_params->err_no);
  uiout->field_string ("hiperr-name", err_params->err_name.get ());
  uiout->field_string ("hiperr-text", err_params->err_str.get ());
```

palves: (strawman names, I didn't think to check what are the best attribute names.)

amd-shahab:
- doc: escape curly braces from MI output with `@`.
- doc: add right-hand-side space for the condition part.
- code: add `hiperr-{code,name,text}` fields for the MI output.
- doc: update to reflect `hiperr-{code,name,text}` fields.

amd-shahab:
- code: get rid of `catch_hiperr_mi()` and use `add_catch_hiperr()` (former `catch_hiperr_core()`)
- test: re-implement the existing 4 testcases with a table an a for-loop.
- test: consolidate `chk_mi_catch_hiperr` and `chk_mi_bkpt` procedures into `hiperr_bkpt_re`.

amd-shahab:
- test: check the output of a triggered catchpoint (specially `hiperr-{code,name}`)
- test: check the parsing of MI command by throwing some tricky and malformed commands

lumachad:
Since we're trying to have a clean gdb.rocm testsuite with non-O0 optimization level, please check if this new test behaves well when executed with O1~O3 and -flto.

amd-shahab:
- test: consolidate MI tests into `hip-catch-errors.exp`. Since now the test is looking for a triggered catchpoint, checking for the existence of the `__hipOnError` and initializing reference values have become necessary. this was already done in `hip-catch-errors.exp`.

amd-shahab:
- add another patch that makes the test still pass under optimizations. now works with these flags:

```
-O0
-O1
-O2
-O3

-O0 -flto
-O1 -flto
-O2 -flto
-O3 -flto
```

gdb/testsuite/gdb.rocm/rocm-test-utils.h
```diff
@@ -54,4 +54,9 @@
 
 #define NOP(N) asm volatile ("s_nop " #N)
 
+/* Insert x86 "nop" into the code.  It's useful for adding pieces of code
+   that we can set breakpoint on without getting optimized away.  */
+
+#define X86_NOP asm volatile ("nop")
```
palves: I'd think calling this something like HOST_NOP instead would be more future-proof.

"set breakpoint on" => "set a breakpoint on"

gdb/testsuite/gdb.rocm/hip-catch-errors.exp
```diff
@@ -187,12 +193,15 @@ proc check_hip_launch_kernel_err {cmd} {
 	$::hip_launch_kernel_err_str
 }
 
-proc do_test {} {
-    with_rocm_gpu_lock {
+# - Check for __hipOnError symbol existence
+# - Initialize references using the test program
+# return false if anything goes wrong
```
palves: This isn't using the usual style.  Missing periods.  Uppercase "return".

gdb/doc/gdb.texinfo
```diff
@@ -37262,6 +37262,56 @@ and @samp{tcatch catch} (@pxref{Set Catchpoints}).
...
+@subsubheading Example
+
+@smallexample
+-catch-hiperr -c $_hiperr==hipErrorInvalidDevice -t
+(gdb)
+-catch-hiperr -c $_hiperr==hipErrorInvalidDevice -t
+^done,bkpt=@{number="1",type="catchpoint",disp="del",enabled="y",
+  what="catch hiperr",catch-type="hiperr",
+  cond="$_hiperr==hipErrorInvalidDevice",times="0",
+  original-location="-function __hipOnError"@}
```
palves: Why does this show original-location, when -catch-throw does not?

```
-catch-throw 
^done,bkpt={
number="4",type="catchpoint",
disp="keep",enabled="y",
what="exception throw",catch-type="throw",
thread-groups=["i1"],times="0"
}
```

gdb/testsuite/gdb.rocm/hip-catch-errors.exp
```diff
@@ -327,6 +354,188 @@ proc do_test {} {
...
+	      {-c "$_hiperr == hipErrorInvalidDevice" -t}
+	      del
+	      {$_hiperr == hipErrorInvalidDevice} }
+	}
+
+	# The relevant part of MI stop after catching a hiperr.
+	append hiperr_stop_re "\\*stopped,bkptno=\"1\","
+	append hiperr_stop_re "reason=\"breakpoint-hit\".*hiperr-code=\"$::hip_set_device_err_no\","
+	append hiperr_stop_re "hiperr-name=\"$::hip_set_device_err_name\","
+	append hiperr_stop_re "hiperr-text=\"$::hip_set_device_err_str\".*"
```
palves: It's curious that we're settling on:

  "hiperr-text" => MI
   "$::hip_set_device_err_str" => tcl variable
    "err_str" => inside gdb (err_params->err_str)

It'd be nice if we used the same nomenclature (text vs str) throughout.

Note that for the MI attribute, whatever we settle on now, we'll be stuck with it.

palves (commented):
Sending some comments now.  I didn't manage to get to the end of the patch, so I'll continue later.

gdb/testsuite/gdb.rocm/hip-catch-errors.exp
amd-shahab: I like your proposed "text" better, but I've used "str" every other place including the comment section in CLR's header file. I guess I'll fallback to `hiperr-str` then :/

palves: The CLR patch does:

      hipGetErrorName(hip::tls.last_command_error_)
      hipGetErrorString(hip::tls.last_command_error_)

Given the HIP API call is hipGetErrorString, "str"/"string" seems to be the better fit.  This way, we're not inventing anything, we're just using the same terminology HIP itself uses.

gdb/doc/gdb.texinfo
```diff
@@ -37057,6 +37262,56 @@ and @samp{tcatch catch} (@pxref{Set Catchpoints}).
...
+
+The corresponding @value{GDBN} commands are @samp{catch hiperr} and
+@samp{tcatch hiperr} (@pxref{catch hiperr}).
+
+@subsubheading Example
+
+@smallexample
+-catch-hiperr -c $_hiperr==hipErrorInvalidDevice -t
+(gdb)
+-catch-hiperr -c $_hiperr==hipErrorInvalidDevice -t
```
palves: Duplicated command line.

The first command line + (gdb) looks like spurious leftovers.  Note that the -catch-throw and -catch-assert examples show the command once, then ^done.

gdb/testsuite/gdb.rocm/hip-catch-errors.exp
```diff
@@ -327,6 +354,188 @@ proc do_test {} {
...
+	append bkpt "cond=\"$cond\","
+    }
+    append bkpt "times=\"0\",original-location=\"-function __hipOnError\""
+    return $bkpt
+}
+
+
+# Check if the following commands are handled correctly:
+#
+# -catch-hiperr -c "$_hiperr == hipErrorInvalidDeice -t" -t
```
palves: Typo: hipErrorInvalidDeice => hipErrorInvalidDevice

gdb/breakpoint.h
```diff
@@ -2099,6 +2099,13 @@ extern void catch_exception_event (enum exception_event_kind ex_event,
 				   const char *regex, bool tempflag,
 				   int from_tty);
 
+/* Catch a HIP error according to the input arguments.  If CONDITION is
+   provided, only failures that satisfy it will trigger the catchpoint.
+   if TEMPFLAG is set, the catchpoint is triggered once and then it
```
palves: Uppercase "if".

gdb/testsuite/gdb.rocm/hip-catch-errors.exp
```diff
@@ -327,6 +354,188 @@ proc do_test {} {
...
+	  ""
+	  "Option -c requires an argument" }
+
+	{ "unknown option"
+	  "-abc"
+	  keep
+	  ""
+	  "Unknown option \\\"abc\\\"" }
+
+	{ "dangling texts"
```
palves: Plural "texts" is not very typical/idiomatic.  Suggest either "dangling text" or "trailing junk".

gdb/testsuite/gdb.rocm/hip-catch-errors.exp
```diff
@@ -327,6 +354,188 @@ proc do_test {} {
...
+	  "-t one two three"
+	  del
+	  ""
+	  "" }
```
palves, line 423: So we just ignore dangling junk.  Bah, I see other catchpoints, like -catch-throw do the same:

```
-catch-throw foo bar 
^done,bkpt={number="6",type="catchpoint",disp="keep",enabled="y",what="exception throw",catch-type="throw",thread-groups=["i1"],times="0"}
(gdb) 
```

Seems like a bug, but guess that's OK given the precedent...

palves (commented):
In patch 1's commit log:

"at then end" => "at the end"

This warrants a CHANGELOG_AMD.md entry.  And gdb/NEWS, for upstream, but that can do done later when we actually upstream this, based on CHANGELOG_AMD.md if you prefer.

palves:
> Sending some comments now. I didn't manage to get to the end of the patch, so I'll continue later.

FYI: I'm done for now.

gdb/doc/gdb.texinfo
amd-shahab: Because `hiperr_catchpoint` uses explicit function name to define the location and `decode_location_spec ()`. Please also see the comment in the snippet below:

```c++
struct hiperr_catchpoint : public code_breakpoint
{
  hiperr_catchpoint (struct gdbarch *gdbarch, bool temp,
		     const char *cond_string)
    : code_breakpoint (gdbarch, bp_catchpoint, temp, cond_string)
  {
    /* Make sure "locspec" is initialized, irrespective of the
       "__hipOnError ()" symbol being found.  This allows listing the
       breakpoint as a pending one when the binary is not loaded yet.
       Otherwise, "info breakpoints" is going to trigger an assert.  */
    locspec = new_explicit_location_spec_function ("__hipOnError");
    pspace = current_program_space;
    re_set (nullptr);
  }
  ...
};

void
hiperr_catchpoint::re_set (program_space * /*ps*/)
{
  std::vector<symtab_and_line> sals;
  try
    {
      sals = this->decode_location_spec (this->locspec.get (),
					                     current_program_space);
    }
  catch (const gdb_exception_error &ex)
    {
      if (ex.error != NOT_FOUND_ERROR)
	    throw;
    }
  update_breakpoint_locations (this, current_program_space, sals, {});
}
```

while `exception_catchpoint` uses "probes" as the first resource which is a success most of the time:

```c++
void
exception_catchpoint::re_set (program_space *pspace)
{
  std::vector<symtab_and_line> sals;
  struct program_space *filter_pspace = current_program_space;

  /* We first try to use the probe interface.  */
  try
    {
      location_spec_up locspec
	    = new_probe_location_spec (exception_functions[kind].probe);
      sals = parse_probes (locspec.get (), filter_pspace, NULL);
    }
  catch (const gdb_exception_error &e)
    {
      /* Using the probe interface failed.  Let's fallback to the normal
	 catchpoint mode.  */
      try
	    {
	      location_spec_up locspec
	        = (new_explicit_location_spec_function
	           (exception_functions[kind].function));
	      sals = this->decode_location_spec (locspec.get (), filter_pspace);
	    }
      catch (const gdb_exception_error &ex)
	    {
	      if (ex.error != NOT_FOUND_ERROR)
	        throw;
	    }
    }
  update_breakpoint_locations (this, filter_pspace, sals, {});
}
```

amd-shahab:
> This warrants a CHANGELOG_AMD.md entry. And gdb/NEWS, for upstream, but that can do done later when we actually upstream this, based on CHANGELOG_AMD.md if you prefer.

Added:
```
- "catch hiperr" MI command: `-catch-hiperr`.  See ROCgdb manual, the
  "HIP Runtime Error" subsection in Catchpoints.
```

gdb/testsuite/gdb.rocm/hip-catch-errors.exp
amd-shahab: I'm going to mark this as resolved if you don't mind. Moreover, I can make the check stricter if you want.

gdb/testsuite/gdb.rocm/hip-catch-errors.exp
amd-shahab: everything is `hiperr-str` now.

amd-shahab:
- doc: remove left-over lines from -catch-hiperr example run.
- mi: use "hiperr-str" instead of "hiperr-text" (code, doc, test).
- test: `X86_NOP` -> `HOST_NOP`.

bunch of text adjustments in the comments:
- code: in `breakpoint.h`: "if TEMPFLAG" -> "If TEMPFLAG".
- test: "hipErrorInvalidDeice" -> "hipErrorInvalidDevice".
- test: "dangling texts" -> "trailing junk".
- test: better wording for "preparation_test" proc.
- test: "set breakpoint on" -> "set a breakpoint on".

amd-shahab:
- rebase to get the latest `CHANGELOG_AMD.md`.

amd-shahab:
- doc: add entry to `CHANGELOG_AMD.md`.
- doc: add a "HIP Runtime Error" subsection to Catchpoints.

gdb/doc/gdb.texinfo
palves: Here's how original-location is documented:

```
@item original-location
The location of the breakpoint as originally specified by the user.
This field is optional.
```

So it seems wrong to show that internal function here, as that's not what the user specified.


>    /* Make sure "locspec" is initialized, irrespective of the
>      "__hipOnError ()" symbol being found.  This allows listing the
>       breakpoint as a pending one when the binary is not loaded yet.
>       Otherwise, "info breakpoints" is going to trigger an assert.  */

...

> while exception_catchpoint uses "probes" as the first resource which is a success most of the time:

Well, that would suggest that if the probes don't work -- which they won't on non-Linux platforms, like Windows -- then we'd be seeing that crash there, no?

I hacked away the probes path, forcing gdb to fallback to the function path, and I couldn't reproduce said crash there.

And then I make the catch hiperr code not save the locspec like catch throw does, and things still worked, no crash, except original-location is gone.

See this patch below.  Your new tests all pass cleanly still, with the tweak to not expect original-location.
Am I missing something?

```
From b47f7ebd6be5dbc34ae1f7a97cda91e27212970f Mon Sep 17 00:00:00 2001
From: Pedro Alves <pedro@palves.net>
Date: Thu, 6 Aug 2026 13:11:46 +0100
Subject: [PATCH] no original-location

Change-Id: I116118e64ee4a3b7ca74c35d6f2f6ca8ee10f908
---
 gdb/break-catch-hiperr.c                    | 7 ++++++-
 gdb/break-catch-throw.c                     | 2 ++
 gdb/testsuite/gdb.rocm/hip-catch-errors.exp | 4 ++--
 3 files changed, 10 insertions(+), 3 deletions(-)

diff --git a/gdb/break-catch-hiperr.c b/gdb/break-catch-hiperr.c
index 2285d5776f0..c4533824564 100644
--- a/gdb/break-catch-hiperr.c
+++ b/gdb/break-catch-hiperr.c
@@ -36,11 +36,13 @@ struct hiperr_catchpoint : public code_breakpoint
 		     const char *cond_string)
     : code_breakpoint (gdbarch, bp_catchpoint, temp, cond_string)
   {
+#if 0
     /* Make sure "locspec" is initialized, irrespective of the
        "__hipOnError ()" symbol being found.  This allows listing the
        breakpoint as a pending one when the binary is not loaded yet.
        Otherwise, "info breakpoints" is going to trigger an assert.  */
     locspec = new_explicit_location_spec_function ("__hipOnError");
+#endif
     pspace = current_program_space;
     re_set (nullptr);
   }
@@ -71,7 +73,10 @@ hiperr_catchpoint::re_set (program_space * /*ps*/)
 
   try
     {
-      sals = this->decode_location_spec (this->locspec.get (),
+      location_spec_up locspec
+	= new_explicit_location_spec_function ("__hipOnError");
+
+      sals = this->decode_location_spec (locspec.get (),
 					 current_program_space);
     }
   catch (const gdb_exception_error &ex)
diff --git a/gdb/break-catch-throw.c b/gdb/break-catch-throw.c
index 03d8c5f772d..e307cb99389 100644
--- a/gdb/break-catch-throw.c
+++ b/gdb/break-catch-throw.c
@@ -198,6 +198,7 @@ exception_catchpoint::re_set (program_space *pspace)
   struct program_space *filter_pspace = current_program_space;
 
   /* We first try to use the probe interface.  */
+#if 0
   try
     {
       location_spec_up locspec
@@ -205,6 +206,7 @@ exception_catchpoint::re_set (program_space *pspace)
       sals = parse_probes (locspec.get (), filter_pspace, NULL);
     }
   catch (const gdb_exception_error &e)
+#endif
     {
       /* Using the probe interface failed.  Let's fallback to the normal
 	 catchpoint mode.  */
diff --git a/gdb/testsuite/gdb.rocm/hip-catch-errors.exp b/gdb/testsuite/gdb.rocm/hip-catch-errors.exp
index 16a5f26f8ec..dbf3321db30 100644
--- a/gdb/testsuite/gdb.rocm/hip-catch-errors.exp
+++ b/gdb/testsuite/gdb.rocm/hip-catch-errors.exp
@@ -360,7 +360,7 @@ proc_with_prefix cli_test {} {
 #
 #   number="BP_NUM",type="catchpoint",disp="DISP",enabled="y",
 #   what="catch hiperr",catch-type="hiperr",cond="COND",
-#   times="0",original-location="-function __hipOnError"
+#   times="0"
 #
 # The 'cond=...' field is added only if the provided COND is
 # not an empty string.
@@ -371,7 +371,7 @@ proc hiperr_bkpt_re {bp_num disp cond} {
     if {$cond != ""} {
 	append bkpt "cond=\"$cond\","
     }
-    append bkpt "times=\"0\",original-location=\"-function __hipOnError\""
+    append bkpt "times=\"0\""
     return $bkpt
 }
 

base-commit: 8bd711f9852af7ae045d87487790e1418e1b7182
-- 
2.54.0
```

CHANGELOG_AMD.md
```diff
@@ -5,6 +5,11 @@ Full documentation for ROCgdb is available at
 
 ## ROCgdb-X for ROCm-next
 
+### Added
+
+- "catch hiperr" MI command: `-catch-hiperr`.  See ROCgdb manual, the
+  "HIP Runtime Error" subsection in Catchpoints.
+
```
palves, line 14: How about:

```
- The "catch hiperr" feature is now exposed to MI too, with a new
  `-catch-hiperr` command and related attributes in `*stopped`
  records.  See the "HIP Runtime Error" subsection of the "GDB/MI
  Catchpoint Commands" section in the ROCgdb manual.
```

Makes clear that there's more than just the command.  Also "in Catchpoints" was a bit too vague.

gdb/doc/gdb.texinfo
```diff
@@ -37083,6 +37083,60 @@ and @samp{tcatch catch} (@pxref{Set Catchpoints}).
...
+(gdb)
+~"\n"
+~"Thread 1 \"hip\" hit Temporary catchpoint 1 (HIP error)\n"
+~"HIP API call failed with error hipErrorInvalidDevice (101):
+  invalid device ordinal\n"
+~"\n"
+~"The $_hiperr convenience variable holds the error number.\n"
+*stopped,bkptno="1",reason="breakpoint-hit",disp="del",
+  hiperr-code="101",hiperr-name="hipErrorInvalidDevice",
+  hiperr-str="invalid device ordinal",thread-id="1",
```
palves, line 37135: `hiperr-code`, `hiperr-name`, and `hiperr-str` appear here in the example, but they should also be properly documented.

amd-shahab:
- doc: update `CHANGELOG_AMD.md` according to Pedro's suggestion.
- doc: mention `hiperr-code`, `hiperr-name`, and `hiperr-str` in `-catch-hiperr` part of the manual.
- doc: add "HIP Runtime Error" to the menu  of Catchpoints.
- code: get rid of `original_location` in `*stopped` record of MI by not setting the `locspec` property of the catchpoint (breakpoint).

gdb/doc/gdb.texinfo
amd-shahab: thanks for this practical experiment. it works for me too and I've incorporated it as a separate patch.

gdb/doc/gdb.texinfo
```diff
@@ -37083,6 +37084,75 @@ and @samp{tcatch catch} (@pxref{Set Catchpoints}).
...
+  stopped-threads="all",core="0"
+=breakpoint-deleted,id="1"
+@end smallexample
+
+The @samp{*stopped} record is extended with three parameters:
```
palves: "parameters" is really the wrong world for this, this is not input, not something you can configure.

I think the word the manual uses is "fields".

CHANGELOG_AMD.md
```diff
@@ -5,6 +5,13 @@ Full documentation for ROCgdb is available at
 
 ## ROCgdb-X for ROCm-next
 
+### Added
+
+- The "catch hiperr" feature is now exposed to MI too, with a new
+  `-catch-hiperr` command and related attributes in `*stopped`
```
palves: While at it, use "fields" here too instead of "attributes" (my bad).

palves (commented):
Let's just correct the "fields" term.  Otherwise LGTM.  Thanks.

amd-shahab:
- doc: change "attributes" to "fields" in the changelog entry.
- doc: use "fields" instead of "parameters" in the manual subsection.
- refactor: use "info" (or "information" or even "fields") wherever "parameters" or one of its derivatives were used in the whole "catch hiperr" part (not only the MI part).

gdb/doc/gdb.texinfo
```diff
@@ -37083,6 +37084,75 @@ and @samp{tcatch catch} (@pxref{Set Catchpoints}).
...
+@samp{tcatch hiperr} (@pxref{catch hiperr}).
+
+@subsubheading Example
+
+@smallexample
+-catch-hiperr -c $_hiperr==hipErrorInvalidDevice -t
+^done,bkpt=@{number="1",type="catchpoint",disp="del",enabled="y",
+  what="catch hiperr",catch-type="hiperr",
+  cond="$_hiperr==hipErrorInvalidDevice",times="0",
+  original-location="-function __hipOnError"@}
```
palves: Stale "original-location".

amd-shahab:
- doc: get rid of stale `original-location` in MI output.

palves (approved):

amd-shahab:
- resolve conflict.

PR #184, opened 2026-06-25
[ci] Use therock-ci-config for dynamic runner selection

lumachad:
## Motivation

Runner labels and weights in rocgdb CI were hardcoded, requiring PRs every
time a label changes. Using ROCm/therock-ci-config as a dynamic source of
truth allows runner changes to propagate automatically, matching the approach
taken in TheRock, rocm-libraries, and rocm-systems.

## Technical Details

Switch rocgdb CI workflows to retrieve runner labels dynamically from
[ROCm/therock-ci-config](https://github.com/ROCm/therock-ci-config) instead
of hardcoding them.

Changes:
- Update `THEROCK_COMMIT_REF` to `76b6f7a` (2026-06-26) which includes
  `get_build_runner_labels()` and `load_external_config()` support
- Add **Checkout CI config** step (with `continue-on-error`) to
  `therock-ci-linux.yml` build job; pass `CI_CONFIG_PATH` to `build_configure.py`
- Replace monolithic `rocgdb-tests` job in `therock-test-packages.yml` with
  a `configure_test_matrix` + `test_components` pattern driven by
  `fetch_test_configurations.py` (`PROJECTS_TO_TEST=rocgdb-cpu,rocgdb-gpu`)
- Add new `therock-test-component.yml` for per-component test execution;
  `rocgdb-cpu` routes to a CPU runner, `rocgdb-gpu` routes to a GPU runner
  selected dynamically from `therock-ci-config`

## Test Plan

- Verify CI triggers on this PR and `configure_test_matrix` produces the
  expected `rocgdb-cpu` and `rocgdb-gpu` components
- Confirm `rocgdb-cpu` runs on `aws-linux-scale-rocm-prod` and `rocgdb-gpu`
  runs on the GPU runner resolved from `therock-ci-config`

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/.github/workflows/therock-test-packages.yml
+++ b/.github/workflows/therock-test-packages.yml
@@ -7,8 +7,9 @@ on:
         type: string
       artifact_group:
         type: string
-      test_runs_on:
+      therock_ref:
         type: string
+        required: true
       artifact_run_id:
         type: string
   workflow_dispatch:
@@ -17,8 +18,9 @@ on:
         type: string
       artifact_group:
         type: string
-      test_runs_on:
+      therock_ref:
         type: string
+        required: true
       artifact_run_id:
         type: string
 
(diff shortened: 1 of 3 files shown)


lumachad:
@lancesix cluster labels will change again, so we will want this change landed and picked by amd-staging-rocgdb-16 as well so we can keep running CI properly.

.github/workflows/therock-test-packages.yml
```diff
@@ -17,54 +18,92 @@ on:
...
-          IS_PR_FROM_FORK: ${{ github.event.pull_request.head.repo.fork }}
+          CI_CONFIG_PATH: ci-config
+        id: configure
+        run: python ./build_tools/github_actions/fetch_test_configurations.py
 
-      - name: Run rocgdb tests
+      - name: Override rocgdb test_script to use artifact-installed script
+        # TheRock's fetch_test_configurations.py points test_rocgdb.py at its own
+        # build_tools copy. Rewrite to the script installed in the artifacts at
+        # OUTPUT_ARTIFACTS_DIR/tests/rocgdb/test_rocgdb.py.
```
lumachad, line 71: If we happen to remove these scripts from TheRock and redirect the call to the component's scripts, then we can remove this. Or we can go with this one and subsequently remove this patching.

.github/workflows/therock-test-packages.yml
```diff
@@ -17,54 +18,92 @@ on:
...
+      artifact_run_id: ${{ inputs.artifact_run_id }}
+      therock_ref: ${{ inputs.therock_ref }}
+      amdgpu_families: ${{ inputs.amdgpu_families }}
+      # Per-component runner selection:
+      #   1. CPU-only components use a standard Linux scale runner
+      #   2. Per-component GPU runner (set by fetch_test_configurations.py)
+      test_runs_on: >-
+        ${{
+          (matrix.components.linux_cpu_runner == true &&
+            'aws-linux-scale-rocm-prod') ||
```
lancesix, line 105: I really find this syntax confusing.

Wouldn't

```js
matrix.components.linux_cpu_runner  ? 'aws-linux-scale-rocm-prod' : matrix.components.test_runner
```

be less confusing?

It really always trips me to read `true && foo` to evaluate to `foo`

.github/workflows/therock-test-packages.yml
lancesix: Don't we want to keep someithing in TheRock to be able to test upstream eventually anymay?

lumachad: Not in TheRock. This is handled here: https://github.com/ROCm/TheRock/pull/5123.

But my plan is to make that happen after we switch to multi-arch CI.

.github/workflows/therock-test-packages.yml
lumachad: Probably. But I want to keep this in sync with other superrepos (rocm-libraries/rocm-systems in particular), so we don't drift too much from how they do things: https://github.com/ROCm/rocm-systems/blob/3eb589e3e8ee1e2873ae54567dd53682f4e453c0/.github/workflows/therock-test-packages.yml#L109

lancesix: ok, but let it be noted that I do not really like those bits of JS.  I find this obscure/arcane for no good reason.

lancesix (approved):
OK if CI runs fine.

.github/workflows/therock-test-packages.yml
lancesix: Indeed, I happened to read this PR in the meantime.

.github/workflows/therock-test-packages.yml
lumachad: I'm not a big fan either. Hopefully all superrepos/modules can start sharing these files and then we can clean them up. Right now they're mostly duplicated.

PR #187, opened 2026-06-26
gdb/testsuite: deduplicate --offload-arch targets in hcc_amdgpu_targets

lumachad:
## Summary

- On systems with multiple identical GPUs, `hcc_amdgpu_targets` was emitting one `--offload-arch` flag per physical device, producing redundant entries (e.g. eight `--offload-arch=gfx942` flags on an 8-GPU MI300 system).
- Deduplicate the target list in `hcc_amdgpu_targets`, preserving order, using an array for O(1) membership tests. Applied to both the `HCC_AMDGPU_TARGET` env-var and `find_amdgpu_devices` code paths.
- Add `gdb.rocm/hcc-amdgpu-targets.exp` to unit-test the deduplication logic without requiring a GPU.

## Test plan

- [x] Run `make check TESTS="gdb.rocm/hcc-amdgpu-targets.exp"` and confirm all cases pass.
- [x] On a multi-GPU system, verify the compiler invocation now emits a single `--offload-arch=gfx942` instead of N copies.

### DIFF
--- a/gdb/testsuite/gdb.rocm/hcc-amdgpu-targets.exp
+++ b/gdb/testsuite/gdb.rocm/hcc-amdgpu-targets.exp
@@ -1,0 +1,81 @@
+# Copyright (C) 2026 Free Software Foundation, Inc.
+# Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+# This file is part of GDB.
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Unit tests for hcc_amdgpu_targets.  These exercise the deduplication
+# logic across both the env-var and device-enumeration code paths, and
+# do not require a GPU or a running GDB instance.
+
+load_lib rocm.exp
+
+# Run BODY with find_amdgpu_devices stubbed to return DEVICES, then
+# restore the original proc.
+proc with_stub_devices {devices body} {
+    rename find_amdgpu_devices __saved_find_amdgpu_devices
+    # tclint-disable-next-line command-args
+    proc find_amdgpu_devices {} [list return $devices]
+    set code [catch {uplevel 1 $body} result]
+    rename find_amdgpu_devices {}
+    rename __saved_find_amdgpu_devices find_amdgpu_devices
+    return -code $code $result
+}
+
+# Tests using find_amdgpu_devices (no HCC_AMDGPU_TARGET env var).
+
+save_vars { env(HCC_AMDGPU_TARGET) } {
+    unset -nocomplain ::env(HCC_AMDGPU_TARGET)
+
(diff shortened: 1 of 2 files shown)


lumachad:
Potential candidate for upstreaming, if we don't have a lot of conflicts.

gdb/testsuite/gdb.rocm/hcc-amdgpu-targets.exp
```diff
@@ -0,0 +1,99 @@
...
+
+save_vars { env(HCC_AMDGPU_TARGET) } {
+    unset -nocomplain ::env(HCC_AMDGPU_TARGET)
+
+    with_stub_devices {gfx942 gfx942 gfx942} {
+	set result [hcc_amdgpu_targets]
+	if {$result eq {gfx942}} {
+	    pass "duplicates removed: 3x gfx942 produces single entry"
+	} else {
+	    fail "duplicates not removed: got $result"
```
lancesix: It seems quite unusual to have a test with a different message based on PASS or FAIL.  This makes diffing tests quite difficult.


```
gdb_assert {$result eq {gf942}} "duplicates removed"
```

if you want to have effective result for debug, we should prefer using `verbose -log "result: $resurt"`, or have the diagnostic in parenthesis after the test name.

Same remark will apply to all further tests.

lancesix (commented):
Just one overall remark regarding using `pass` / `fail` with different messages for the same test.  We should have the same msg so we can diff a passing and failing cases, and identify those as the same tests, not different tests.

gdb/testsuite/gdb.rocm/hcc-amdgpu-targets.exp
lumachad: Ooops. Indeed. Turned them into gdb_asserts and verbose -log calls.

lancesix (approved):
OK, this LGTM.

Note that I expect the GPU detection code is also present upstream ([here](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/testsuite/lib/rocm.exp;h=15ee22dad260088a03513e10d98a8febddf75321;hb=refs/heads/master#l83)).  Would you mind checking if this applies upstream and sending it there?

If this causes inconvenience, we can merge here first, but we need to be sure to maintain this upstream.

lumachad:
I can apply this to master with minimal conflict. I'll park this one for now and pursue it upstream.

lumachad:
@lancesix  On the mailing list: https://inbox.sourceware.org/gdb-patches/20260629135613.2738058-1-luis.machado@amd.com/T/#u

lumachad:
Pushed upstream. We should pick this up in the following merges from master.

```
commit 95cd59ef89cac7621b121a723078c39daf5676a9
Author:     Luis Machado <luis.machado@amd.com>
AuthorDate: Mon Jun 29 08:56:13 2026 -0500
Commit:     Luis Machado <luis.machado.foss@gmail.com>
CommitDate: Fri Jul 3 09:32:34 2026 +0100

    gdb/testsuite: deduplicate --offload-arch targets in hcc_amdgpu_targets

    On systems with multiple identical GPUs, find_amdgpu_devices returns one
    entry per device, causing hcc_amdgpu_targets to emit redundant
    --offload-arch flags when compiling HIP test programs (e.g. eight
    --offload-arch=gfx942 flags on an 8-GPU MI300 system).

    Deduplicate the target list in hcc_amdgpu_targets, preserving order, so
    each unique architecture appears exactly once regardless of how many
    physical devices share it.  The same deduplication is applied when the
    list comes from the HCC_AMDGPU_TARGET environment variable.  An array
    is used for the seen-set to give O(1) membership tests.

    Add gdb.rocm/hcc-amdgpu-targets.exp to unit-test the deduplication
    logic across both the env-var and device-enumeration code paths, without
    requiring a GPU.

    Approved-by: Lancelot Six <lancelot.six@amd.com>
```

PR #190, opened 2026-06-30
cleanup around wave coordinates

aktemur:
## Motivation

Code simplification and reuse.

## Technical Details

Use existing `wave_coordinates` to reduce duplication.

## Test Plan

No new tests needed. This is a refactoring.

### DIFF
--- a/gdb/amd-dbgapi-target.c
+++ b/gdb/amd-dbgapi-target.c
@@ -709,53 +709,41 @@ struct work_item_info
   }
 };
 
-/* Populate WI, a work_item_info object.
-   Returns true on success, false if info is not available.  */
-
-static bool
-make_work_item_info (thread_info *tp, work_item_info *wi)
-{
-  wave_info &info = get_thread_wave_info (tp);
-  if (info.coords.dispatch_id == AMD_DBGAPI_DISPATCH_NONE)
-    {
-      /* The dispatch associated with a wave is not available.  A wave
-	 may not have an associated dispatch if attaching to a process
-	 with already existing waves.  */
-      return false;
-    }
-
-  wi->dispatch_id = info.coords.dispatch_id;
-  wi->queue_id = info.coords.queue_id;
-  wi->agent_id = info.coords.agent_id;
-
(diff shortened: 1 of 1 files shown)


palves (approved):
Very cool.  I had wanted something like this since 7d1fd6713582f2436a9a569ac7a8700c75e35a42 .  Thanks for doing this!

LGTM.

lumachad (approved):

gdb/amd-dbgapi-target.c
```diff
@@ -436,13 +447,12 @@ get_amd_dbgapi_inferior_info (inferior *inferior)
...
+wave_coordinates::hierarchy_str () const
 {
-  std::string str = "AMDGPU Wave";
-
-  str += (agent_id != AMD_DBGAPI_AGENT_NONE
-	  ? string_printf (" %ld", agent_id.handle)
-	  : " ?");
+  std::string str
+    = (agent_id != AMD_DBGAPI_AGENT_NONE
+       ? string_printf ("%ld", agent_id.handle)
```
lancesix: I think the handle are uint64_t.  I expect `%ld` to work on Linux, but Windows might need `lld`.

I would suggest using something more portable.

lancesix (commented):
Just a very quick scan through, and I expect this PR has one portability problem.

My usual question: how much of this can be sent upstream directly?

gdb/amd-dbgapi-target.c
aktemur: > I think the handle are uint64_t. I expect `%ld` to work on Linux, but Windows might need `lld`.
> 
> I would suggest using something more portable.

There is now a patch in upstream that uses pulongest. I added another patch to this PR to use pulongest for the cases that exist downstream only.

aktemur:
> My usual question: how much of this can be sent upstream directly?

Two patches went to upstream: One that converts `%ld` cases to pulongest, and another that splits `wave_coordinates::tro_string`. I revised this PR on top of those upstream patches, which are now available in amd-staging.

aktemur:
The latest update is just a rebase on the current amd-staging. Patch contents did not change.

lancesix (approved):
With the rebase, if tests are happy, then all good for me as well.

Thanks.

PR #191, opened 2026-06-30
gdb, testsuite: increase timeout in gdb.rocm/step-schedlock-spurious-…

aktemur:
In a test where we are doing "next", the thread steps over a for-loop header, single-stepping roughly 90 instructions.  With a debug build of GDB and dbgapi, the command takes about 10 seconds and may occasionally fail with a timeout.  Increase the timeout temporarily.

### DIFF
--- a/gdb/testsuite/gdb.rocm/step-schedlock-spurious-waves.exp
+++ b/gdb/testsuite/gdb.rocm/step-schedlock-spurious-waves.exp
@@ -113,8 +113,10 @@ proc do_test {} {
 	# (the only printed line is the source line number and the source line).
 	set num_nexts 5
 	for {set i 0} {$i < 5} {incr i} {
-	    gdb_test "next" "^${::decimal}\[^\r\n\]+" \
-		"next, iter $i"
+	    with_timeout_factor 2 {
+		gdb_test "next" "^${::decimal}\[^\r\n\]+" \
+		    "next, iter $i"
+	    }
 	}
 
 	gdb_test "thread" "Current thread is $cur_thread, .*" \


lumachad (approved):
OK by me if CI is good.

palves:
Did you figure out where was the time spent, with e.g., "set debug amd-dbgapi-lib log-level"?

I'm surprised optimization in gdb and dbgapi makes so much of a difference here; I'd assume most of the time is spent on the actual stepping the wave, and parsing the wave states from the save area.  

I ran this test so many times when working on memory access perf improvements and never saw this timeout, which makes me wonder if we're doing something we shouldn't in a debug build.

aktemur:
> Did you figure out where was the time spent, with e.g., "set debug amd-dbgapi-lib log-level"?

Yes,  it's when going from
```
[amd-dbgapi-lib] resuming queue_2 (forward progress required)
```
to
```
[amd-dbgapi-lib] queue_2 has pending exceptions (QUEUE_WAVE_TRAP)
```

I guess you're going after more details, though.

I also starting seeing this timeout failure recently. I don't know if it was already on the verge of timing out and something small just made it pass the threshold.

lumachad:
If it's a heavy test and we happen to find busy GPU's on the testing system it doesn't help.

aktemur:
I'm merging this. I looked at the dbgapi verbose logs and I didn't see an obvious, low-hanging fruit. This will need more investigation and discussion. The patch will at least prevent the test failures. Note that the test here spawns thousands of GPU waves. Stepping in a program with a handful of waves is much faster.

PR #192, opened 2026-06-30
gdb/testsuite: replace hipcc with amdclang++ as the HIP compiler

spatrang:
## Summary

Switch the GDB testsuite's HIP compilation from the deprecated `hipcc` to `amdclang++`, in preparation for upstreaming to sourceware. This is a sanity-review PR against the fork's `master`; it is intended to be **closed (not merged)** once reviewed, and the change submitted upstream via `git send-email` to `gdb-patches@sourceware.org`.

Testsuite-only change (21 files):

- **Compiler discovery** (`lib/future.exp`): find/invoke `amdclang++` instead of `hipcc`; rename `HIPCC_FOR_TARGET` → `HIP_COMPILER_FOR_TARGET` and `gdb_find_hipcc`/`find_hipcc` → `gdb_find_hip_compiler`/`find_hip_compiler`. The `hipcompiler` board key is kept for compatibility.
- **Compile/link flags** (`lib/gdb.exp`): per-input `-x hip`/`-x none` tagging, `-mllvm=-amdgpu-spill-cfi-saved-regs`, `--hip-link` for executables, `--rocm-path` only when `ROCM_PATH` is set, one `--offload-arch` per device; drop the now-redundant explicit `-O0` (amdclang++ already defaults to `-O0`).
- **Testcases** (`lib/rocm.exp` + 18 `gdb.rocm/*.exp`): rename `allow_hipcc_tests` → `allow_hip_tests`.

## Testing

Built `master` with this change and ran all `gdb.rocm/*.exp` on gfx942: **252 expected passes, 0 failures** (17 of 18 tests). Confirmed the actual compile uses `amdclang++ ... -x hip ... --hip-link --offload-arch=gfx942`.

## Known pre-existing failure (not caused by this change)

`gdb.rocm/interrupt-twice.exp` errors with `ERROR: Args left unparsed: allow-pending temporary`.

This is **pre-existing on `master`** and unrelated to the compiler switch: the test calls `gdb_breakpoint "loop" {allow-pending} {temporary}` (bare flags), while `gdb_breakpoint`/`parse_args` expect the dashed form `-allow-pending -temporary` (which every other test in the tree uses). Verified by running the test on pristine `master` (before this patch) — it produces the identical error. `gdb_breakpoint` and `parse_args` are byte-identical between `master` and `amd-staging`. This will be addressed in a separate PR.


spatrang:
This PR is for review purposes before submitting the change to upstream. Once everyone is OK with it, I'll proceed with the upstream work (submitting via `git send-email` to `gdb-patches@sourceware.org`) and close this PR.

gdb/testsuite/lib/gdb.exp
```diff
@@ -6754,11 +6754,64 @@ proc gdb_compile {source dest type options} {
...
+	# like:
+	#
+	#  clang: warning: argument unused during compilation: \
+	#    '-mllvm=-amdgpu-spill-cfi-saved-regs' [-Wunused-command-line-argument]
+	#
+	# These happen when we're compiling an executable _and_ all
+	# the input files are .o files, like: amdclang++ foo.o -o foo ...
+	#
+	# amdclang++ does not infer the input language from the file
+	# extension.  Rather than passing a single global "-x hip" in
```
lancesix: Strictly speaking, amdclang++ does infer the language from the extension.

We can build a hip application without `-x hip` if we use the `.hip` extension:

```
$ amdclang++ -g -O0 --offload-arch=gfx1031 foo.hip -o foo
```

works.

So we *could* decide to rename all the testcases with `.hip`, but this is not that usual and we prefer to keep then named as .cpp.

I think this part of the comment should reflect this (as well as the commit message).

gdb/testsuite/lib/gdb.exp
```diff
@@ -6930,6 +6983,27 @@ proc gdb_compile {source dest type options} {
     # Automatically handle includes in testsuite/lib/.
     auto_lappend_include_files options $source
 
+    # For HIP, amdclang++ does not infer the input language from the
```
lancesix: same remark as above, this is not strictly true.

gdb/testsuite/lib/gdb.exp
lumachad: > Strictly speaking, amdclang++ does infer the language from the extension.
> 
> We can build a hip application without `-x hip` if we use the `.hip` extension:
> 
> ```
> $ amdclang++ -g -O0 --offload-arch=gfx1031 foo.hip -o foo
> ```
> 
> works.
> 
> So we _could_ decide to rename all the testcases with `.hip`, but this is not that usual and we prefer to keep then named as .cpp.
> 
> I think this part of the comment should reflect this (as well as the commit message).

Given we merged this change to amd-staging, I'd be cautious revisiting this under the risk of creating more upstream churn and more downstream conflicts. If we're inclined to change this, we should do it in amd-staging first I think. And then update the change that is going to go upstream.

palves: > Given we merged this change to amd-staging, I'd be cautious revisiting this under the risk of creating more upstream churn and more downstream conflicts. If we're inclined to change this, we should do it in amd-staging first I think. And then update the change that is going to go upstream.

Agreed if it's some design change.  For some simple small things though, that would be more trouble/difficult IMO than just handling the conflicts later (need to open new PR against amd-staging for any comments on this PR), which we _will_ have, since upstream does not contain some of the pieces we have downstream, yet, around rocm_agent_enumerator replacement, and Windows testing, etc.

Also, someone during upstream review could ask for changes too.  Like, imagine the patch had been sent to the mailing list directly without this PR.

So IMO, don't worry about that, and we'll handle any conflicts later.

spatrang: Good catch, thanks. Updated this comment (and the commit message) to say that amdclang++ does infer the language from the extension and compiles .cpp as C++; we pass -x hip explicitly because we keep the testcases named .cpp rather than renaming them all to .hip.

gdb/testsuite/lib/gdb.exp
spatrang: Same fix applied here.

spatrang:
**Updated (force-push `8e0f832` → `6f21205`):** addressed @lancesix's review comments — reworded the two `lib/gdb.exp` comment blocks and the commit message to state that amdclang++ *does* infer the input language from the file extension (it compiles `.cpp` as C++). Since the HIP testcases are kept named `.cpp` (rather than renamed to `.hip`), we pass `-x hip` explicitly. No functional change — comment/commit-message wording only.

czidev-amd:
My only concern is this change haven't been tested also on a windows host, but it is not a blocker.

lumachad:
> My only concern is this change haven't been tested also on a windows host, but it is not a blocker.

Could we have some feedback from Windows folks on this? Otherwise it is a bit tricky for us to validate when there isn't a proper setup to do it.

palves:
> > My only concern is this change haven't been tested also on a windows host, but it is not a blocker.
> 
> Could we have some feedback from Windows folks on this? Otherwise it is a bit tricky for us to validate when there isn't a proper setup to do it.

As I said before, upstream does not contain some of the pieces we have downstream, yet, around rocm_agent_enumerator replacement, and Windows testing.  This version cannot be tested on Windows.

I had tested the original amd-staging-based version on Windows (by rebasing it on amd-temp-windows and handling a few of the conflicts, locally).

gdb/testsuite/lib/gdb.exp
```diff
@@ -6754,11 +6754,67 @@ proc gdb_compile {source dest type options} {
...
+	# unfortunately also affects code generation.  We don't want
+	# to hardcode -g here though, because some tests want to test
+	# debugging without debug info, and yet others want to compile
+	# their binary without debug info or with -gline-info-only,
+	# and then use the DWARF assembler to generate their own debug
+	# info.  Some features of the DWARF assembler rely on nodebug
+	# and debug binaries having the exact same generated code, so,
+	# explicitly compile with the LLVM option that -g would
+	# enable, and leave adding -g to the testcase, via the "debug"
+	# option, as normal.
```
palves, line 6760: This whole comment and the actual change that comes with it, is unrelated to the hipcc => amdclang++ conversion.  

It comes from 78465a34282ccd7aeb2f4f191ba3f8f9443ef145 .  

The hipcc => amdclang++ change should not grab unrelated changes like this.

It'll probably be easier if we upstream that one first.  Let me see about doing that quickly.

palves: Sent upstream here:

https://inbox.sourceware.org/gdb-patches/20260707111527.1211730-1-pedro@palves.net/T/

@lancesix ^^^

palves: That is now merged upstream.  @spatrang please rebase once our "master" branch mirror picks that up.

spatrang: > That is now merged upstream. @spatrang please rebase once our "master" branch mirror picks that up.

Sure, Let me do that.

spatrang:
**Rebased onto `master` (force-push `6f21205` → `da5c1338`).**

Now that Pedro's `78465a3` ("gdb/hip: Avoid hardcoding -ggdb for all testcases") has landed in `master`, I rebased on it and **trimmed this patch to only the hipcc → amdclang++ conversion**. The `-g`/`-mllvm=-amdgpu-spill-cfi-saved-regs` comment block and flag are no longer part of this patch — they're now unchanged context inherited from `master`. The diff is down to the amdclang++ delta: compiler discovery, `-x hip`/`-x none` tagging, `--hip-link`/`--rocm-path`/`--offload-arch`, dropping the explicit `-O0`, and the `allow_hipcc_tests` → `allow_hip_tests` rename.

palves:
This looks ready to be sent upstream, to me.

palves (approved):

lumachad:
Closing so we don't try to merge it. @spatrang Thanks for the updates. Please pursue this version upstream.

PR #196, opened 2026-07-07
gdb/testsuite: Fix runtime-core.exp wave ordering dependency

amd-bfilipov:
The "identify waves" test was failing when waves appeared in different
order because the patterns were not anchored to line boundaries, causing
expect to match patterns out of order even with -lbl mode.

Problem:
The test uses gdb_test_multiple with -lbl (line-by-line mode) expecting
each pattern to match complete lines in sequence. However, without proper
anchoring, expect could match a later pattern before an earlier one if
the later pattern's text appeared first in the buffer, consuming the
buffer incorrectly.

When aux_kernel wave appeared before pagefault_kernel wave:
- pagefault_kernel pattern matched its line first (even though it appeared
  later in output)
- Buffer was consumed up to that match
- aux_kernel line (which appeared earlier) was already consumed
- Second pattern failed to match → test FAIL

Solution:
Anchor patterns to line boundaries using:
- `^\r\n` - enforce beginning of line after previous line ending
- `[^\r\n]+` - match line prefix before the wave info
- `(?=\r\n)` - lookahead to ensure pattern matches up to line ending

This ensures each pattern matches a complete line from start to finish,
making the test properly order-independent as -lbl mode intended.

The same anchoring is applied to the gdb_prompt pattern for consistency.

### DIFF
--- a/gdb/testsuite/gdb.rocm/runtime-core.exp
+++ b/gdb/testsuite/gdb.rocm/runtime-core.exp
@@ -86,15 +86,15 @@ proc do_test { fault core_type output_type } {
     set faulty_wave 0
     set aux_wave 0
     gdb_test_multiple "info thread" "identify waves" -lbl {
-	-re "($::decimal) *AMDGPU Wave \[^\r\n\]*$fault_loc \[^\\n\]*" {
+	-re "^\r\n\[^\r\n\]+($::decimal) *AMDGPU Wave \[^\r\n\]*$fault_loc \[^\r\n\]*(?=\r\n)" {
 	    set faulty_wave $expect_out(1,string)
 	    exp_continue
 	}
-	-re "($::decimal) *AMDGPU Wave \[^\r\n\]* aux_kernel \[^\\n\]*" {
+	-re "^\r\n\[^\r\n\]+($::decimal) *AMDGPU Wave \[^\r\n\]* aux_kernel \[^\r\n\]*(?=\r\n)" {
 	    set aux_wave $expect_out(1,string)
 	    exp_continue
 	}
-	-re "$::gdb_prompt $" {
+	-re "^\r\n$::gdb_prompt $" {
 	    gdb_assert {($faulty_wave != 0) && ($aux_wave != 0)} $gdb_test_name
 	}
     }


gdb/testsuite/gdb.rocm/runtime-core.exp
```diff
@@ -86,12 +86,13 @@ proc do_test { fault core_type output_type } {
...
     set aux_wave 0
     gdb_test_multiple "info thread" "identify waves" -lbl {
-	-re "($::decimal) *AMDGPU Wave \[^\r\n\]*$fault_loc \[^\\n\]*" {
-	    set faulty_wave $expect_out(1,string)
-	    exp_continue
-	}
-	-re "($::decimal) *AMDGPU Wave \[^\r\n\]* aux_kernel \[^\\n\]*" {
-	    set aux_wave $expect_out(1,string)
+	-re "($::decimal) *AMDGPU Wave \[^\r\n\]*($fault_loc|aux_kernel)\[^\\n\]*" {
+	    if {[string match "*$fault_loc*" $expect_out(0,string)]} {
```
lancesix: you could probably extract `$expect_out(1,string)` first, then match it against `$fault_loc` or `aux_kernel`.

gdb/testsuite/gdb.rocm/runtime-core.exp
```diff
@@ -86,12 +86,13 @@ proc do_test { fault core_type output_type } {
     set faulty_wave 0
     set aux_wave 0
     gdb_test_multiple "info thread" "identify waves" -lbl {
-	-re "($::decimal) *AMDGPU Wave \[^\r\n\]*$fault_loc \[^\\n\]*" {
-	    set faulty_wave $expect_out(1,string)
-	    exp_continue
-	}
-	-re "($::decimal) *AMDGPU Wave \[^\r\n\]* aux_kernel \[^\\n\]*" {
-	    set aux_wave $expect_out(1,string)
+	-re "($::decimal) *AMDGPU Wave \[^\r\n\]*($fault_loc|aux_kernel)\[^\\n\]*" {
```
lancesix: still has the `\[^\\n\]`, should be `\[^\r\n\]`.

lancesix (commented):
I am quite surprised that when having

```
gdb_test_multiple "foo" "foo" -lbl {
  -re "$a" {
    exp_continue
  }
  -re "$b" {
    exp_continue
  }
}
```

In our case, $b does not cover $a and vice and versa, and expect that the `-lbl` should make it so we consume line by line, so should not match $b first if $a could have matched a previous line.

@palves any idea?

Anyway, given some comments, the alternative approach from this patch should be OK.

aktemur:
> I am quite surprised that when having
> 
> ```
> gdb_test_multiple "foo" "foo" -lbl {
>   -re "$a" {
>     exp_continue
>   }
>   -re "$b" {
>     exp_continue
>   }
> }
> ```
> 
> In our case, $b does not cover $a and vice and versa, and expect that the `-lbl` should make it so we consume line by line, so should not match $b first if $a could have matched a previous line.
> 
> @palves any idea?
> 
> Anyway, given some comments, the alternative approach from this patch should be OK.

I think the best approach for such cases is to use `^` to enforce the beginning of the buffer and to match up to `\r\n`. Like this, for instance:
```diff
diff --git a/gdb/testsuite/gdb.rocm/runtime-core.exp b/gdb/testsuite/gdb.rocm/runtime-core.exp
index 7a4cf7036f1..4c8d6506744 100644
--- a/gdb/testsuite/gdb.rocm/runtime-core.exp
+++ b/gdb/testsuite/gdb.rocm/runtime-core.exp
@@ -86,15 +86,15 @@ proc do_test { fault core_type output_type } {
     set faulty_wave 0
     set aux_wave 0
     gdb_test_multiple "info thread" "identify waves" -lbl {
-       -re "($::decimal) *AMDGPU Wave \[^\r\n\]*$fault_loc \[^\\n\]*" {
+       -re "^\r\n\[^\r\n\]+($::decimal) *AMDGPU Wave \[^\r\n\]*$fault_loc \[^\r\n\]*(?=\r\n)" {
            set faulty_wave $expect_out(1,string)
            exp_continue
        }
-       -re "($::decimal) *AMDGPU Wave \[^\r\n\]* aux_kernel \[^\\n\]*" {
+       -re "^\r\n\[^\r\n\]+($::decimal) *AMDGPU Wave \[^\r\n\]* aux_kernel \[^\r\n\]*(?=\r\n)" {
            set aux_wave $expect_out(1,string)
            exp_continue
        }
-       -re "$::gdb_prompt $" {
+       -re "^\r\n$::gdb_prompt $" {
            gdb_assert {($faulty_wave != 0) && ($aux_wave != 0)} $gdb_test_name
        }
     }
```

lancesix:
indeed @aktemur, I missed to the ^ bit in the match, and that was the important part…  Thanks.

aktemur:
@amd-bfilipov, can you confirm the diff I gave above works for you?

amd-bfilipov:
> @amd-bfilipov, can you confirm the diff I gave above works for you?

Works, I amended the commit to use your diff. Thanks.

aktemur (approved):

lancesix (approved):

palves:
> I am quite surprised that when having
> 
> ```
> gdb_test_multiple "foo" "foo" -lbl {
>   -re "$a" {
>     exp_continue
>   }
>   -re "$b" {
>     exp_continue
>   }
> }
> ```
> 
> In our case, $b does not cover $a and vice and versa, and expect that the `-lbl` should make it so we consume line by line, so should not match $b first if $a could have matched a previous line.
> 
> @palves any idea?

Sorry, I didn't notice your question until now.  For the record:

I think you are right expect-wise, but the issue is that your premise "$b does not cover $a and vice and versa" is not correct, due to the implicit leading `.*`.

The issue is that waves can appear in any order.  So if 'a' is one wave, and 'b' is another, we can see either:

`$a\r\n$b` or `$b\r\n$a`

If we get `$b\r\n$a` all at once in the expect buffer, then the match for `.*$a` (note implicit `.*`) will consume it, as it is the earlier match.

The anchoring fix is the right one, thanks.

PR #197, opened 2026-07-08
gdb/testsuite: add HIP graph-launch kernel debugging test

spatrang:
## Summary

Adds a new `gdb.rocm` dejagnu test, `hip-graph-launch`, covering debugging of
kernels dispatched through a HIP graph (`hipGraphLaunch`) rather than a plain
`kernel<<<>>>()` launch. The program captures two kernels (`add_one` then
`times_three`) from a stream into a single graph, instantiates it, and replays
it 3 times; each replay computes `out = (out + 1) * 3` (0 -> 3 -> 12 -> 39).

The test exercises the graph-specific debugger behavior:
- pending breakpoints in graph-launched kernels resolve and hit,
- both graph nodes stop in order on every replay (dispatch numbering / re-dispatch),
- correct per-replay data is observed at each node,
- symbol breakpoint + backtrace + kernel-argument inspection,
- single-stepping inside a graph-launched kernel mutates device memory,
- `info dispatches` reports the active dispatch and associates it with the kernel.

## Test plan

Validated 22/22 assertions under both GCC and LLVM host compilers on gfx942
(MI300X), against:
- locally built ROCgdb (gdb 18.0.50), and
- nightly ROCm 7.14 rocgdb (gdb 16.3).

## JIRA

[AIROCGDB-580](https://amd-hub.atlassian.net/browse/AIROCGDB-580)

### DIFF
--- a/gdb/testsuite/gdb.rocm/hip-graph-launch.cpp
+++ b/gdb/testsuite/gdb.rocm/hip-graph-launch.cpp
@@ -1,0 +1,106 @@
+/* Copyright (C) 2026 Free Software Foundation, Inc.
+   Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+   This file is part of GDB.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.  */
+
+/* Build a HIP graph by capturing two kernel launches from a stream into a
+   single graph, instantiate it, and then launch (replay) the executable
+   graph several times with hipGraphLaunch.  This exercises the debugger's
+   ability to associate stops with the right kernel and source line when the
+   kernels reach the GPU through a graph launch -- the host submits a whole
+   pre-recorded graph of operations with hipGraphLaunch, which dispatches the
+   kernels -- rather than through a direct kernel<<<>>>() call.
+
+   The two kernels operate on the same buffer, in order, so each graph
+   replay computes out = (out + 1) * 3.  Starting from 0 this gives 3, 12
+   and 39 after the first, second and third replay respectively.  The
+   deterministic values let the .exp verify that each node executed in the
+   right order with the right data on every replay.  */
+
+#include <stdio.h>
+#include <hip/hip_runtime.h>
+
+#include "rocm-test-utils.h"
+
+/* Number of times the executable graph is replayed.  The .exp overrides
+   this via -DNUM_REPLAYS=... so the source and test stay in sync; the
+   default lets the program build on its own.  */
+#ifndef NUM_REPLAYS
+#define NUM_REPLAYS 3
+#endif
+
+/* First graph node.  */
+
+__global__ void
+add_one (int *out)
+{
+  int tid = threadIdx.x;
+  out[tid] = out[tid] + 1; /* break add_one */
+}
+
+/* Second graph node, run after add_one on the same buffer.  */
+
+__global__ void
+times_three (int *out)
+{
+  int tid = threadIdx.x;
+  out[tid] = out[tid] * 3; /* break times_three */
+}
+
+int
+main ()
+{
+  constexpr unsigned int num_elems = 1;
+
+  int *result_ptr;
+  CHECK (hipMalloc (&result_ptr, num_elems * sizeof (int)));
+  CHECK (hipMemset (result_ptr, 0, num_elems * sizeof (int)));
+
+  hipStream_t stream;
+  CHECK (hipStreamCreate (&stream));
+
+  /* Capture two kernel launches into a single graph.  */
+  CHECK (hipStreamBeginCapture (stream, hipStreamCaptureModeGlobal));
+  add_one<<<dim3 (1), dim3 (num_elems), 0, stream>>> (result_ptr);
+  times_three<<<dim3 (1), dim3 (num_elems), 0, stream>>> (result_ptr);
+  hipGraph_t graph;
+  CHECK (hipStreamEndCapture (stream, &graph));
+
+  /* Turn the graph into an executable graph.  */
+  hipGraphExec_t graph_exec;
+  CHECK (hipGraphInstantiate (&graph_exec, graph, nullptr, nullptr, 0));
+
+  /* This is the "launch graph" execution path: instead of the host
+     issuing each kernel<<<>>>(), hipGraphLaunch submits the whole graph and
+     re-dispatches both kernels on every replay.  The debugger should stop
+     in them on each launch.  */
+  for (int i = 0; i < NUM_REPLAYS; i++)
+    {
+      CHECK (hipGraphLaunch (graph_exec, stream));
+      CHECK (hipStreamSynchronize (stream));
+    }
+
+  int result;
+  CHECK (hipMemcpyDtoH (&result, result_ptr, sizeof (int)));
+  printf ("result is %d\n", result);
+
+  CHECK (hipGraphExecDestroy (graph_exec));
+  CHECK (hipGraphDestroy (graph));
+  CHECK (hipStreamDestroy (stream));
+  CHECK (hipFree (result_ptr));
+
+  return 0;
+}
--- a/gdb/testsuite/gdb.rocm/hip-graph-launch.exp
+++ b/gdb/testsuite/gdb.rocm/hip-graph-launch.exp
@@ -1,0 +1,94 @@
+# Copyright (C) 2026 Free Software Foundation, Inc.
+# Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+# This file is part of GDB.
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test debugging kernels launched through a HIP graph (graph replay).
+#
+# The program captures two kernels (add_one then times_three) from a stream
+# into a single graph, instantiates it, and replays it NUM_REPLAYS times with
+# hipGraphLaunch.  Because the kernels reach the GPU through a graph launch
+# (the host submits a whole pre-recorded graph with hipGraphLaunch, which
+# dispatches the kernels) rather than a direct kernel<<<>>>() call, this
+# checks that the debugger still:
+#
+#   - resolves pending breakpoints in graph-launched kernels
+#   - associates each stop with the right kernel and source line
+#   - stops in each graph node, in order, on every replay (dispatch
+#     numbering / re-dispatch)
+#   - observes the correct per-replay data produced by each node
+
+load_lib rocm.exp
+
+require allow_hip_tests
+
+standard_testfile .cpp
+
+# Number of times the graph is replayed.  Passed to the C source via
+# -DNUM_REPLAYS so the source and test always stay in sync.
+set num_replays 3
+
+if {[build_executable "failed to prepare" $testfile $srcfile \
+	 [list debug hip additional_flags=-DNUM_REPLAYS=$num_replays]]} {
+    return
+}
+
+set add_one_line [gdb_get_line_number "break add_one"]
+set times_three_line [gdb_get_line_number "break times_three"]
+
+# Both kernel nodes are hit, in order, on every graph replay, and each
+# observes the value expected for that replay.
+proc_with_prefix test_replays_and_ordering {} {
+    with_rocm_gpu_lock {
+	clean_restart $::testfile
+
+	if {![runto_main]} {
+	    return
+	}
+
+	gdb_breakpoint "$::srcfile:$::add_one_line" -allow-pending
+	gdb_breakpoint "$::srcfile:$::times_three_line" -allow-pending
+
+	# out starts at 0.  Each replay computes out = (out + 1) * 3, so at the
+	# add_one stop out holds the value from the previous replay, and at the
+	# times_three stop it holds that value + 1.
+	set val 0
+	for {set i 1} {$i <= $::num_replays} {incr i} {
+	    with_test_prefix "replay $i" {
+		gdb_test "continue" \
+		    "Breakpoint .* add_one .*$::srcfile:$::add_one_line.*" \
+		    "stop in add_one"
+		gdb_test "print out\[0\]" " = $val" \
+		    "add_one sees value from previous replay"
+
+		gdb_test "continue" \
+		    "Breakpoint .* times_three .*$::srcfile:$::times_three_line.*" \
+		    "stop in times_three"
+		gdb_test "print out\[0\]" " = [expr {$val + 1}]" \
+		    "times_three sees add_one result"
+
+		set val [expr {($val + 1) * 3}]
+	    }
+	}
+
+	# After all replays the program prints the final result and exits.
+	gdb_test "continue" \
+	    "result is $val.*$::inferior_exited_re normally.*" \
+	    "continue to end, result $val"
+    }
+}
+
+test_replays_and_ordering


spatrang:
Rebased onto latest amd-staging to pick up the record-full.c build fix and re-trigger CI.

gdb/testsuite/gdb.rocm/hip-graph-launch.cpp
```diff
@@ -0,0 +1,110 @@
...
+   replay computes out = (out + 1) * 3.  Starting from 0 this gives 3, 12
+   and 39 after the first, second and third replay respectively.  The
+   deterministic values let the .exp verify that each node executed in the
+   right order with the right data on every replay.  */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <hip/hip_runtime.h>
+
+#define CHECK(cmd)							  \
```
lancesix: You should be able to `#include "rocm-test-utils.h"`, and have the `CHECK` macro from there.

gdb/testsuite/gdb.rocm/hip-graph-launch.cpp
```diff
@@ -0,0 +1,110 @@
...
+times_three (int *out)
+{
+  int tid = threadIdx.x;
+  out[tid] = out[tid] * 3; /* break times_three */
+}
+
+int
+main ()
+{
+  const unsigned int num_elems = 1;
```
lancesix: could even be constexpr.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
```diff
@@ -0,0 +1,169 @@
...
+require allow_hip_tests
+
+standard_testfile .cpp
+
+if {[build_executable "failed to prepare" $testfile $srcfile {debug hip}]} {
+    return
+}
+
+# The C source uses "#define NUM_REPLAYS 3".  Keep this in sync.
+set num_replays 3
```
lancesix: If we really want to have them in sync, you could build with `-DNUM_REPLAYS=$num_replays`.  This way, things would always be in sync.

See example of doing this in `gdb.rocm/nonstop-mode.exp` for example.

If we want to be able to build the cpp on its own (without running dejagnu), we can have

```
#ifndef NUM_REPLAYS
#define NUM_REPLAYS 3
#endif
```

lancesix (commented):
Just minor comments for now, will get a deeper review at a later time.

gdb/testsuite/gdb.rocm/hip-graph-launch.cpp
spatrang: Done — now using the shared CHECK macro from rocm-test-utils.h.

gdb/testsuite/gdb.rocm/hip-graph-launch.cpp
spatrang: Done — num_elems is now constexpr.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
spatrang: Done — the .exp now passes -DNUM_REPLAYS=$num_replays and the source guards the define with #ifndef, so the source and test stay in sync (following the nonstop-mode.exp pattern).

spatrang:
Addressed review comments :
- Use the shared `CHECK` macro from `rocm-test-utils.h` instead of a local copy.
- `num_elems` is now `constexpr`.
- Keep the replay count in sync: the `.exp` passes `-DNUM_REPLAYS=$num_replays` and the source guards the define with `#ifndef` (per the `nonstop-mode.exp` pattern).

Also rebased onto latest amd-staging.

aktemur:
Assigning to myself to reduce the load on @lancesix.
@lancesix, feel free to assign it back to yourself.

gdb/testsuite/gdb.rocm/hip-graph-launch.cpp
```diff
@@ -0,0 +1,102 @@
...
+}
+
+int
+main ()
+{
+  constexpr unsigned int num_elems = 1;
+  int *result_ptr, result;
+  hipStream_t stream;
+  hipGraph_t graph;
+  hipGraphExec_t graph_exec;
```
aktemur: It's not a hard requirement, but the code style prefers to declare variables at the first time of definition. E.g. Decls above can be moved down, like so:
```
hipStream_t stream;
CHECK (hipStreamCreate (&stream));
...
hipGraph_t graph;
CHECK (hipStreamEndCapture (stream, &graph));

hipGraphExec_t graph_exec;
CHECK (hipGraphInstantiate (&graph_exec, graph, nullptr, nullptr, 0))
...
```

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
```diff
@@ -0,0 +1,171 @@
...
+# Number of times the graph is replayed.  Passed to the C source via
+# -DNUM_REPLAYS so the source and test always stay in sync.
+set num_replays 3
+
+if {[build_executable "failed to prepare" $testfile $srcfile \
+	 [list debug hip additional_flags=-DNUM_REPLAYS=$num_replays]]} {
+    return
+}
+
+# Line numbers of the breakpoint markers in the two kernels.
```
aktemur: Nit: I think this comment can be removed. It's rather repeating the code.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
```diff
@@ -0,0 +1,171 @@
...
+	# times_three stop it holds that value + 1.
+	set val 0
+	for {set i 1} {$i <= $::num_replays} {incr i} {
+	    with_test_prefix "replay $i" {
+		continue_to_kernel "add_one" $::add_one_line \
+		    "stop in add_one"
+		gdb_test "print out\[0\]" " = $val" \
+		    "add_one sees value from previous replay"
+
+		continue_to_kernel "times_three" $::times_three_line \
```
aktemur: It's mostly a personal taste: The proc `continue_to_kernel` is a wrapper around one `gdb_test` only. We can rather inline it and remove the proc. What we are testing would become clearer.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
```diff
@@ -0,0 +1,171 @@
...
+
+# Line numbers of the breakpoint markers in the two kernels.
+set add_one_line [gdb_get_line_number "break add_one"]
+set times_three_line [gdb_get_line_number "break times_three"]
+
+# Continue and expect to stop in KERNEL at LINE, with lane 0.  DESC is the
+# test description.
+proc continue_to_kernel {kernel line desc} {
+    gdb_test "continue" \
+	"Thread $::decimal .* hit Breakpoint $::decimal, with lane 0, $kernel \\(.*\\) at .*$::srcfile:$line.*" \
```
aktemur: If we want to make this shorter, this would also be ok, without losing quality of the test:
`Breakpoint .* $kernel .*$::srcfile:$line.*`

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
```diff
@@ -0,0 +1,171 @@
...
+	$desc
+}
+
+# Restart, run to main and set a pending breakpoint in each kernel node.
+proc restart_and_break_kernels {} {
+    clean_restart
+    gdb_load $::binfile
```
aktemur: `clean_restart $::testfile`

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
```diff
@@ -0,0 +1,171 @@
...
+		    "times_three sees add_one result"
+
+		set val [expr {($val + 1) * 3}]
+	    }
+	}
+
+	# After all replays the final result must be 39, and the program exits
+	# cleanly.
+	gdb_test "continue" \
+	    "result is $val.*\\\[Inferior 1 .* exited normally\\\]" \
```
aktemur: Can we use `$inferior_exited_re` here?

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
```diff
@@ -0,0 +1,171 @@
...
+	    "continue to end, result $val"
+    }
+}
+
+# Breakpoint by kernel symbol name (not line) resolves and hits under a graph
+# launch, and the backtrace shows the graph-launched kernel frame.
+proc_with_prefix test_breakpoint_by_name_and_backtrace {} {
+    with_rocm_gpu_lock {
+	clean_restart
+	gdb_load $::binfile
```
aktemur: `clearn_restart $::testfile`

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
```diff
@@ -0,0 +1,171 @@
...
+
+	if {![runto_main]} {
+	    return
+	}
+
+	gdb_breakpoint "add_one" -allow-pending
+	gdb_test "continue" \
+	    "Thread $::decimal .* hit Breakpoint $::decimal, with lane 0, add_one \\(.*\\) at .*$::srcfile:.*" \
```
aktemur: Nit: The shorter variant I gave above may be used.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
```diff
@@ -0,0 +1,171 @@
...
+	    "stop in add_one by symbol"
+
+	gdb_test "backtrace" \
+	    "#0\[ \t\]+add_one \\(out=$::hex\\) at .*$::srcfile:$::decimal.*" \
+	    "backtrace shows graph-launched kernel"
+
+	gdb_test "print out" " = \\(int \\*\\) $::hex" \
+	    "kernel argument is readable"
```
aktemur: We can print and check the value of `out[0]`, too. The value can be passed as an argument to the proc.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
```diff
@@ -0,0 +1,171 @@
...
+	gdb_test "info dispatches" \
+	    "\\* +$::decimal +AMDGPU Dispatch\[^\r\n\]*add_one\\(.*" \
+	    "info dispatches shows active graph-launched kernel"
+    }
+}
+
+test_replays_and_ordering
+test_breakpoint_by_name_and_backtrace
+test_single_step
+test_info_dispatches
```
aktemur: I'm not sure what testing value we gain by these 3 tests. The first one tested being able to hit the breakpoint inside graph-launched kernels, that's understood. But the other 3 seem like they test debug info and basic debug functionality, for which there already exist tests. I don't see why it should matter to do these for graph-launched kernels. I'd be in favor of removing them for making this test file simpler and less expensive (such tests accumulate and the overall test time may increase to unpleasant levels over time).

gdb/testsuite/gdb.rocm/hip-graph-launch.cpp
spatrang: Done — declarations moved to first use.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
spatrang: Done — comment removed.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
spatrang: Done — inlined the gdb_test and removed the proc.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
spatrang: Done — using the shorter regex.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
spatrang: Done — using clean_restart $::testfile.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
spatrang: Done — now using $inferior_exited_re.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
spatrang: Done — this test was removed (see below); the remaining test uses clean_restart $::testfile.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
spatrang: Done — this test was removed.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
spatrang: Done — this test was removed.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
spatrang: Agreed — removed the three tests and kept only the graph-specific replay/ordering test.

spatrang:
Addressed @aktemur's review comments (force-pushed):
- Removed the `backtrace`, `single-step`, and `info dispatches` tests; kept only the graph-specific replay/ordering test.
- Inlined `continue_to_kernel` into plain `gdb_test` calls and shortened the stop regexps.
- `clean_restart $::testfile` and `$inferior_exited_re` for the exit match.
- Declared variables at first use in the `.cpp`; removed a redundant comment.

lancesix:
> Assigning to myself to reduce the load on @lancesix. @lancesix, feel free to assign it back to yourself.

Thanks.

I had a quick pass at the current iteration, and that seems reasonable to me. Make that a +1 from me.

aktemur (approved):
LGTM. Thanks.

gdb/testsuite/gdb.rocm/hip-graph-launch.cpp
```diff
@@ -0,0 +1,103 @@
...
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.  */
+
+/* Build a HIP graph by capturing two kernel launches from a stream into a
+   single graph, instantiate it, and then launch (replay) the executable
+   graph several times with hipGraphLaunch.  This exercises the debugger's
+   ability to associate stops with the right kernel and source line when
+   the CPU launch path is "launch graph" rather than a plain
```
lumachad: I'd try to clarify what path "launch graph" is, if there is even a better way to describe it.

gdb/testsuite/gdb.rocm/hip-graph-launch.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+  hipStream_t stream;
+  CHECK (hipStreamCreate (&stream));
+
+  /* Capture two kernel launches into a single graph.  */
+  CHECK (hipStreamBeginCapture (stream, hipStreamCaptureModeGlobal));
+  add_one<<<dim3 (1), dim3 (num_elems), 0, stream>>> (result_ptr);
+  times_three<<<dim3 (1), dim3 (num_elems), 0, stream>>> (result_ptr);
+  hipGraph_t graph;
+  CHECK (hipStreamEndCapture (stream, &graph));
```
lumachad, line 80: This is the "launch graph" path, right?

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
```diff
@@ -0,0 +1,92 @@
...
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test debugging kernels launched through a HIP graph (graph replay).
+#
+# The program captures two kernels (add_one then times_three) from a stream
+# into a single graph, instantiates it, and replays it NUM_REPLAYS times with
+# hipGraphLaunch.  Because the CPU launch path is "launch graph" rather than a
```
lumachad: Same thing about the "launch graph" path. Can we describe it in a better way, if possible?

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
```diff
@@ -0,0 +1,92 @@
...
+# plain kernel<<<>>>(), this checks that the debugger still:
+#
+#   - resolves pending breakpoints in graph-launched kernels,
+#   - associates each stop with the right kernel and source line,
+#   - stops in each graph node, in order, on every replay (dispatch
+#     numbering / re-dispatch), and
+#   - observes the correct per-replay data produced by each node.
```
lumachad: NIT: If this is to be a list, we shouldn't use commas at the end of the entries I guess. Otherwise if we're using commas, we should probably keep it on the same paragraph.

I prefer the list to be honest.

lumachad (approved):
Some nits, but otherwise LGTM.

gdb/testsuite/gdb.rocm/hip-graph-launch.cpp
spatrang: Done — reworded to describe the path: the host submits a whole pre-recorded graph via hipGraphLaunch, which dispatches the kernels.

gdb/testsuite/gdb.rocm/hip-graph-launch.cpp
spatrang: Right — added a comment there making the "launch graph" path explicit.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
spatrang: Done — reworded the same way as the .cpp.

gdb/testsuite/gdb.rocm/hip-graph-launch.exp
spatrang: Done — kept it as a list and dropped the trailing commas.

spatrang:
Addressed @lumachad's comments (force-pushed): clarified the "launch graph" execution path in both the `.cpp` and `.exp` headers (and near `hipGraphLaunch`), and cleaned up the bullet list (kept as a list, dropped trailing commas). Comment-only changes; retested 13/13 under GCC and LLVM.

PR #199, opened 2026-07-10
Fix __builtin_verbose_trap support for GPU and CPU

amd-bfilipov:
Ticket: AIROCGDB-558

### Summary
Implements complete support for `__builtin_verbose_trap` on both AMD GPU and CPU (x86_64), making verbose trap diagnostic messages visible in backtraces.

### Problem
When `__builtin_verbose_trap` fired, the diagnostic inline frames were not visible in backtraces, making it impossible to see the trap category and message. Two issues prevented this:

1. **PC pointing past trap instruction**: After s_trap 2 on GPU, PC was not rewound to point at the trap instruction, causing GDB to miss the inlined DWARF frame containing the diagnostic info
2. **Inline frames hidden by default**: GDB's `skip_inline_frames()` intentionally hides inline frames to improve stepping experience, but this also hid the verbose trap frames

### Solution

This PR provides a two-part fix:

#### 1. PC Adjustment for s_trap 2 (GPU)
Rewind PC by 4 bytes when a wave stops with s_trap 2, similar to breakpoint handling. This ensures the PC remains in the inlined DWARF frame containing `__builtin_verbose_trap` debug info.

**Files changed:**
- `gdb/amd-dbgapi-target.c` - PC adjustment logic
- `gdb/testsuite/gdb.rocm/builtin_verbose_trap.{cpp,exp}` - Test coverage

#### 2. gdbarch Hook for Inline Frame Control (GPU + CPU)
Adds architecture-specific control over inline frame visibility using GDB's gdbarch pattern. Architectures can now prevent skipping of compiler-generated inline frames containing diagnostic information.

**Changes:**
- `gdb/gdbarch_components.py` - Define `show_verbose_trap_inline_frame` method
- `gdb/arch-utils.{c,h}` - Default implementation (preserves current behavior)
- `gdb/amdgpu-tdep.c` - AMD GPU implementation (checks `SIGABRT` + name pattern)
- `gdb/amd64-linux-tdep.c` - CPU x86_64 implementation (checks `SIGILL` + name pattern)
- `gdb/inline-frame.c` - Call gdbarch hook in skip loop
- `gdb/gdbarch-gen.{c,h}` - Auto-generated

**Detection logic:**
- Only shows frame when **both** conditions met:
  1. Stop signal matches trap type (`SIGABRT` for GPU s_trap 2, `SIGILL` for CPU ud2)
  2. Symbol name starts with `__clang_trap_msg$<category>$<message>`
- Preserves normal stepping behavior (frames hidden when user steps into them)

### Results

**Before:**
```
Thread 6 received signal SIGABRT, Aborted.
#0  test_trap_kernel () at test.cpp:24
```

**After (GPU):**
```
Thread 6 received signal SIGABRT, Aborted.
#0  __clang_trap_msg$check verbose$This is verbose trap! ()
    at test.cpp:23
#1  test_trap_kernel () at test.cpp:24
```

**After (CPU):**
```
Program received signal SIGILL, Illegal instruction.
#0  __clang_trap_msg$check negative$Value must be non-negative! ()
    at test.cpp:4
#1  check_value (value=-1) at test.cpp:5
#2  main () at test.cpp:12
```

### DIFF
--- a/gdb/amd-dbgapi-target.c
+++ b/gdb/amd-dbgapi-target.c
@@ -2076,14 +2076,18 @@ process_one_event (amd_dbgapi_inferior_info &info,
 	    if (thread == nullptr)
 	      thread = add_gpu_thread (info.inf, event_ptid);
 
-	    /* If the wave is stopped because of a software breakpoint, the
-	       program counter needs to be adjusted so that it points to the
-	       breakpoint instruction.
+	    /* If the wave is stopped because of a software breakpoint or an abort
+	       trap (s_trap 2, used by __builtin_verbose_trap), the program counter
+	       needs to be adjusted so that it points to the trap instruction.
+
+	       For abort traps, this ensures the PC remains in the inlined frame
+	       containing DWARF debug info for verbose trap messages.
 
 	       When dealing with a corefile, it is expected that the PC has
 	       been adjusted before generating the corefile, so no need to
 	       re-do it now.  */
-	    if ((stop_reason & AMD_DBGAPI_WAVE_STOP_REASON_BREAKPOINT) != 0
+	    if (((stop_reason & AMD_DBGAPI_WAVE_STOP_REASON_BREAKPOINT) != 0
+		|| (stop_reason & AMD_DBGAPI_WAVE_STOP_REASON_ASSERT_TRAP) != 0)
 		&& get_inferior_core_bfd (info.inf) == nullptr)
 	      {
 		regcache *regcache = get_thread_regcache (thread);
--- a/gdb/amd64-linux-tdep.c
+++ b/gdb/amd64-linux-tdep.c
@@ -2088,6 +2088,32 @@ amd64_linux_fetch_hiperr_info (frame_info_ptr frame)
   return amd64_fetch_hiperr_info (frame, struct_addr);
 }
 
+/* Determine whether to show an inline frame.
+   Show verbose trap frames (__clang_trap_msg$...) when SIGILL occurs.  */
+
+static bool
+amd64_linux_should_show_inline_frame (struct gdbarch *gdbarch,
+				      const struct symbol *func,
+				      enum gdb_signal stop_signal)
+{
+  /* Only show verbose trap frames when stopped due to illegal instruction.
+     The ud2 instruction used by verbose traps on x86_64 generates SIGILL.
+     This ensures we only show the frame when the trap actually fired,
+     not when user stepped into it with commands like "step".  */
+  if (stop_signal != GDB_SIGNAL_ILL)
+    return false;
+
+  /* Check if this is a verbose trap inline frame by looking for the
+     compiler-generated function name pattern.  */
+  const char *name = func->linkage_name ();
+  if (name == nullptr)
+    return false;
+
+  /* Verbose trap frames have names like:
+     "__clang_trap_msg$<category>$<message>"  */
+  return startswith (name, "__clang_trap_msg$");
+}
+
 static void
 amd64_linux_init_abi_common (struct gdbarch_info info, struct gdbarch *gdbarch,
 			     int num_disp_step_buffers)
@@ -2154,6 +2180,11 @@ amd64_linux_init_abi_common (struct gdbarch_info info, struct gdbarch *gdbarch,
   /* Extract hiperr info for 'catch hiperr'.  */
   set_gdbarch_fetch_hiperr_info
     (gdbarch, amd64_linux_fetch_hiperr_info);
+
+  /* Show verbose trap inline frames when stopped due to illegal
+     instruction.  */
+  set_gdbarch_should_show_inline_frame
+    (gdbarch, amd64_linux_should_show_inline_frame);
 }
 
 static void
--- a/gdb/amd64-linux-tdep.c
+++ b/gdb/amd64-linux-tdep.c
@@ -2088,6 +2088,33 @@ amd64_linux_fetch_hiperr_parameters (frame_info_ptr frame)
   return amd64_fetch_hiperr_parameters (frame, struct_addr);
 }
 
+/* Return true if the inline frame FUNC should be shown when stopped
+   due to STOP_SIGNAL.  This allows showing verbose trap inline frames
+   on x86_64.  */
+
+static bool
+amd64_linux_show_verbose_trap_inline_frame (struct gdbarch *gdbarch,
+					     const struct symbol *func,
+					     enum gdb_signal stop_signal)
+{
+  /* Only show verbose trap frames when stopped due to illegal instruction.
+     The ud2 instruction used by verbose traps on x86_64 generates SIGILL.
+     This ensures we only show the frame when the trap actually fired,
+     not when user stepped into it with commands like "step".  */
+  if (stop_signal != GDB_SIGNAL_ILL)
+    return false;
+
+  /* Check if this is a verbose trap inline frame by looking for the
+     compiler-generated function name pattern.  */
+  const char *name = func->linkage_name ();
+  if (name == nullptr)
+    return false;
+
+  /* Verbose trap frames have names like:
+     "__clang_trap_msg$<category>$<message>"  */
+  return startswith (name, "__clang_trap_msg$");
+}
+
 static void
 amd64_linux_init_abi_common (struct gdbarch_info info, struct gdbarch *gdbarch,
 			     int num_disp_step_buffers)
@@ -2154,6 +2181,10 @@ amd64_linux_init_abi_common (struct gdbarch_info info, struct gdbarch *gdbarch,
   /* Extract hiperr parameters for 'catch hiperr'.  */
   set_gdbarch_fetch_hiperr_parameters
     (gdbarch, amd64_linux_fetch_hiperr_parameters);
+
+  /* Show verbose trap inline frames when stopped due to abort.  */
+  set_gdbarch_show_verbose_trap_inline_frame
+    (gdbarch, amd64_linux_show_verbose_trap_inline_frame);
 }
 
 static void
--- a/gdb/amdgpu-tdep.c
+++ b/gdb/amdgpu-tdep.c
@@ -2023,6 +2023,31 @@ amdgpu_supports_arch_info (const struct bfd_arch_info *info)
   return status == AMD_DBGAPI_STATUS_SUCCESS;
 }
 
+/* Determine whether to show an inline frame.
+   Show verbose trap frames (__clang_trap_msg$...) when SIGABRT occurs.  */
+
+static bool
+amdgpu_should_show_inline_frame (struct gdbarch *gdbarch,
+				 const struct symbol *func,
+				 enum gdb_signal stop_signal)
+{
+  /* Only show verbose trap frames when stopped due to abort signal.
+     This ensures we only show the frame when the trap actually fired,
+     not when user stepped into it with commands like "step".  */
+  if (stop_signal != GDB_SIGNAL_ABRT)
+    return false;
+
+  /* Check if this is a verbose trap inline frame by looking for the
+     compiler-generated function name pattern.  */
+  const char *name = func->linkage_name ();
+  if (name == nullptr)
+    return false;
+
+  /* Verbose trap frames have names like:
+     "__clang_trap_msg$<category>$<message>"  */
+  return startswith (name, "__clang_trap_msg$");
+}
+
 static struct gdbarch *
 amdgpu_gdbarch_init (struct gdbarch_info info, struct gdbarch_list *arches)
 {
@@ -2274,6 +2299,9 @@ amdgpu_gdbarch_init (struct gdbarch_info info, struct gdbarch_list *arches)
     error (_("amd_dbgapi_architecture_get_info failed"));
 
   set_gdbarch_decr_pc_after_break (gdbarch, pc_adjust);
+
+  set_gdbarch_should_show_inline_frame
+    (gdbarch, amdgpu_should_show_inline_frame);
 
   /* Get info about address spaces.  */
   size_t address_space_count;
--- a/gdb/amdgpu-tdep.c
+++ b/gdb/amdgpu-tdep.c
@@ -2019,6 +2019,30 @@ amdgpu_supports_arch_info (const struct bfd_arch_info *info)
   return status == AMD_DBGAPI_STATUS_SUCCESS;
 }
 
+/* Implementation of gdbarch_show_verbose_trap_inline_frame for AMDGPU.  */
+
+static bool
+amdgpu_show_verbose_trap_inline_frame (struct gdbarch *gdbarch,
+				       const struct symbol *func,
+				       enum gdb_signal stop_signal)
+{
+  /* Only show verbose trap frames when stopped due to abort signal.
+     This ensures we only show the frame when the trap actually fired,
+     not when user stepped into it with commands like "step".  */
+  if (stop_signal != GDB_SIGNAL_ABRT)
+    return false;
+
+  /* Check if this is a verbose trap inline frame by looking for the
+     compiler-generated function name pattern.  */
+  const char *name = func->linkage_name ();
+  if (name == nullptr)
+    return false;
+
+  /* Verbose trap frames have names like:
+     "__clang_trap_msg$<category>$<message>"  */
+  return startswith (name, "__clang_trap_msg$");
+}
+
 static struct gdbarch *
 amdgpu_gdbarch_init (struct gdbarch_info info, struct gdbarch_list *arches)
 {
@@ -2270,6 +2294,9 @@ amdgpu_gdbarch_init (struct gdbarch_info info, struct gdbarch_list *arches)
     error (_("amd_dbgapi_architecture_get_info failed"));
 
   set_gdbarch_decr_pc_after_break (gdbarch, pc_adjust);
+
+  set_gdbarch_show_verbose_trap_inline_frame
+    (gdbarch, amdgpu_show_verbose_trap_inline_frame);
 
   /* Get info about address spaces.  */
   size_t address_space_count;
--- a/gdb/arch-utils.c
+++ b/gdb/arch-utils.c
@@ -1613,6 +1613,18 @@ core_file_exec_context::environment () const
   return e;
 }
 
+/* See arch-utils.h.  */
+
+/* Default implementation: don't show inline frames.  */
+
+bool
+default_should_show_inline_frame (struct gdbarch *gdbarch,
+				  const struct symbol *func,
+				  enum gdb_signal stop_signal)
+{
+  return false;
+}
+
 INIT_GDB_FILE (gdbarch_utils)
 {
   add_setshow_enum_cmd ("endian", class_support,
--- a/gdb/arch-utils.c
+++ b/gdb/arch-utils.c
@@ -1613,6 +1613,17 @@ core_file_exec_context::environment () const
   return e;
 }
 
+/* See arch-utils.h.  */
+
+bool
+default_show_verbose_trap_inline_frame (struct gdbarch *gdbarch,
+					const struct symbol *func,
+					enum gdb_signal stop_signal)
+{
+  /* By default, do not show verbose trap inline frames.  */
+  return false;
+}
+
 INIT_GDB_FILE (gdbarch_utils)
 {
   add_setshow_enum_cmd ("endian", class_support,
--- a/gdb/gdbarch_components.py
+++ b/gdb/gdbarch_components.py
@@ -2952,3 +2952,30 @@ SHADOW_STACK_ENABLED to false.
     predefault="default_get_shadow_stack_pointer",
     invalid=False,
 )
+
+Method(
+    comment="""
+Determine whether an inline frame should be shown in backtraces.
+
+This hook is called when GDB encounters an inline frame to decide whether
+it should be displayed to the user or hidden.  By default, inline frames
+are hidden during normal stepping to provide a source-level debugging
+experience.  However, some inline frames contain important diagnostic
+information that should always be visible.
+
+A common use case is verbose trap frames (e.g., __builtin_verbose_trap)
+which embed crash diagnostic messages in specially-named inline frames.
+When such a trap fires, the inline frame should be shown even though
+inline frames are normally hidden.
+
+The hook receives the inline frame's symbol and the signal that stopped
+execution, allowing architecture-specific code to make the decision.
+
+Return true if the inline frame should be shown, false to hide it.
+""",
+    type="bool",
+    name="should_show_inline_frame",
+    params=[("const struct symbol *", "func"), ("enum gdb_signal", "stop_signal")],
+    predefault="default_should_show_inline_frame",
+    invalid=False,
+)
--- a/gdb/gdbarch_components.py
+++ b/gdb/gdbarch_components.py
@@ -2944,3 +2944,26 @@ SHADOW_STACK_ENABLED to false.
     predefault="default_get_shadow_stack_pointer",
     invalid=False,
 )
+
+Method(
+    comment="""
+Return true if the inline frame represented by FUNC should NOT be skipped
+when stopped due to STOP_SIGNAL.  This allows architectures to show
+compiler-generated inline frames that contain verbose trap messages
+(e.g., __builtin_verbose_trap).
+
+By default, inline frames are skipped to provide better stepping experience.
+However, for verbose trap scenarios, the inline frame contains important
+diagnostic information that should be visible in backtraces.
+
+Architectures can use this hook to detect verbose trap frames by checking
+both the symbol name (e.g., starts with "__clang_trap_msg$") and the stop
+signal (e.g., GDB_SIGNAL_ABRT for abort traps), returning true only when
+both conditions indicate a verbose trap scenario.
+""",
+    type="bool",
+    name="show_verbose_trap_inline_frame",
+    params=[("const struct symbol *", "func"), ("enum gdb_signal", "stop_signal")],
+    predefault="default_show_verbose_trap_inline_frame",
+    invalid=False,
+)
--- a/gdb/inline-frame.c
+++ b/gdb/inline-frame.c
@@ -28,6 +28,7 @@
 #include "regcache.h"
 #include "symtab.h"
 #include "frame.h"
+#include "gdbarch.h"
 #include "cli/cli-cmds.h"
 #include "cli/cli-style.h"
 #include <algorithm>
@@ -428,10 +429,14 @@ skip_inline_frames (thread_info *thread, bpstat *stop_chain)
      which contains all of the inlined functions, we never skip this.  */
   int skipped_frames = 0;
 
+  struct gdbarch *gdbarch = get_frame_arch (get_current_frame ());
+  enum gdb_signal stop_signal = thread->stop_signal ();
+
   for (const auto sym : function_symbols)
     {
       if (stopped_by_user_bp_inline_frame (sym, stop_chain)
-	  || sym == function_symbols.back ())
+	  || sym == function_symbols.back ()
+	  || gdbarch_show_verbose_trap_inline_frame (gdbarch, sym, stop_signal))
 	break;
 
       ++skipped_frames;
--- a/gdb/inline-frame.c
+++ b/gdb/inline-frame.c
@@ -28,6 +28,7 @@
 #include "regcache.h"
 #include "symtab.h"
 #include "frame.h"
+#include "gdbarch.h"
 #include "cli/cli-cmds.h"
 #include "cli/cli-style.h"
 #include <algorithm>
@@ -428,10 +429,14 @@ skip_inline_frames (thread_info *thread, bpstat *stop_chain)
      which contains all of the inlined functions, we never skip this.  */
   int skipped_frames = 0;
 
+  struct gdbarch *gdbarch = get_frame_arch (get_current_frame ());
+  enum gdb_signal stop_signal = thread->stop_signal ();
+
   for (const auto sym : function_symbols)
     {
       if (stopped_by_user_bp_inline_frame (sym, stop_chain)
-	  || sym == function_symbols.back ())
+	  || sym == function_symbols.back ()
+	  || gdbarch_should_show_inline_frame (gdbarch, sym, stop_signal))
 	break;
 
       ++skipped_frames;
--- a/gdb/testsuite/gdb.base/builtin_verbose_trap.exp
+++ b/gdb/testsuite/gdb.base/builtin_verbose_trap.exp
@@ -1,0 +1,96 @@
+# Copyright (C) 2026 Free Software Foundation, Inc.
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test that __builtin_verbose_trap inline frames are shown when trap fires.
+# __builtin_verbose_trap is a Clang-only builtin, added in Clang 19.
+
+require {expr {[test_compiler_info clang*]
+	       && ![test_compiler_info {clang-1[0-8]-*}]
+	       && ![test_compiler_info {clang-[1-9]-*}]}}
+
+# Compiler Line Table Bug:
+# Some Clang versions emit incorrect line table information for
+# __builtin_verbose_trap, associating the trap instruction with the
+# previous source line instead of the __builtin_verbose_trap line.
+# This causes GDB to receive the trap signal immediately when stepping past
+# that line, rather than stopping at the __builtin_verbose_trap call first.
+# The tests below use xfail to handle this known compiler issue.
+
+standard_testfile .cpp
+
+if {[prepare_for_testing "failed to prepare" $testfile $srcfile {debug c++}]} {
+    return
+}
+
+if {![runto test_trap_function]} {
+    return
+}
+
+# Continue to trigger the trap.
+gdb_test "continue" \
+    "received signal SIGILL.*" \
+    "received trap signal"
+
+# Verify verbose trap message appears in backtrace.
+gdb_test "bt" \
+    "This is verbose trap.*" \
+    "verbose trap message appears in backtrace"
+
+# Test single stepping to the trap line.
+# Expected: step should stop at __builtin_verbose_trap line, then
+# next step triggers the trap.
+clean_restart $testfile
+
+if {![runto_main]} {
+    return
+}
+
+# Set breakpoint at "int x = 1" line to avoid line table ambiguity
+# when running to the test function.
+set line_x [gdb_get_line_number "int x = 1"]
+gdb_breakpoint "$srcfile:$line_x"
+gdb_test "continue" "Breakpoint.*$line_x.*" "run to line $line_x"
+
+# Check for compiler line table bug using info line.
+# If the __builtin_verbose_trap line shows "contains no code",
+# the trap instruction is not mapped to that line (bug exists).
+set has_line_table_bug 0
+set line_trap [gdb_get_line_number "__builtin_verbose_trap"]
+
+gdb_test_multiple "info line $line_trap" "" {
+    -re "contains no code" {
+	set has_line_table_bug 1
+	exp_continue
+    }
+    -re -wrap "" {
+	pass $gdb_test_name
+    }
+}
+
+gdb_test_multiple "next" "step to trap line" {
+    -re -wrap "SIGILL.*" {
+	if {$has_line_table_bug} {
+	    xfail "$gdb_test_name (compiler line table bug)"
+	} else {
+	    fail "$gdb_test_name"
+	}
+    }
+    -re -wrap "__builtin_verbose_trap.*" {
+	pass $gdb_test_name
+
+	# Next step should trigger the trap.
+	gdb_test "next" "SIGILL.*" "step triggers trap"
+    }
+}
--- a/gdb/testsuite/gdb.base/builtin_verbose_trap.exp
+++ b/gdb/testsuite/gdb.base/builtin_verbose_trap.exp
@@ -1,0 +1,60 @@
+# Copyright (C) 2026 Free Software Foundation, Inc.
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test that __builtin_verbose_trap inline frames are shown when trap fires.
+# __builtin_verbose_trap is a Clang-only builtin, added in Clang 17.
+
+require {expr {[test_compiler_info clang*] && ![test_compiler_info {clang-1[0-6]-*}]}}
+
+standard_testfile .cpp
+
+if {[prepare_for_testing "failed to prepare" $testfile $srcfile {debug c++}]} {
+    return
+}
+
+if {![runto_main]} {
+    return
+}
+
+# Set breakpoint in function before inline function call.
+gdb_breakpoint "test_trap_function"
+
+# Continue to breakpoint.
+gdb_test "continue" \
+    ".*Breakpoint.*test_trap_function.*" \
+    "hit breakpoint in function"
+
+# Step through inline function calls with next.
+# The inline frames for add_numbers should be hidden.
+gdb_test "next" ".*" "next 1"
+gdb_test "next" ".*" "next 2"
+
+# Verify backtrace does NOT show add_numbers inline frame.
+gdb_test "bt" \
+    ".*test_trap_function.*" \
+    "inline frame not shown during stepping"
+
+gdb_test_no_output "set confirm off"
+
+# Continue to trigger the trap.
+# CPU uses ud2 instruction which generates SIGILL (not SIGABRT like GPU).
+gdb_test "continue" \
+    ".*received signal SIGILL.*" \
+    "received trap signal"
+
+# Verify verbose trap message appears in backtrace.
+gdb_test "bt" \
+    ".*This is verbose trap.*" \
+    "verbose trap message appears in backtrace"
--- a/gdb/testsuite/gdb.rocm/builtin_verbose_trap.cpp
+++ b/gdb/testsuite/gdb.rocm/builtin_verbose_trap.cpp
@@ -1,0 +1,34 @@
+/* Copyright (C) 2026 Free Software Foundation, Inc.
+   Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+   This file is part of GDB.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <hip/hip_runtime.h>
+
+__global__ void
+test_trap_kernel ()
+{
+  int x = 1;
+  __builtin_verbose_trap ("check verbose", "This is verbose trap!");
+}
+
+int
+main ()
+{
+  test_trap_kernel<<<1, 1>>> ();
+  return hipDeviceSynchronize () != hipSuccess;
+}
--- a/gdb/testsuite/gdb.rocm/builtin_verbose_trap.cpp
+++ b/gdb/testsuite/gdb.rocm/builtin_verbose_trap.cpp
@@ -1,0 +1,44 @@
+/* Copyright (C) 2026 Free Software Foundation, Inc.
+   Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+   This file is part of GDB.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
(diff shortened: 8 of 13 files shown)


gdb/amd-dbgapi-target.c
```diff
@@ -2076,14 +2076,22 @@ process_one_event (amd_dbgapi_inferior_info &info,
...
-	       breakpoint instruction.
+	    /* If the wave is stopped because of a software breakpoint or an abort
+	       trap (s_trap 2, used by __builtin_verbose_trap), the program counter
+	       needs to be adjusted so that it points to the trap instruction.
+
+	       For abort traps, this ensures the PC remains in the inlined frame
+	       containing DWARF debug info for verbose trap messages.
+
+	       Do not adjust PC for abort traps when single-stepping, as stepping
+	       has its own PC management logic.
```
lancesix: The real question is: should we allow to single step past an abort?  Is __builtin_verbose_trap marked noreturn?

gdb/testsuite/gdb.rocm/builtin_trap.exp
```diff
@@ -0,0 +1,30 @@
+# Test that __builtin_trap PC is correctly adjusted
```
lancesix: The file should have a copyright notice.

gdb/testsuite/gdb.rocm/builtin_trap.exp
```diff
@@ -0,0 +1,30 @@
...
+    gdb_test "continue" \
+	".*received signal SIGABRT.*" \
+	"received trap signal"
+
+    # Check PC points to trap instruction
+    # If PC adjustment is working, PC should be at the s_trap instruction
+    gdb_test "x/i \$pc" \
+	".*s_trap.*" \
+	"PC points to trap instruction"
+}
```
lancesix: If the core of GDB has some logic for single stepping & s_trap, we should probably test for it.

gdb/amd-dbgapi-target.c
```diff
@@ -2076,14 +2076,22 @@ process_one_event (amd_dbgapi_inferior_info &info,
 	    if (thread == nullptr)
 	      thread = add_gpu_thread (info.inf, event_ptid);
 
-	    /* If the wave is stopped because of a software breakpoint, the
-	       program counter needs to be adjusted so that it points to the
-	       breakpoint instruction.
+	    /* If the wave is stopped because of a software breakpoint or an abort
+	       trap (s_trap 2, used by __builtin_verbose_trap), the program counter
```
lancesix, line 2080: The test is not exercising __builtin_verbose_trap, it probably should.

gdb/testsuite/gdb.rocm/builtin_trap.cpp
```diff
@@ -0,0 +1,12 @@
+#include <hip/hip_runtime.h>
```
lancesix: Missing copyright header.

gdb/testsuite/gdb.rocm/builtin_trap.cpp
```diff
@@ -0,0 +1,12 @@
...
+__global__ void test_trap_kernel()
+{
+    __builtin_trap();  // s_trap 2
+}
+
+int main()
+{
+    test_trap_kernel<<<1, 1>>>();
+    return hipDeviceSynchronize() != hipSuccess ? 1 : 0;
+}
```
lancesix: Coding style.

Should be

```c++
__global__ void
test_trap_kernel ()
{
  __builtin_trap ();
}

int
main ()
{
 test_trap_kernel<<<1, 1>>> ();
  return hipDeviceSynchronize () != hipSuccess;
}
```

gdb/testsuite/gdb.rocm/builtin_trap.cpp
```diff
@@ -0,0 +1,12 @@
...
+
+__global__ void test_trap_kernel()
+{
+    __builtin_trap();  // s_trap 2
+}
+
+int main()
+{
+    test_trap_kernel<<<1, 1>>>();
+    return hipDeviceSynchronize() != hipSuccess ? 1 : 0;
```
lancesix: not a big fan of the `(foo ? 1 : 0)`.  The 1 if true 0 otherwise is tautologic, this is how `foo` would convert to int anyway.

gdb/testsuite/gdb.rocm/builtin_trap.exp
```diff
@@ -0,0 +1,30 @@
...
+clean_restart
+gdb_load $::binfile
+
+with_rocm_gpu_lock {
+    if {![runto_main]} {
+	fail "cannot run to main"
+	return
+    }
+
+    # Continue to the trap
```
lancesix: Missing period at the end of the sentence.

gdb/testsuite/gdb.rocm/builtin_trap.exp
```diff
@@ -0,0 +1,30 @@
...
+    gdb_test "continue" \
+	".*received signal SIGABRT.*" \
+	"received trap signal"
+
+    # Check PC points to trap instruction
+    # If PC adjustment is working, PC should be at the s_trap instruction
```
lancesix: dito.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.cpp
```diff
@@ -0,0 +1,30 @@
+/* Copyright (C) 2021-2026 Advanced Micro Devices, Inc. All rights reserved.
```
aktemur: Did you mean 2026 only?

gdb/testsuite/gdb.rocm/builtin_verbose_trap.cpp
```diff
@@ -0,0 +1,30 @@
...
+__global__ void test_trap_kernel()
+{
+    __builtin_verbose_trap("check verbose", "This is verbose trap!");
+}
+
+int main()
+{
+    test_trap_kernel<<<1, 1>>>();
+    return hipDeviceSynchronize() != hipSuccess;
+}
```
aktemur: Let's please apply GNU style.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,47 @@
+# Copyright (C) 2021-2026 Advanced Micro Devices, Inc. All rights reserved.
```
aktemur: Did you mean 2026?

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,47 @@
...
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test that __builtin_trap PC is correctly adjusted
```
aktemur: End with a period.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,47 @@
...
+load_lib rocm.exp
+require allow_hip_tests
+standard_testfile .cpp
+
+if {[build_executable "failed to build" $testfile $srcfile {debug hip}]} {
+    return
+}
+
+clean_restart
+gdb_load $::binfile
```
aktemur: Nit: `clean_restart $::testfile`

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,47 @@
...
+
+with_rocm_gpu_lock {
+    if {![runto_main]} {
+	fail "cannot run to main"
```
aktemur: `runto_main` already emits a FAIL. Let's please remove `fail` here,

aktemur (commented):
Please squash the second patch to the second, so that the test accompanies the fix.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
lancesix: The test is now using __builtin_verbose_trap, this test is about that.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,47 @@
...
+    # Continue to the trap.
+    gdb_test "continue" \
+	".*received signal SIGABRT.*" \
+	"received trap signal"
+
+    # Check PC points to trap instruction.
+    # If PC adjustment is working, PC should be at the s_trap instruction.
+    gdb_test "x/i \$pc" \
+	".*s_trap.*" \
+	"PC points to trap instruction"
```
lancesix, line 65: The main thing we'd want to test here is wether the current frame gives us the trap, i.e. does your message show un `bt` or similar message.

The expectation is to see that in the message with the SIGABRT signal, saying where we stopped.

gdb/amd64-linux-tdep.c
```diff
@@ -2087,6 +2087,32 @@ amd64_linux_fetch_hiperr_parameters (frame_info_ptr frame)
 
   return amd64_fetch_hiperr_parameters (frame, struct_addr);
 }
+/* Return true if the inline frame FUNC should be shown when stopped
```
lumachad, line 2091: Missing blank line between the closing `}` of `amd64_linux_fetch_hiperr_parameters` and the `/* Return true if… */` comment. GDB style requires one blank line between top-level definitions.

gdb/amdgpu-tdep.c
```diff
@@ -2271,6 +2295,10 @@ amdgpu_gdbarch_init (struct gdbarch_info info, struct gdbarch_list *arches)
 
   set_gdbarch_decr_pc_after_break (gdbarch, pc_adjust);
 
+  /* Register verbose trap inline frame handler.  */
+  set_gdbarch_show_verbose_trap_inline_frame
```
lumachad, line 2298: `/* Register verbose trap inline frame handler.  */` is obvious from the `set_gdbarch_` call itself. Drop it.

gdb/arch-utils.c
```diff
@@ -1613,6 +1613,19 @@ core_file_exec_context::environment () const
...
 
+/* See arch-utils.h.  */
+
+bool
+default_show_verbose_trap_inline_frame (struct gdbarch *gdbarch,
+					struct symbol *func,
+					enum gdb_signal stop_signal)
+{
+  /* By default, don't show verbose trap inline frames.
+     This maintains backward compatibility with existing behavior.
```
lumachad: Three sentences for a function that just returns `false` is more than needed. Trim to one, e.g.:

```c
/* By default, do not show verbose trap inline frames.  */
```

gdb/inline-frame.c
```diff
@@ -428,10 +429,15 @@ skip_inline_frames (thread_info *thread, bpstat *stop_chain)
...
+  enum gdb_signal stop_signal = thread->stop_signal ();
+
   for (const auto sym : function_symbols)
     {
       if (stopped_by_user_bp_inline_frame (sym, stop_chain)
-	  || sym == function_symbols.back ())
+	  || sym == function_symbols.back ()
+	  || gdbarch_show_verbose_trap_inline_frame
+	       (gdbarch, const_cast<symbol *> (sym), stop_signal))
 	break;
```
lumachad, line 440: This `const_cast<symbol *> (sym)` is a bit awkward. Should we just make the hook accept whatever it is we're passing here?

gdb/testsuite/gdb.rocm/builtin_verbose_trap.cpp
```diff
@@ -0,0 +1,32 @@
+/* Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+   This file is part of GDB.
```
lumachad, line 4: The copyright header is missing the `Free Software Foundation, Inc.` line required by GDB's contribution policy. Compare with `gdb/testsuite/gdb.rocm/simple.cpp` which starts with `Copyright 2022-2026 Free Software Foundation, Inc.` followed by the AMD copyright. Please add the FSF line first.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,50 @@
+# Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+# This file is part of GDB.
```
lumachad, line 4: Same missing FSF copyright line as in the `.cpp` file. Please add `# Copyright 2026 Free Software Foundation, Inc.` as the first copyright line.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,50 @@
...
+
+with_rocm_gpu_lock {
+    if {![runto_main]} {
+	return
+    }
+
+    # Continue to the trap and verify verbose message appears.
+    gdb_test "continue" \
+	".*received signal SIGABRT.*This is verbose trap.*" \
+	"received trap signal with verbose message"
```
lumachad: In Tcl/expect, `.` does not match newlines in `-re` mode. GDB outputs `received signal SIGABRT` and `This is verbose trap!` on separate lines, so the pattern `".*received signal SIGABRT.*This is verbose trap.*"` will never match. Use two separate `gdb_test` calls, or use `gdb_test_multiple` with `exp_continue` to match successive lines.

lumachad (dismissed):

gdb/amd64-linux-tdep.c
amd-bfilipov: Done

gdb/amdgpu-tdep.c
amd-bfilipov: dropped

gdb/inline-frame.c
amd-bfilipov: Changed hook signature to accept const struct symbol * instead of struct symbol

lancesix (commented):
The test should exercise the stepping as well.  And since you add host support, I suppose this should have a test as well.

amd-bfilipov:
> The test should exercise the stepping as well. And since you add host support, I suppose this should have a test as well.

Ok. Added another inline function to the test, so the test steps over it and checks whether regular inline frames stay hidden. 
Added a cpu version of the test. It's in a separate commit but we can rearrange that if needed.

jhuber6:
I'm not intimately familiar with gdb, but the `lldb` output seems much more descriptive here. Any chance we can do something similar? From the example with `-fsanitize-runtime=trap -fsanitize=undefined` in Clang.

```
(lldb) run
Process 302532 launched: 'llvm-project/build/a.out' (x86_64)
Process 302532 stopped
* thread #1, name = 'a.out', stop reason = Undefined Behavior Sanitizer: signed integer addition overflow in 'x + 1'
       frame #1: 0x0000555555555750 a.out`signed_overflow(x=2147483647) at ubsan_test.c:4:39
   1     #include <limits.h>
   2     #include <stdio.h>
   3     
-> 4     int signed_overflow(int x) { return x + 1; }
   5     
   6     int shift_oob(int x, int amount) { return x << amount; }
   7     
```

gdb/amd-dbgapi-target.c
```diff
@@ -2076,14 +2076,18 @@ process_one_event (amd_dbgapi_inferior_info &info,
...
+
+	       For abort traps, this ensures the PC remains in the inlined frame
+	       containing DWARF debug info for verbose trap messages.
 
 	       When dealing with a corefile, it is expected that the PC has
 	       been adjusted before generating the corefile, so no need to
 	       re-do it now.  */
-	    if ((stop_reason & AMD_DBGAPI_WAVE_STOP_REASON_BREAKPOINT) != 0
+	    if (((stop_reason & AMD_DBGAPI_WAVE_STOP_REASON_BREAKPOINT) != 0
+		|| (stop_reason & AMD_DBGAPI_WAVE_STOP_REASON_ASSERT_TRAP) != 0)
```
lancesix, line 2090: FYI, I am not sure this should be the way to go.

I have https://github.com/ROCm/rocm-systems/pull/9776 to change the trap handler behaviour so this is not necessary.  Such change would fix other issues reported independent to this PR.

palves (commented):
Re. the "gdb: Add gdbarch hook to show verbose trap inline frames" patch:

The commit log needs to be turned upside down.

This adds the gdbarch hook, yes, but it needs to begin by stating what the problem is, talk about __builtin_verbose_trap, show what goes wrong today without the fix.

I.e., establish rationale for why a change is needed.

Explain that this affects both CPU and GPU.

Then state how we're fixing this.  The new gdbarch hook.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
palves: @lumachad That is not true.  There is nothing special about newlines, other than being more than one character (\r\n), so a single `'.'` will not match it fully.  But `'.*'` definitely does.

lumachad:
Dropping to a draft since no action in a while.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
amd-bfilipov: Done. It's shown with `bt`

amd-bfilipov:
> The test should exercise the stepping as well. And since you add host support, I suppose this should have a test as well.

Added single stepping to the test.

gdb/amd-dbgapi-target.c
amd-bfilipov: Integrated the changes with https://github.com/ROCm/rocm-systems/pull/9776 and with https://sourceware.org/pipermail/gdb-patches/2026-August/229200.html

This means that this PR only needs to show inline frames when dealing with `builtin_verbose_trap`.
We can now also single-step into it without hanging the program.

amd-bfilipov:
> Re. the "gdb: Add gdbarch hook to show verbose trap inline frames" patch:
> 
> The commit log needs to be turned upside down.
> 
> This adds the gdbarch hook, yes, but it needs to begin by stating what the problem is, talk about __builtin_verbose_trap, show what goes wrong today without the fix.
> 
> I.e., establish rationale for why a change is needed.
> 
> Explain that this affects both CPU and GPU.
> 
> Then state how we're fixing this. The new gdbarch hook.

Squashed the test commit with the commit that introduces the changes, per Baris' suggestion. Dropped the PC adjust commit since Lancelots patch solves that. That leaves us with a single commit, and message for that commit has been changed according to your suggestion.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,90 @@
...
+
+load_lib rocm.exp
+require allow_hip_tests
+standard_testfile .cpp
+
+if {[build_executable "failed to build" $testfile $srcfile {debug hip}]} {
+    return
+}
+
+clean_restart $::testfile
```
aktemur: Let's please use `prepare_for_testing` so tthat we don't need `clean_restart`.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,90 @@
...
+	return
+    }
+
+    # Set breakpoint in kernel before inline function call.
+    gdb_breakpoint "test_trap_kernel" -allow-pending
+
+    # Continue to breakpoint.
+    gdb_test "continue" \
+	".*Breakpoint.*test_trap_kernel.*" \
+	"hit breakpoint in kernel"
```
aktemur: Nit: These can be simplified to
```
if {[runto test_trap_kernel -allow-pending]} {
    return
}
```

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,90 @@
...
+    gdb_test "bt" \
+	".*This is verbose trap.*" \
+	"verbose trap message appears in backtrace"
+
+    # Test single stepping into the trap.
+    if {![runto_main]} {
+	return
+    }
+
+    gdb_breakpoint "builtin_verbose_trap.cpp:34" -allow-pending
```
aktemur: No hard-coded line numbers, please.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,90 @@
...
+
+    gdb_test "continue" \
+	".*Breakpoint.*add_numbers.*" \
+	"hit breakpoint before trap for stepping test"
+
+    gdb_test "next" \
+	".*received signal SIGABRT.*" \
+	"next triggers trap signal"
```
aktemur: I don't understand this. If we do a "next" while at line 34, we should reach line 35 but not execute it yet, meaning the abort should not have triggered. The signal we would get there wouldn't be ABRT, right?

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,79 @@
...
+    return
+}
+
+# Set breakpoint in function before inline function call.
+gdb_breakpoint "test_trap_function"
+
+# Continue to breakpoint.
+gdb_test "continue" \
+    ".*Breakpoint.*test_trap_function.*" \
+    "hit breakpoint in function"
```
aktemur: Again, `runto test_trap_function` would be a simpler alternative.

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,79 @@
...
+gdb_test "bt" \
+    ".*This is verbose trap.*" \
+    "verbose trap message appears in backtrace"
+
+# Test single stepping into the trap.
+if {![runto_main]} {
+    return
+}
+
+gdb_breakpoint "builtin_verbose_trap.cpp:29"
```
aktemur: No hard-coded line numbers, please.

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,79 @@
...
+
+gdb_test "continue" \
+    ".*Breakpoint.*add_numbers.*" \
+    "hit breakpoint before trap for stepping test"
+
+gdb_test "next" \
+    ".*received signal SIGILL.*" \
+    "next triggers trap signal"
```
aktemur: If I do a "next" at line 29, shouldn't I reach line 30 normally via stepping? I wouldn't expect to see SIGILL there, because I haven't executed line 30, yet.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
amd-bfilipov: Done. Changed to prepare_for_testing

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
amd-bfilipov: Done. Simplified to if {![runto test_trap_kernel -allow-pending]} { return }

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
amd-bfilipov: Done. Removed all hardcoded line numbers. The stepping test now uses runto test_trap_kernel followed by
  next commands with generic .* patterns

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
amd-bfilipov: Added a comment explaining this. Due to DWARF line table mapping, the trap instruction gets associated
  with the line before it (compiler optimization/inlining). So when stepping with next, the trap triggers
  immediately because the s_trap 2 instruction is mapped to the current line, not the next line.

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
amd-bfilipov: Done

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
amd-bfilipov: Done

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
amd-bfilipov: Same as GPU - added comment explaining DWARF line table mapping. The trap instruction is associated with
  the line before it due to inlining, so next triggers the trap immediately

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,62 @@
...
+
+# Test single stepping to the trap line.
+# Expected: step should stop at __builtin_verbose_trap line, then
+# next step triggers the trap.
+# Due to a compiler line table issue, the trap may fire immediately.
+clean_restart $testfile
+
+if {![runto test_trap_function]} {
+    return
+}
```
aktemur: Sorry for the back and forth, but here we shall use run to breakpoint. Running to the function may not necessarily stop at the "int x = 1;" line. It may stop as "{", depending on the line table. So, we should rather define a bp at "int x = 1;" and run to it, and then do "next".

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,62 @@
...
+
+gdb_test_multiple "next" "step to trap line" {
+    -re -wrap ".*SIGILL.*" {
+	xfail "$gdb_test_name (compiler line table issue)"
```
aktemur: Ideally we should put the bug number.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,72 @@
...
+if {[prepare_for_testing "failed to prepare" $testfile $srcfile {debug hip}]} {
+    return
+}
+
+with_rocm_gpu_lock {
+    if {![runto test_trap_kernel -allow-pending]} {
+	return
+    }
+
+    gdb_test_no_output "set confirm off"
```
aktemur: Why is this needed?

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,62 @@
...
+
+# Verify verbose trap message appears in backtrace.
+gdb_test "bt" \
+    ".*This is verbose trap.*" \
+    "verbose trap message appears in backtrace"
+
+# Test single stepping to the trap line.
+# Expected: step should stop at __builtin_verbose_trap line, then
+# next step triggers the trap.
+# Due to a compiler line table issue, the trap may fire immediately.
```
aktemur: We can remove "Due to a compiler ....". This is covered by xfail below.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,72 @@
...
+
+    # Verify verbose trap message appears in backtrace.
+    gdb_test "bt" \
+	".*This is verbose trap.*" \
+	"verbose trap message appears in backtrace"
+
+    # Test single stepping to the trap line.
+    # Expected: step should stop at __builtin_verbose_trap line, then
+    # next step triggers the trap.
+    # Due to a compiler line table issue, the trap may fire immediately.
```
aktemur: We can remove "Due to a compiler...". It's covered by the xfail below.

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,62 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test that __builtin_verbose_trap inline frames are shown when trap fires.
+# __builtin_verbose_trap is a Clang-only builtin, added in Clang 17.
+
+require {expr {[test_compiler_info clang*] && ![test_compiler_info {clang-1[0-6]-*}]}}
```
palves: The comment says Clang 17, and compiler check matches (though it fails to consider clang 9 and lower), but I tried this on ubuntu 24.04, which has:

```
$ clang -v
Ubuntu clang version 18.1.3 (1ubuntu1)
```

And got:

```
$ make check RUNTESTFLAGS="CC_FOR_TARGET=clang CXX_FOR_TARGET=clang++" TESTS="gdb.base/builtin_verbose_trap.exp"
...
get_compiler_info: clang-18-1-3
gdb_do_cache: get_compiler_info_1 ( c++ )
get_compiler_info: clang-18-1-3
Executing on host: clang++    -fdiagnostics-color=never -Wno-unknown-warning-option  -c -g -o /home/pedro/rocm/gdb/build/gdb/testsuite/outputs/gdb.base/builtin_verbose_trap/builtin_verbose_trap0.o /home/pedro/rocm/gdb/build/gdb/testsuite/../../../src/gdb/testsuite/gdb.base/builtin_verbose_trap.cpp    (timeout = 300)
builtin_spawn -ignore SIGHUP clang++ -fdiagnostics-color=never -Wno-unknown-warning-option -c -g -o /home/pedro/rocm/gdb/build/gdb/testsuite/outputs/gdb.base/builtin_verbose_trap/builtin_verbose_trap0.o /home/pedro/rocm/gdb/build/gdb/testsuite/../../../src/gdb/testsuite/gdb.base/builtin_verbose_trap.cpp
/home/pedro/rocm/gdb/build/gdb/testsuite/../../../src/gdb/testsuite/gdb.base/builtin_verbose_trap.cpp:21:3: error: use of undeclared identifier '__builtin_verbose_trap'
   21 |   __builtin_verbose_trap ("check verbose", "This is verbose trap!");
      |   ^
1 error generated.
compiler exited with status 1
gdb compile failed, /home/pedro/rocm/gdb/build/gdb/testsuite/../../../src/gdb/testsuite/gdb.base/builtin_verbose_trap.cpp:21:3: error: use of undeclared identifier '__builtin_verbose_trap'
   21 |   __builtin_verbose_trap ("check verbose", "This is verbose trap!");
      |   ^
1 error generated.
UNTESTED: gdb.base/builtin_verbose_trap.exp: failed to prepare
testcase /home/pedro/rocm/gdb/build/gdb/testsuite/../../../src/gdb/testsuite/gdb.base/builtin_verbose_trap.exp completed in 0 seconds

                === gdb Summary ===

# of untested testcases         1
```

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,62 @@
...
+}
+
+if {![runto test_trap_function]} {
+    return
+}
+
+# Continue to trigger the trap.
+# CPU uses ud2 instruction which generates SIGILL (not SIGABRT like GPU).
+gdb_test "continue" \
+    ".*received signal SIGILL.*" \
```
palves: The leading ".*" in the regexp passed to gdb_test, and gdb_test_multiple, is unnecessary, it is implied.  (Many places in the testcase have this, I won't point at them all.)

gdb/gdbarch_components.py
```diff
@@ -2952,3 +2952,26 @@
...
+However, for verbose trap scenarios, the inline frame contains important
+diagnostic information that should be visible in backtraces.
+
+Architectures can use this hook to detect verbose trap frames by checking
+both the symbol name (e.g., starts with "__clang_trap_msg$") and the stop
+signal (e.g., GDB_SIGNAL_ABRT for abort traps), returning true only when
+both conditions indicate a verbose trap scenario.
+""",
+    type="bool",
+    name="show_verbose_trap_inline_frame",
```
aktemur: Nit: Based on how this gdbarch method is used and implemented, it makes more sense to me to name it "is_verbose_trap_frame".

gdb/gdbarch_components.py
```diff
@@ -2952,3 +2952,26 @@
     predefault="default_get_shadow_stack_pointer",
     invalid=False,
 )
+
+Method(
+    comment="""
+Return true if the inline frame represented by FUNC should NOT be skipped
+when stopped due to STOP_SIGNAL.  This allows architectures to show
+compiler-generated inline frames that contain verbose trap messages
+(e.g., __builtin_verbose_trap).
```
palves: It seems to me that the hook's name and description is a bit more specific than necessary.

The hook is called to decide whether to skip inline frames.

The implementation should then be able to use it for verbose trap.  But it could be used for other things.

We can already see a little bit of awkwardness in the semantics, because the answer to "show verbose trap inline frame?" is always "YES!".

The real question the hook is asking is "show THIS inline frame?", and then the answer is "YES, IF it's a verbose trap frame."

aktemur:
Please include this at the end of the commit message:
```
Bug: AIROCGDB-558
```

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
amd-bfilipov: Fixed. Changed to:
  1. runto_main
  2. Set breakpoint at "int x = 1" line
  3. Continue to that breakpoint
  4. Proceed with stepping tests

  This avoids ambiguity about which line in test_trap_function to stop at.

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
amd-bfilipov: Response:
  Fixed. Added comprehensive 7-line explanatory paragraph at the beginning of both test files explaining:
  - What the compiler line table bug is
  - How it manifests (trap instruction associated with wrong source line)
  - Why it causes the observed test behavior
  - How tests handle it (xfail)

  The xfail now references "compiler line table bug" without a specific bug number, since the explanation paragraph
  provides full context.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
amd-bfilipov: Fixed. Removed gdb_test_no_output

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
amd-bfilipov: Fixed. Removed inline compiler bug comment from gdb.base test (covered by explanation paragraph).

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
amd-bfilipov: done

gdb/gdbarch_components.py
amd-bfilipov: Fixed, but used Pedro's more general-purpose suggestion: renamed to `should_show_inline_frame` instead.

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
amd-bfilipov: Test fails on Clang 18 - __builtin_verbose_trap doesn't exist until Clang 19. Fixed.

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
amd-bfilipov: Fixed

gdb/gdbarch_components.py
amd-bfilipov: Fixed. Designed hook as fully general-purpose:

  Hook name: should_show_inline_frame (generic, not verbose-trap-specific)

  Documentation describes generic purpose: "Determine whether an inline frame should be shown to the user." Uses
  verbose trap as example use case, not requirement.

  Signature: (const struct symbol *func, enum gdb_signal stop_signal)
  - Generic parameters suitable for other scenarios
  - Not tied to verbose trap specifics

  Default returns false (hide inline frames during stepping, existing behavior)
  Architectures can override with their own visibility logic.

amd-bfilipov:
> Please include this at the end of the commit message:
> 
> ```
> Bug: AIROCGDB-558
> ```

Fixed

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,93 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test that __builtin_verbose_trap inline frames are shown when trap fires.
+# __builtin_verbose_trap is a Clang-only builtin, added in Clang 19.
+
+require {expr {[test_compiler_info clang*] && ![test_compiler_info {clang-1[0-8]-*}] && ![test_compiler_info {clang-[1-9]-*}]}}
```
palves: Line too long.  Break it:

```
require {expr {[test_compiler_info clang*]
	       && ![test_compiler_info {clang-1[0-8]-*}]
	       && ![test_compiler_info {clang-[1-9]-*}]}}
```

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,93 @@
...
+if {[prepare_for_testing "failed to prepare" $testfile $srcfile {debug c++}]} {
+    return
+}
+
+if {![runto test_trap_function]} {
+    return
+}
+
+# Continue to trigger the trap.
+# CPU uses ud2 instruction which generates SIGILL (not SIGABRT like GPU).
```
palves: The testcase should stand alone without GPU references.  And it'll be run on all kinds of CPUs, don't assume CPU == x86-64.

Also, not all CPUs will generate SIGILL.  I pointed Claude at the LLVM codebase, and asked it to build me a table with what instruction __builtin_verbose_trap uses and what signal does it raise on all the supported architectures.  Here's what I got:

## llvm.trap lowering

| Target | Instruction | Host signal |
|--------|-------------|-------------|
| x86-64 / x86 | `ud2` | SIGILL |
| AArch64 | `brk #1` | SIGTRAP |
| ARM (A32/T32, v5T+) | `trap` (undefined-insn class) | SIGTRAP |
| ARM (no v5T) | `udf 254` | SIGILL |
| RISC-V | `unimp` | SIGILL |
| PowerPC | `trap` (`tw 31,r0,r0`) | SIGTRAP |
| MIPS | `break` | SIGTRAP |
| SPARC | `ta 5` | SIGTRAP |
| SystemZ | 4-byte trap alias | SIGILL |
| LoongArch | `ud 0` (alias `amswap.w`) | SIGILL |
| Hexagon | `PS_crash` pseudo | — |
| WebAssembly | `unreachable` (0x00) | trap (no POSIX signal) |
| AMDGPU | `s_trap 2` | queue error → SIGABRT |


So ignoring AMD GPU, it's either SIGILL or SIGTRAP.

For SIGTRAP archs, I think we'll need to teach GDB to distinguish between the trap actually executing vs a single step or other reasons for a SIGTRAP.  I'll need to think about how best to make that work.

For now, in the interest of time, I'm OK with going as is.  Just tweak the testcase comment.

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,93 @@
...
+gdb_breakpoint "$srcfile:$line_x"
+gdb_test "continue" "Breakpoint.*$line_x.*" "run to line $line_x"
+
+# Check for compiler line table bug using disassembly.
+# If ud2 appears under "int x = 1" line instead of the
+# __builtin_verbose_trap line, the bug exists.
+set has_line_table_bug 0
+
+gdb_test_multiple "disassemble /m test_trap_function" "" {
+    -re "$line_x\[^\r\n\]*\r\n(\[^\r\n\]*\r\n)*?.*ud2" {
```
palves: Consider what will need to happen to the testcase when someone registers the new gdbarch for another CPU.  With this approach, they'll have to come here and hardcode some other instruction.

Did you try instead to ask GDB about the line that should contain the trap, like Baris mentioned in the call earlier today?  E.g., I see this:

```
(gdb) info line 21
Line 21 of "/home/pedro/rocm/gdb/build/gdb/testsuite/../../../src/gdb/testsuite/gdb.base/builtin_verbose_trap.cpp"
   is at address 0x55555555576d <_Z18test_trap_functionv+13> but contains no code.
```

The signal is "but contains no code".  If the compiler put ud2 at line 21, we would not see that.  That seems like a general way to test this without having to hardcode specific instructions.

lumachad:
Tests seem to be failing.

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
amd-bfilipov: Fixed

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
amd-bfilipov: Fixed. Removed all architecture-specific references (ud2, x86-64, GPU, CPU) from comments in both CPU and GPU test files. Comments now use generic "trap instruction" terminology.

gdb/testsuite/gdb.base/builtin_verbose_trap.exp
amd-bfilipov: Fixed. Replaced architecture-specific disassembly checks (looking for ud2/s_trap instructions) with portable info line approach in both CPU and GPU tests. Now checks for "contains no code" message to detect the compiler line table bug.

gdb/amd64-linux-tdep.c
```diff
@@ -2088,6 +2088,36 @@ amd64_linux_fetch_hiperr_info (frame_info_ptr frame)
   return amd64_fetch_hiperr_info (frame, struct_addr);
 }
 
+/* Return true if the inline frame FUNC should be shown when stopped
+   due to STOP_SIGNAL.  This allows showing verbose trap inline frames
+   on x86_64.  */
+
+/* Determine whether to show an inline frame.
+   Show verbose trap frames (__clang_trap_msg$...) when SIGILL occurs.  */
```
palves, line 2092: Two comment blocks for the same function?  Did you intend to drop the first one, or something like that?

gdb/amd64-linux-tdep.c
```diff
@@ -2154,6 +2184,10 @@ amd64_linux_init_abi_common (struct gdbarch_info info, struct gdbarch *gdbarch,
   /* Extract hiperr info for 'catch hiperr'.  */
   set_gdbarch_fetch_hiperr_info
     (gdbarch, amd64_linux_fetch_hiperr_info);
+
+  /* Show verbose trap inline frames when stopped due to abort.  */
```
palves: "stopped due to abort"

"Abort" doesn't apply for the CPU, does it?

gdb/amd64-linux-tdep.c
amd-bfilipov: fixed

gdb/amd64-linux-tdep.c
amd-bfilipov: fixed, for cpu its "illegal instruction"

palves:
@jhuber6 

> I'm not intimately familiar with gdb, but the `lldb` output seems much more descriptive here. Any chance we can do something similar? From the example with `-fsanitize-runtime=trap -fsanitize=undefined` in Clang.
> 
> ```
> (lldb) run
> Process 302532 launched: 'llvm-project/build/a.out' (x86_64)
> Process 302532 stopped
> * thread #1, name = 'a.out', stop reason = Undefined Behavior Sanitizer: signed integer addition overflow in 'x + 1'
>        frame #1: 0x0000555555555750 a.out`signed_overflow(x=2147483647) at ubsan_test.c:4:39
>    1     #include <limits.h>
>    2     #include <stdio.h>
>    3     
> -> 4     int signed_overflow(int x) { return x + 1; }
>    5     
>    6     int shift_oob(int x, int amount) { return x << amount; }
>    7     
> ```

Hi Joseph.  I'm not familiar with "-fsanitize-runtime=trap", and a google search is finding no hits, and neither my system clang nor the one that comes with my rocm build is aware of it.  Is that a new feature?  I don't suppose it's the same as "-fsanitize-trap=undefined"?  

Does it make the ubsan runtime use __builtin_verbose_trap?  __builtin_verbose_trap takes constant strings and embeds them in the generated inline function's name (i.e., compile time), so I'm curious to see how that works.

From your output, lldb seems to be hiding frame #0 (it presents a stop at frame #1), and decoding the stop reason out of somewhere.  If ubsan in that mode is really using __builtin_verbose_trap, then I guess lldb is probably extracting the arguments from the function name.

palves:
I found this https://lists.llvm.org/pipermail/lldb-commits/Week-of-Mon-20251117/155573.html, and the code seems to be in upstream ldlb already (in lldb/source/Plugins/InstrumentationRuntime/BoundsSafety/InstrumentationRuntimeBoundsSafety.cpp).

That talks about a "-fbounds-safety" option, and I _also_ don't have...  

https://github.com/swiftlang/llvm-project/pull/11645 describes it.

So does https://clang.llvm.org/docs/BoundsSafety.html, and then and https://clang.llvm.org/docs/BoundsSafetyImplPlans.html says that it's still not fully implemented.

So is this all a swift-llvm-fork-only feature?  But I'm confused since you used a different command line option.

In any case, afaict from the lldb code, it is extracting the stop reason from the (runtime) argument passed to  __bounds_safety_soft_trap, the magic function the ubsan runtime calls.

This is looking like should be a completely separate discussion.

gdb/testsuite/gdb.rocm/builtin_verbose_trap.exp
```diff
@@ -0,0 +1,97 @@
...
+    # the trap instruction is not mapped to that line (bug exists).
+    set has_line_table_bug 0
+    set line_trap [gdb_get_line_number "__builtin_verbose_trap"]
+
+    gdb_test_multiple "info line $line_trap" "" {
+	-re "contains no code" {
+	    set has_line_table_bug 1
+	    exp_continue
+	}
+	-re -wrap "" {
```
palves, line 78: This is missing a "pass $gdb_test_name" call.  (check the CPU version too).

amd-bfilipov: done

lumachad:
Labelling ci:skip here as we already had a green CI. We don't need to run it again for a trivial change. Once we have code/substantial changes, let's remove the ci:skip label and let it validate fully.

palves (approved):

palves:
@jhuber6 

Turns out I was doing the rookie mistake of not looking at current llvm/lldb master yesterday (I must have done git fetch instead of git pull).   I found the lldb code that knows about handling raw __builtin_verbose_frame.  It's at:

   lldb/source/Plugins/LanguageRuntime/CPlusPlus/VerboseTrapFrameRecognizer.cpp

and the extracting the stop reason is there too, in the VerboseTrapFrameRecognizer::RecognizeFrame call.

Reading the history of that file is interesting (with `git blame` and targeted hashes, `git log` doesn't show it all).  We'll continue discussing all this internally.

That still isn't the ubsan stuff though, that's still like I described earlier.  Grepping the llvm master tree for sanitize-runtime comes out empty.

PR #202, opened 2026-07-13
Add automerge workflow

lumachad:
Adds a GitHub Actions workflow that every 6 hours syncs sourceware/master into
origin/amd-staging. The workflow also triggers on closure of fast-forward
or conflict PRs to immediately start the next sync.

On a clean merge, a fast-forward PR is opened against amd-staging for CI
validation. On a conflict, a conflict PR is opened for manual resolution.
Only one PR of either type is allowed open at a time and further syncs are
postponed until it is resolved.

The fast-forward PR range is expressed as merge-base..last-clean-commit
so the title and compare URL are always meaningful even when only one
commit is being merged.

The script lives in .github/scripts/automerge.py and is documented in
.github/scripts/automerge.md.

### DIFF
--- a/.github/scripts/automerge.py
+++ b/.github/scripts/automerge.py
@@ -1,0 +1,556 @@
+#!/usr/bin/env python3
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+"""Sync ROCgdb sourceware/master into origin/amd-staging.
+
+Fetches the upstream sourceware master, fast-forwards origin/master to match,
+then merges cleanly into origin/amd-staging. On conflict, pushes a conflict
+branch and opens a PR for manual resolution.
+"""
+
+import argparse
+import json
+import os
+import shlex
+import subprocess
+import sys
+import time
+from datetime import date
+from pathlib import Path
+
+# ---------------------------------------------------------------------------
+# Configuration
+# ---------------------------------------------------------------------------
+
+ROCGDB_REPO = "ROCm/ROCgdb"
+ROCGDB_ORIGIN_URL = f"https://github.com/{ROCGDB_REPO}.git"
+SOURCEWARE_URL = "https://sourceware.org/git/binutils-gdb.git"
+
+
+def _parse_args() -> argparse.Namespace:
+    p = argparse.ArgumentParser(description="Sync sourceware/master into origin/amd-staging.")
+    p.add_argument(
+        "--workspace",
+        metavar="DIR",
+        help="Root directory for the working clone (overrides $WORKSPACE). "
+             "The clone is created at DIR/rocgdb_sync/binutils-gdb.",
+    )
+    p.add_argument(
+        "--dry-run",
+        action="store_true",
+        help="Fetch and probe as normal, but skip all pushes and PR creation.",
+    )
+    return p.parse_args()
+
+
+_args = _parse_args()
+DRY_RUN: bool = _args.dry_run
+_workspace = _args.workspace or os.environ.get("WORKSPACE", "")
+if _workspace:
+    _workspace_path = Path(_workspace).resolve()
+    if not _workspace_path.exists():
+        print(
+            f"Error: workspace '{_workspace}' is not a valid path.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+    WORK_DIR = _workspace_path / "rocgdb_sync"
+else:
+    WORK_DIR = Path.cwd() / "rocgdb_sync"
+
+# Never pass secrets as CLI args — use env vars (e.g. GH_TOKEN) instead.
+
+CONFLICT_BRANCH_PREFIX = "users/github/master-to-amd-staging-conflict"
+FF_BRANCH_PREFIX = "users/github/master-to-amd-staging-ff"
+CONFLICT_LABEL = "merge-conflict"
+FF_LABEL = "merge-testing"
+CI_SKIP_LABEL = "ci:skip"
+
+# Sleep durations (seconds) between successive network attempts.
+# Total attempts = len(RETRY_DELAYS) + 1.
+RETRY_DELAYS = [60, 120]
+
+
+# ---------------------------------------------------------------------------
+# Subprocess / git helpers
+# ---------------------------------------------------------------------------
+
+
+def run(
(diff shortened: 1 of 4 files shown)


.github/scripts/automerge.py
```diff
@@ -0,0 +1,487 @@
...
+            .stdout.strip()
+            .splitlines()
+        )
+        print("Conflicted files:")
+        for f in conflicted_files:
+            print(f"  {f}")
+
+        run_net(["git", "push", "origin", conflict_branch], cwd=repo)
+
+        pr_base = last_clean if last_clean else merge_base
```
spatrang: Execution only reaches here when `last_clean` is None (line 427 already returns whenever `last_clean` is set). So this is always `merge_base`. 

```suggestion
        pr_base = merge_base
```

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,37 @@
+name: GDB Automerge
+on:
+  schedule: # runs hourly
+    - cron: '0 */6 * * *'
```
spatrang, line 4: The comment on line 3 says "runs hourly" but `0 */6 * * *` runs every 6 hours. Change the comment to `# runs every 6 hours`.

.github/scripts/automerge.py
```diff
@@ -0,0 +1,487 @@
+#!/usr/bin/env python3
+"""Sync ROCgdb sourceware/master into origin/amd-staging.
+
+Fetches the upstream sourceware master, fast-forwards origin/master to match,
```
spatrang, line 6: Here "fast-forwards origin/master" - the script never touches origin/master

.github/scripts/automerge.py
```diff
@@ -0,0 +1,487 @@
...
+            "--base",
+            "amd-staging",
+            "--state",
+            "open",
+            "--label",
+            CONFLICT_LABEL,
+            "--json",
+            "url",
+            "--jq",
+            ".[0].url",
```
spatrang: `--jq ".[0].url"` returns `"null"`, so the gate always trips - automerge.py:203 and automerge.py:272
When no PR is open,` jq '.[0].url'` on `[]` prints the literal string null. So `find_open_conflict_pr()/find_open_ff_pr()` return ` "null"`, `main()` prints "Open … PR found: null" and returns early. **The script never merges anything**.

```suggestion
            "--jq", ".[0].url // empty",
```

.github/scripts/automerge.py
```diff
@@ -0,0 +1,487 @@
...
+
+        conflict_info = run(
+            ["git", "log", "-1", "--format=%H %s (%aN <%aE>)", first_conflict],
+            cwd=repo,
+        ).stdout.strip()
+        print(f"\nConflict commit: {conflict_info}")
+
+        # Checkout a fresh conflict branch from the current amd-staging state.
+        run(["git", "checkout", "-b", conflict_branch], cwd=repo)
+        run(["git", "merge", "--no-edit", first_conflict], cwd=repo, check=False)
```
spatrang: After this `git merge` fails, git stops mid-merge: conflict markers are written to the working tree but nothing is committed. The branch pushed on line 470 is therefore identical to `amd-staging`, so `open_conflict_pr()` → `gh pr create` fails with "No commits between amd-staging and …" and the run aborts. The conflict path never produces a PR.

We need to record the conflicted state as a commit before pushing. Note that `git commit` alone won't work here (it refuses with unmerged files), so we `git add -A` first to keep the markers in a committed diff a human can resolve.

```suggestion
        run(["git", "merge", "--no-edit", first_conflict], cwd=repo, check=False)
        # A conflicting merge leaves the tree unmerged with nothing committed.
        # Stage and commit the conflict markers so the branch carries a diff
        # that `gh pr create` can open (git refuses to commit unmerged paths).
        run(["git", "add", "-A"], cwd=repo)
        run(
            ["git", "commit", "--no-verify", "-m",
             f"Merge {first_conflict[:12]} into amd-staging (unresolved conflicts)"],
            cwd=repo,
        )
```

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,37 @@
...
+  workflow_dispatch:
+  pull_request:
+    types: [closed]
+    branches: [amd-staging]
+
```
spatrang, line 9: No concurrency guard - overlapping runs can open duplicate PRs.

Schedule + `workflow_dispatch` +` pull_request:closed` runs can overlap and both pass the PR gate (TOCTOU). 
Timeline of the race:

> 10:00:00 Run A calls find_open_ff_pr() → "none open."
> 10:00:01 Run B calls find_open_ff_pr() → also "none open" (A hasn't created its PR yet).
> 10:00:30 Run A creates a PR.
> 10:00:31 Run B creates a second PR.

Result: two PRs open at once, which breaks the workflow's "only one PR open at a time" rule - even though each run's check looked correct.

Add a top-level concurrency group:

```suggestion
concurrency:
  group: gdb-automerge
  cancel-in-progress: false
```

.github/scripts/automerge.py
lumachad: Done.

.github/workflows/automerge.yml
lumachad: Done.

.github/scripts/automerge.py
lumachad: Done. The script now fast-forwards `origin/master` to match `sourceware/master` on every run, right after the fetches.

.github/scripts/automerge.py
lumachad: Done. Changed to `.[0].url // empty` in both `find_open_conflict_pr()` and `find_open_ff_pr()`.

.github/scripts/automerge.py
lumachad: Done. Added `git add -A` followed by `git commit` to stage and record the conflict markers before pushing the conflict branch.

.github/workflows/automerge.yml
lumachad: Done. Added a top-level `concurrency` group (`gdb-automerge`, `cancel-in-progress: false`) to serialize workflow runs.

spatrang (approved):
LGTM

PR #206, opened 2026-07-14
handle clusters

aktemur:
## Motivation

New feature: workgroup clusters

## Technical Details

Workgroups can be organized in clusters. dbgapi provides new query functions. Use these queries to show cluster information.

## Test Plan

A basic test is included.

### DIFF
--- a/gdb/doc/gdb.texinfo
+++ b/gdb/doc/gdb.texinfo
@@ -24475,6 +24475,12 @@ work-items of the heterogeneous dispatch.  The number of dimensions displayed
 matches the dimensionality of the heterogeneous dispatch.
 
 @item
+The x, y, and z heterogeneous cluster dimensions, in that order, in terms of
+work-items of the heterogeneous dispatch.  The number of dimensions displayed
+matches the dimensionality of the heterogeneous dispatch.  If the architecture
+of the dispatch does not support clusters, a @code{-} is shown.
+
+@item
 The x, y, and z heterogeneous work-group dimensions, in that order, in terms of
 work-items of the heterogeneous dispatch.  The number of dimensions displayed
 matches the dimensionality of the heterogeneous dispatch.
@@ -24520,8 +24526,8 @@ For example,
 
 @smallexample
 (@value{GDBP}) info dispatches -full
-  Id   Target Id                      Grid      Workgroup Fence   Address Spaces          Kernel Descriptor  Kernel Args        Completion Signal  Kernel Function
-* 1    AMDGPU Dispatch 1:1:1 (PKID 0) [256,1,1] [128,1,1] B|As    Shared(0), Private(220) 0x00007ffde5409800 0x00007ffff7e00000 (nil)              bit_extract_kernel(unsigned int*, unsigned int const*, unsigned long)
+  Id   Target Id                      Grid      Cluster Workgroup Fence   Address Spaces          Kernel Descriptor  Kernel Args        Completion Signal  Kernel Function
+* 1    AMDGPU Dispatch 1:1:1 (PKID 0) [256,1,1] -       [128,1,1] B|As    Shared(0), Private(220) 0x00007ffde5409800 0x00007ffff7e00000 (nil)              bit_extract_kernel(unsigned int*, unsigned int const*, unsigned long)
 @end smallexample
 
 If you're debugging multiple inferiors, @value{GDBN} displays
@@ -29478,6 +29484,12 @@ AMDGPU Wave @var{agent-id}:@var{queue-id}:@var{dispatch-id}:@var{wave-id} (@var{
 @end smallexample
 
 It is used in the @samp{Target ID} column of the @samp{info threads} command.
+If the work-group is in a cluster, the cluster coordinates are shown, too,
+as shown below.
+
+@smallexample
+AMDGPU Wave @var{agent-id}:@var{queue-id}:@var{dispatch-id}:@var{wave-id} (@var{cluster-x},@var{cluster-y},@var{cluster-z})/(@var{work-group-x},@var{work-group-y},@var{work-group-z})/@var{work-group-thread-index}
+@end smallexample
 
 @item Lane Target ID
 The @acronym{AMD} @acronym{GPU} lane target identifier (@var{lane_systag})
@@ -29488,6 +29500,12 @@ AMDGPU Lane @var{agent-id}:@var{queue-id}:@var{dispatch-id}:@var{wave-id}/@var{l
 @end smallexample
 
 It is used in the @samp{Target ID} column of the @samp{info lanes} command.
+If the work-group is in a cluster, the cluster coordinates are shown, too,
+as shown below.
+
+@smallexample
+AMDGPU Lane @var{agent-id}:@var{queue-id}:@var{dispatch-id}:@var{wave-id}/@var{lane-index} (@var{cluster-x},@var{cluster-y},@var{cluster-z})/(@var{work-group-x},@var{work-group-y},@var{work-group-z})[@var{work-item-x},@var{work-item-y},@var{work-item-z}]
+@end smallexample
 
 @end table
 
@@ -29502,6 +29520,13 @@ variable has the following format:
 
 @smallexample
 (@var{work-group-x},@var{work-group-y},@var{work-group-z})/@var{work-group-thread-index}
+@end smallexample
+
+If the work-group is in a cluster, cluster coordinates are included, too,
+using the following format:
+
+@smallexample
+(@var{cluster-x},@var{cluster-y},@var{cluster-z})/(@var{work-group-x},@var{work-group-y},@var{work-group-z})/@var{work-group-thread-index}
 @end smallexample
 
 @item $_thread_workgroup_pos
@@ -29544,10 +29569,19 @@ identifier identifier respectively.  The identifiers are per process.
 The @acronym{AMD} @acronym{GPU} target driver packet identifier.  The
 identifier is per queue.
 
+@item cluster-x
+@itemx cluster-y
+@itemx cluster-z
+The grid position of the thread's work-group's cluster within the
+heterogeneous dispatch.  This position is not shown if the work-group
+does not belong to a cluster.
+
 @item work-group-x
 @itemx work-group-y
 @itemx work-group-z
-The grid position of the thread's work-group within the heterogeneous dispatch.
+The position of the thread's work-group within the heterogeneous
+dispatch, if the work-group belongs to a cluster.  Otherwise
+it is the position of the work-group within the grid.
 
 @item work-group-thread-index
 The thread's number within the heterogeneous work-group.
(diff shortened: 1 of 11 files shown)


aktemur:
Here is a sample output of "info threads" with clusters:
```
info threads
  Id   Target Id                                   Frame 
...
* 6    AMDGPU Wave 1:2:1:1 (0,0,0)/(0,0,0)/0       kernel () at ...
  7    AMDGPU Wave 1:2:1:2 (0,0,0)/(0,0,0)/1       kernel () at ...
...
  36   AMDGPU Wave 1:2:1:31 (2,0,1)/(0,0,0)/0      kernel () at ...
  37   AMDGPU Wave 1:2:1:32 (2,0,1)/(0,0,0)/1      kernel () at ...
...
  71   AMDGPU Wave 1:2:1:66 (1,0,0)/(1,0,0)/0      kernel () at ...
...
  95   AMDGPU Wave 1:2:1:90 (1,0,0)/(2,1,0)/4      kernel () at ...
```

And "info dispatches":
```
info dispatches
  Id   Target Id                      Grid       Cluster   Workgroup Fence   Kernel Function 
* 1    AMDGPU Dispatch 1:2:1 (PKID 1) [270,20,2] [90,10,1] [30,5,1]  B|As|Rs 0x00007ffff6084000 
```

aktemur:
Test is updated. It now includes synchronization and checks blockIdx dimensions as obtained from the program to those calculated based on cluster&wg coordinates.

gdb/doc/gdb.texinfo
```diff
@@ -29474,20 +29480,24 @@ The @acronym{AMD} @acronym{GPU} thread target identifier (@var{systag})
 string has the following format:
 
 @smallexample
-AMDGPU Wave @var{agent-id}:@var{queue-id}:@var{dispatch-id}:@var{wave-id} (@var{work-group-x},@var{work-group-y},@var{work-group-z})/@var{work-group-thread-index}
+AMDGPU Wave @var{agent-id}:@var{queue-id}:@var{dispatch-id}:@var{wave-id} (@var{cluster-x},@var{cluster-y},@var{cluster-z})/(@var{work-group-x},@var{work-group-y},@var{work-group-z})/@var{work-group-thread-index}
```
lancesix: just random thought.

When imagining this initially, I kind of assumed we would use something like `{cx,cy,cz}(x,y,z)/n` with cluster dimensions in another set of brackets (`{}`). Not to say I mind the `cluster/wg/wave` approach, I am not sure I have a strong opinion at this point, but wanted to drop this as a note.

gdb/doc/gdb.texinfo
```diff
@@ -29498,10 +29508,11 @@ convenience variables:
 
 @item $_dispatch_pos
 The string returned by the @code{$_dispatch_pos} debugger convenience
-variable has the following format:
+variable has the following format, where cluster coordinates are included
+only if the work-group is in a cluster.
 
 @smallexample
-(@var{work-group-x},@var{work-group-y},@var{work-group-z})/@var{work-group-thread-index}
+(@var{cluster-x},@var{cluster-y},@var{cluster-z})/(@var{work-group-x},@var{work-group-y},@var{work-group-z})/@var{work-group-thread-index}
```
lancesix, line 29529: I would probably split the examples in 2.  One cluster aware case and one "classic" case.  I would find this more "eye grabbing" when searching through the manual.

gdb/testsuite/gdb.rocm/cluster.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+  /* Index of the calling block within cluster.  */
+  auto block_in_cluster = c.block_index ();
+
+  int bx = blockIdx.x;
+  int by = blockIdx.y;
+  int bz = blockIdx.z;
+
+  /* Define a barrier to also test debuggability in the existence of
+     synchronization.  */
```
lancesix: the interesting debugability part here would be if some waves / workgroups reached the barrier when some other have not.  Seems that in this case, breaking everyone after the barrier limits the probabilities of having some waves before and some after the barrier.

aktemur: > the interesting debugability part here would be if some waves / workgroups reached the barrier when some other have not. Seems that in this case, breaking everyone after the barrier limits the probabilities of having some waves before and some after the barrier.

I don't fully follow. In principle, when we break at or before the barrier, there is always a possibility that some waves are behind. Only after the barrier we have the guarantee that all waves are now beyond the barrier. That's the purpose of having a barrier, no? 
I mean, if I break before the barrier and not all waves are there, why should that be a bug? It would be just a timing issue.

gdb/doc/gdb.texinfo
aktemur: > just random thought.
> 
> When imagining this initially, I kind of assumed we would use something like `{cx,cy,cz}(x,y,z)/n` with cluster dimensions in another set of brackets (`{}`). Not to say I mind the `cluster/wg/wave` approach, I am not sure I have a strong opinion at this point, but wanted to drop this as a note.

I have no objections to using braces.

gdb/testsuite/gdb.rocm/cluster.cpp
lancesix: The isa has a split barrier model, where waves need to notify when they arrive at the barrier, then wait on it.  The barrier has an internal state counting how many waves have signaled (s_barrier_signal), and will keep signaled waves from continuing (s_barrier_wait) until all have signaled.  During debug events, we trigger context save.  The main thing (which has had issues in the past) which could get wrong during context save is not saving / restoring the barrier state correctly.  This is what is tested by `gdb.rocm/device-barrier.exp` at the workgroup level.  I think my point is that while at introducing cluster concepts, it might be worth introducing a similar test exercising cluster barrier save and restore.

lancesix (commented):
also, shouldn't we have some configury stuff to check dbgapi's version to be sure `AMD_DBGAPI_DISPATCH_INFO_CLUSTER_SIZES` is provided?

gdb/testsuite/gdb.rocm/cluster.cpp
aktemur: Ok, thanks for the clarification. The cluster API has `barrier_arrive` and `barrier_wait` separately, for which there is a comment for `sync`:
```
// Sync the cluster, equivalent to c.barrier_wait(c.barrier_arrive());
```

How about having `barrier_arrive` and `barrier_wait`, with a breakpoint in-between? That sounds like it should satisfy the previously problematic scenario.

aktemur:
> also, shouldn't we have some configury stuff to check dbgapi's version to be sure `AMD_DBGAPI_DISPATCH_INFO_CLUSTER_SIZES` is provided?

In nowhere else we have a similar check. I thought we'd first merge dbgapi PR, which bumps the version, and then build ROCgdb with that version, which brings us the guard here:
```
  if (major != AMD_DBGAPI_VERSION_MAJOR || minor < AMD_DBGAPI_VERSION_MINOR)
    error (_("amd-dbgapi library version mismatch, got %d.%d.%d, need %d.%d+"),
           major, minor, patch, AMD_DBGAPI_VERSION_MAJOR,
           AMD_DBGAPI_VERSION_MINOR);
```

gdb/testsuite/gdb.rocm/cluster.cpp
lancesix: A good test would probably test the following cases:
- some of the waves have called arrived, and are now waiting (or about to)
- all the waves have called arrived, but none have called wait
- all have called arrived, some have called wait (so might be past the barrier)

but maybe it feels this particular matrix would be in a different test as checking the workgroup cluster dimensions.

lancesix:
> > also, shouldn't we have some configury stuff to check dbgapi's version to be sure `AMD_DBGAPI_DISPATCH_INFO_CLUSTER_SIZES` is provided?
> 
> In nowhere else we have a similar check. I thought we'd first merge dbgapi PR, which bumps the version, and then build ROCgdb with that version, which brings us the guard here:
> 
> ```
>   if (major != AMD_DBGAPI_VERSION_MAJOR || minor < AMD_DBGAPI_VERSION_MINOR)
>     error (_("amd-dbgapi library version mismatch, got %d.%d.%d, need %d.%d+"),
>            major, minor, patch, AMD_DBGAPI_VERSION_MAJOR,
>            AMD_DBGAPI_VERSION_MINOR);
> ```

We have https://github.com/ROCm/ROCgdb/blob/amd-staging/gdb/configure.ac#L330-L331 which currently guards building GDB with 0.80 or more.

Either we wait for dbgapi to land, and when we have the right version bump configure.ac in this PR, or have something more dynamic (we do not have this yet) where configure can test dbgapi for this feature (use `AC_COMPILE_IFELSE` to try to build a program including dbgapi and using this new API, and set a compile flag based on that).  I don't think I have a strong opinion either way, the `AC_COMPILE_IFELSE` could allow merging GDB first.

gdb/doc/gdb.texinfo
aktemur: > I would probably split the examples in 2. One cluster aware case and one "classic" case. I would find this more "eye grabbing" when searching through the manual.

I'm updating the document. Please check the next revision.

aktemur:
> Either we wait for dbgapi to land, and when we have the right version bump configure.ac in this PR, or have something more dynamic (we do not have this yet) where configure can test dbgapi for this feature (use `AC_COMPILE_IFELSE` to try to build a program including dbgapi and using this new API, and set a compile flag based on that). I don't think I have a strong opinion either way, the `AC_COMPILE_IFELSE` could allow merging GDB first.

Ok, gotcha. I'll go with incrementing the configure version in this PR and leave the dynamic detection to a separate feature task.

gdb/testsuite/gdb.rocm/cluster.cpp
aktemur: > but maybe it feels this particular matrix would be in a different test as checking the workgroup cluster dimensions.

Agree.

aktemur:
The last push updates the document, removes `c.sync ()` from the test (will be handled separately), bumps the version check in configure.

aktemur:
The last push converts in-grid group ids to in-cluster group ids.

gdb/doc/gdb.texinfo
aktemur: In an internal meeting we decided to use parentheses. I'm closing this discussion.

aktemur:
The force-push above is a rebase.

aktemur:
The force-push above uses `AMD_DBGAPI_WAVE_INFO_WORKGROUP_COORD_IN_CLUSTER` to get the wg coordinate within the cluster.

gdb/testsuite/gdb.rocm/cluster.cpp
aktemur: > The isa has a split barrier model, where waves need to notify when they arrive at the barrier, then wait on it.

I added a new test for this.

aktemur:
The update above adds a new test that uses barrier arrive/wait.

PR #208, opened 2026-07-14
.github/scripts: add --retry-ignored-tests switch to test_rocgdb.py

lumachad:
## Motivation

Tests on the ignore list are known to fail. When they fail (e.g. with a
timeout), the current retry logic re-runs them up to `--max-failed-retries`
times for each of the two compiler runs, adding significant wall-clock time
for no diagnostic value.

## Technical Details

Add a `--retry-ignored-tests` switch to `test_rocgdb.py` (default: off).

When off (the default), any test that appears in the ignore list
(`rocgdb_ignore_list.json`) and fails on the first run is excluded from
subsequent retry iterations.  If all remaining failures are on the ignore
list, iterations stop early.  Skipped tests are logged with a note pointing
to `--retry-ignored-tests`.

When `--retry-ignored-tests` is passed (or `--no-xfail` is active), the
previous behaviour is preserved: every failing test is retried regardless of
ignore-list membership.

The new flag is shown in the configuration summary printed at the start of
each run.

## Test Plan

- Manual inspection of the retry-filtering logic.
- Verified that `--retry-ignored-tests` and `--no-xfail` both bypass the
  filter correctly.

## Test Result

N/A (CI-script change only; no GPU hardware required to validate the logic).

## Submission Checklist

### DIFF
--- a/.github/scripts/test_rocgdb.py
+++ b/.github/scripts/test_rocgdb.py
@@ -372,6 +372,12 @@ def parse_arguments() -> argparse.Namespace:
         help="Do not use ignore lists for failed tests. Default is to use the ignore lists.",
     )
     parser.add_argument(
+        "--retry-ignored-tests",
+        action="store_true",
+        help="Retry tests that are on the ignore list when they fail. "
+        "Default is off: ignored tests that fail on the first run are not retried.",
+    )
+    parser.add_argument(
         "--optimization",
         type=str,
         default="",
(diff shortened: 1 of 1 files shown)


aktemur (commented):
I don't know this CI Python script much, but the patch looked reasonable.

lancesix (approved):
LGTM, thanks.

PR #209, opened 2026-07-14
.github: add automerge script and workflow for upstream sync

lumachad:
## Summary

- Make the automerge script fully generic: repo slug, target branch, and upstream branch are read from required repository variables (`AUTOMERGE_TARGET_BRANCH`, `AUTOMERGE_UPSTREAM_BRANCH`) instead of being hardcoded. The script exits with an error if any are unset.
- Push conflict-free branches by direct refspec to avoid local merge commits.
- Add `--dry-run` mode: fetches and probes normally but skips all pushes and PR creation, printing what it would do instead (branch name, commit range, PR title, and conflicted files if applicable).
- Wire `--dry-run` to a boolean `workflow_dispatch` input so it can be toggled from the Actions UI.
- Use "conflict-free" instead of "fast-forward" throughout: the script merges rather than pointer-advances, so the old term was misleading.
- Activate the workflow: remove the `if: false` guard and enable the schedule and `workflow_dispatch` triggers. The `pull_request` trigger has been dropped in favour of relying solely on the 6-hour schedule.

## Repository variables required before merge

- `AUTOMERGE_TARGET_BRANCH` — branch to merge into (e.g. `amd-staging`)
- `AUTOMERGE_UPSTREAM_BRANCH` — upstream branch to pull from (e.g. `master`)

## Test plan

- [x] Trigger manually from the Actions UI with dry-run checked — verify no pushes or PRs are created, output shows what would happen
- [x] Trigger manually with dry-run unchecked — verify `origin/master` is updated and a conflict-free or conflict PR is opened
- [x] Unit tests pass: `python3 -m pytest .github/scripts/test_automerge.py`

### DIFF
--- a/.github/scripts/automerge.py
+++ b/.github/scripts/automerge.py
@@ -1,10 +1,10 @@
 #!/usr/bin/env python3
 # Copyright Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
-"""Sync ROCgdb sourceware/master into origin/amd-staging.
+"""Sync ROCgdb upstream branch into the target branch.
 
-Fetches the upstream sourceware master, fast-forwards origin/master to match,
-then merges cleanly into origin/amd-staging. On conflict, pushes a conflict
+Fetches the upstream branch, fast-forwards the origin mirror to match,
+then merges cleanly into the target branch. On conflict, pushes a conflict
 branch and opens a PR for manual resolution.
 """
 
@@ -22,20 +22,52 @@ from pathlib import Path
 # Configuration
 # ---------------------------------------------------------------------------
 
-ROCGDB_REPO = "ROCm/ROCgdb"
+SOURCEWARE_URL = "git://sourceware.org/git/binutils-gdb.git"
+SOURCEWARE_REMOTE = "sourceware"
+
+_rocgdb_repo = os.environ.get("GITHUB_REPOSITORY", "")
+if not _rocgdb_repo:
+    print(
+        "Error: GITHUB_REPOSITORY environment variable is required.",
+        file=sys.stderr,
+    )
+    sys.exit(1)
+ROCGDB_REPO = _rocgdb_repo
 ROCGDB_ORIGIN_URL = f"https://github.com/{ROCGDB_REPO}.git"
-SOURCEWARE_URL = "https://sourceware.org/git/binutils-gdb.git"
+
+_target_branch = os.environ.get("AUTOMERGE_TARGET_BRANCH", "")
+if not _target_branch:
+    print(
+        "Error: AUTOMERGE_TARGET_BRANCH environment variable is required.",
+        file=sys.stderr,
+    )
+    sys.exit(1)
+TARGET_BRANCH = _target_branch
+
+_upstream_branch = os.environ.get("AUTOMERGE_UPSTREAM_BRANCH", "")
+if not _upstream_branch:
+    print(
+        "Error: AUTOMERGE_UPSTREAM_BRANCH environment variable is required.",
+        file=sys.stderr,
+    )
+    sys.exit(1)
+UPSTREAM_BRANCH = _upstream_branch
 
 
 def _parse_args() -> argparse.Namespace:
     p = argparse.ArgumentParser(
-        description="Sync sourceware/master into origin/amd-staging."
+        description=f"Sync {SOURCEWARE_REMOTE}/{UPSTREAM_BRANCH} into origin/{TARGET_BRANCH}."
     )
     p.add_argument(
         "--workspace",
         metavar="DIR",
         help="Root directory for the working clone (overrides $WORKSPACE). "
         "The clone is created at DIR/rocgdb_sync/binutils-gdb.",
+    )
+    p.add_argument(
+        "--repo",
+        metavar="DIR",
+        help="Path to an existing ROCgdb clone to use directly, skipping the clone step.",
     )
     p.add_argument(
         "--dry-run",
@@ -62,10 +94,11 @@ else:
 
 # Never pass secrets as CLI args — use env vars (e.g. GH_TOKEN) instead.
 
-CONFLICT_BRANCH_PREFIX = "users/github/master-to-amd-staging-conflict"
-FF_BRANCH_PREFIX = "users/github/master-to-amd-staging-ff"
+_branch_infix = f"{UPSTREAM_BRANCH}-to-{TARGET_BRANCH}".replace("/", "-")
+CONFLICT_BRANCH_PREFIX = f"users/github/{_branch_infix}-conflict"
+CONFLICT_FREE_BRANCH_PREFIX = f"users/github/{_branch_infix}-conflict-free"
 CONFLICT_LABEL = "merge-conflict"
-FF_LABEL = "merge-testing"
+CONFLICT_FREE_LABEL = "merge-testing"
 CI_SKIP_LABEL = "ci:skip"
 
 # Sleep durations (seconds) between successive network attempts.
@@ -118,6 +151,8 @@ def _is_network_error(exc: subprocess.CalledProcessError) -> bool:
         "no route to host",
         "fatal: unable to access",
         "ssh: connect",
+        "http 429",
+        "rpc failed",
     )
     return any(k in stderr for k in keywords)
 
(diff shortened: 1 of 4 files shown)


lumachad:
Updated to do a general FF -> conflict-free merge renaming.

.github/scripts/automerge.py
```diff
@@ -454,49 +489,74 @@ def main() -> None:
...
             )
             if DRY_RUN:
-                print(f"[DRY RUN] would push {ff_branch} and open fast-forward PR.")
+                print(
+                    f"[DRY RUN] Would open conflict-free PR:\n"
+                    f"  Branch : {conflict_free_branch}\n"
+                    f"  Commits: {merge_base_sha[:12]}..{last_clean[:12]}"
+                    f" ({clean_count} commit(s))\n"
+                    f"  Title  : {conflict_free_title}"
+                )
```
spatrang, line 567: When `--dry-run` is set and there is a partial clean prefix (`last_clean is not None`), this block prints the conflict-free dry-run output but does not `return`. Execution continues into §8 and also prints conflict dry-run output.

Please add `return` after the dry-run `print` (same as §7a at L525).

.github/scripts/automerge.py
```diff
@@ -22,14 +22,41 @@
...
+if not _rocgdb_repo:
+    print(
+        "Error: GITHUB_REPOSITORY environment variable is required.",
+        file=sys.stderr,
+    )
+    sys.exit(1)
+ROCGDB_REPO = _rocgdb_repo
+ROCGDB_ORIGIN_URL = f"https://github.com/{ROCGDB_REPO}.git"
+
+_target_branch = os.environ.get("AUTOMERGE_TARGET_BRANCH", "")
```
spatrang, line 38: Required env vars are validated at import time (`GITHUB_REPOSITORY`, `AUTOMERGE_TARGET_BRANCH`, `AUTOMERGE_UPSTREAM_BRANCH`). `test_automerge.py` loads the module without setting these and will fail.

Either defer validation to `main()`, or update `_load_automerge()` to inject the env vars before `exec_module`.

.github/scripts/automerge.py
```diff
@@ -454,49 +489,74 @@ def main() -> None:
...
+                    f"  Title  : {conflict_free_title}"
+                )
             else:
-                run(["git", "checkout", "-B", ff_branch], cwd=repo)
-                run(["git", "merge", "--no-edit", commits[-1]], cwd=repo)
-                run_net(["git", "push", "origin", ff_branch], cwd=repo)
-                open_ff_pr(
-                    branch=ff_branch,
+                run_net(
+                    ["git", "push", "origin", f"{commits[-1]}:refs/heads/{conflict_free_branch}"],
```
spatrang: Conflict-free branches are now pushed via direct refspec (`{commit}:refs/heads/{branch}`) instead of creating a local merge commit first. Please confirm this matches expected CI behavior — the PR head is the upstream commit, not a pre-merged branch tip on top of the target.

.github/workflows/automerge.yml
```diff
@@ -3,9 +3,13 @@ on:
   schedule: # runs every 6 hours
     - cron: '0 */6 * * *'
   workflow_dispatch:
+    inputs:
+      dry_run:
+        description: 'Dry run (fetch and probe, but skip pushes and PR creation)'
+        type: boolean
+        default: false
   pull_request:
     types: [closed]
```
spatrang: Removing `branches: [amd-staging]` means every closed PR in the repo starts this workflow (job is skipped by `if`, but checkout still runs). If that's intentional for variable-driven target branch, fine — otherwise consider restoring a branch filter to reduce Actions noise/cost.

.github/workflows/automerge.yml
```diff
@@ -14,19 +18,20 @@ concurrency:
...
-    # Remove the above and uncomment below to activate:
-    # if: |
-    #   github.event_name != 'pull_request' ||
-    #   (github.event.pull_request.merged == true &&
-    #    (contains(github.event.pull_request.labels.*.name, 'merge-testing') ||
-    #     contains(github.event.pull_request.labels.*.name, 'merge-conflict')))
+    if: |
+      github.event_name != 'pull_request' ||
+      (github.event.pull_request.merged == true &&
+       github.event.pull_request.base.ref == vars.AUTOMERGE_TARGET_BRANCH &&
```
spatrang: This PR activates scheduled and post-merge triggers. Please confirm these repository variables are set before merge:

- `AUTOMERGE_TARGET_BRANCH` (e.g. `amd-staging`)
- `AUTOMERGE_UPSTREAM_BRANCH` (e.g. `master`)

Without them, scheduled runs fail immediately and the re-sync condition (`base.ref == vars.AUTOMERGE_TARGET_BRANCH`) never matches.

spatrang (changes requested):
## Summary

Nice improvement — generic config, dry-run wiring, conflict-free naming, and enabling the workflow all look solid.

**Action items before merge:**
- Fix dry-run fall-through in §7b (bug)
- Confirm repo variables are configured
- Update tests/docs for renamed API (this PR or immediate follow-up)

---

### Comments on files not changed in this PR

**`.github/scripts/test_automerge.py` — renamed symbols**

`test_automerge.py` still references removed/renamed APIs: `find_open_ff_pr`, `open_ff_pr`, `push_master_mirror`. Please update to `find_open_conflict_free_pr`, `open_conflict_free_pr`, and `push_upstream_mirror` in this PR or an immediate follow-up.

**`.github/scripts/test_automerge.py` ~L203–221 — merge abort test expectation**

`test_cleanup_runs_even_when_abort_raises` expects `probe_clean_prefix` to raise when `merge --abort` fails. This PR changed abort to `check=False` (automerge.py L378), so cleanup completes and the function returns instead of raising. Please update the test to assert cleanup still runs and `(last_clean, first_conflict)` is returned correctly.

**`.github/scripts/test_automerge.py` ~L344–351 — missing dry-run regression test**

`TestDryRun.test_dry_run_partial_prefix_skips_push_and_pr` mocks `probe_result=("aaa", "bbb")` but does not assert that conflict PR dry-run output is *not* printed. Once the §7b `return` is fixed, please add an assertion that only the conflict-free dry-run path runs for partial prefix.

**`.github/scripts/automerge.md` — documentation out of date**

Documentation still describes hardcoded `amd-staging`/`master`, "fast-forward" terminology, and old branch prefixes/constants. Please update to match configurable branches, conflict-free naming, dry-run, and required repo variables.

.github/scripts/automerge.py
lumachad: Fixed. Added `return` after the dry-run print so section 7b now mirrors section 7a.

.github/scripts/automerge.py
lumachad: Fixed in `test_automerge.py`. `_load_automerge()` now uses `patch.dict(os.environ, _TEST_ENV)` to inject the three required env vars before `exec_module` runs, so the module-level validation no longer calls `sys.exit(1)`.

.github/scripts/automerge.py
lumachad: Confirmed, this is intentional. The refspec push `{commit}:refs/heads/{branch}` creates the branch pointing directly at the upstream commit rather than a local merge commit on top of the target. The PR head is the upstream commit itself, which is what we want: reviewers see exactly the upstream commits with no synthetic merge noise, and CI runs against that state.

.github/workflows/automerge.yml
lumachad: Intentional. Since the target branch is now variable-driven (`vars.AUTOMERGE_TARGET_BRANCH`), we cannot put a static branch name in the `on: pull_request: branches:` filter. The `if:` condition on the job already gates execution, so only PRs that (a) are merged, (b) target the configured branch, and (c) carry the relevant label will do any real work. All others skip at the job level. The checkout overhead on skipped runs is minor and the tradeoff was accepted to keep the branch configurable.

.github/workflows/automerge.yml
lumachad: Yes, both repository variables need to be set before the scheduled runs activate. This is part of the merge checklist. The variables should be configured as: `AUTOMERGE_TARGET_BRANCH=amd-staging` and `AUTOMERGE_UPSTREAM_BRANCH=master`.

.github/workflows/automerge.yml
lumachad: Update: the `pull_request` trigger has been dropped entirely. The workflow now only runs on the 6-hour schedule and `workflow_dispatch`. Simpler and no spurious checkouts.

spatrang (approved):
## Second-round review

All first-round items are addressed. Unit tests pass (`19` tests).

| First-round item | Status |
| --- | --- |
| §7b dry-run missing `return` | Fixed (L549) |
| Import-time env vars break tests | Fixed — `_load_automerge()` injects `_TEST_ENV` |
| Tests still used old APIs (`find_open_ff_pr`, `open_ff_pr`, `push_master_mirror`) | Fixed |
| Abort test expected a raise after `check=False` | Fixed — `test_cleanup_runs_even_when_abort_fails` asserts return values + cleanup |
| Dry-run partial-prefix missing conflict-path assertion | Fixed — `mock_conflict_pr.assert_not_called()` |
| `automerge.md` out of date | Fixed |
| Direct refspec push | Confirmed intentional |
| Broad `pull_request` trigger | Resolved by dropping the trigger; schedule + `workflow_dispatch` only |
| Repo variables must be set before merge | Confirmed (`AUTOMERGE_TARGET_BRANCH=amd-staging`, `AUTOMERGE_UPSTREAM_BRANCH=master`) |

**Pre-merge reminder (not a code change):** set those two repository variables before the scheduled job starts running.

LGTM.

PR #216, opened 2026-07-16
gdb/testsuite: Add runtime-core-attached test

amd-bfilipov:
Test core dump generation while program runs under GDB. With the fix for core dump collection during debugging, verify that running under GDB in batch mode with signal passthrough generates both host and GPU cores, which can be merged and loaded.

What we're testing:
1. Run a HIP program under GDB
2. Configure GDB to pass signals through (not stop on them)
3. Program crashes with GPU fault
4. ROCm runtime generates core dumps even though GDB is attached
5. Verify GDB can load those cores and see GPU threads

Uses batch mode: -ex 'handle SIGABRT/SIGSEGV nostop pass' -ex 'run' This allows GDB to catch faults but pass them to runtime/kernel.

## Motivation

Tests the new capability from ROCm/rocm-systems#6811 that enables core dump generation even when a debugger is
 attached

## Technical Details

  - Runs test program under GDB in batch mode with signal passthrough configuration
  - Verifies both host (`core.*`) and GPU (`core.*.gpu`) cores are generated
  - Merges cores using coremerge and validates they can be loaded in GDB
  - Uses `system` command to inherit test environment (LD_LIBRARY_PATH)
  - Tests pagefault and abort scenarios


gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,113 @@
...
+    set filtered_hostcore_list {}
+    foreach core $hostcore_list {
+        if {![string match "*.gpu" $core]} {
+            lappend filtered_hostcore_list $core
+        }
+    }
+    set hostcore_list $filtered_hostcore_list
+
+    if {[llength $gpucore_list] != 1} {
+        fail "GPU core not found for $fault (found [llength $gpucore_list])"
```
lancesix: Not finding the corefiles would probably not be a hard error.

The system could be configured in a way which does not allow generating core dump, or the cpu core could use a different naming pattern than what you expect.

Also, when running with GDB, if you build GDB with `-fsanitize=address`, this implicitly sets `ulimit -c 0` (can be overriden, but this is what is done), which would prevent the creation of core dumps.

So if core dumps are not found, it is an UNSUPPORTED, but if they do exist, then they must be valid, and we can merge them / open them in GDB.

lancesix (commented):
Some preliminary comments.  Will do a more in depth review later.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,113 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test core dump generation while GDB is attached.
+# With the fix for core dump collection during debugging, verify that
```
aktemur: It's not clear which "fix" this refers to. I think we can remove the phrase and start the sentence with "Verify that.."

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,113 @@
...
+load_lib rocm.exp
+
+require allow_rocm_core_tests
+require allow_hip_tests
+
+standard_testfile runtime-core.cpp
+
+if { [build_executable "failed to prepare" ${testfile} ${srcfile} \
+	  {debug hip}] } {
+	return -1
```
aktemur: Please check the indentation in the entire file. In Tcl files, the convention is to indent by 4 spaces, and use tabs for each 8-space.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,113 @@
...
+}
+
+proc do_attached_test { fault } {
+	global binfile
+	global GDB
+	global decimal hex
```
aktemur: The upstream tendency is to use global variables like `$::binfile` instead of using the `global` keyword.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,113 @@
...
+
+	set coredir [standard_output_file "coredir-$fault"]
+	remote_exec build "mkdir -p $coredir"
+	remote_exec build "rm -f $coredir/core.*"
+
+	set cmd "(cd $coredir && ulimit -c unlimited && $GDB -batch -ex 'handle SIGABRT nostop pass' -ex 'handle SIGSEGV nostop pass' -ex 'run' -args $binfile $fault; true)"
```
aktemur: Can we not run these commands like in regular tests after a `clean_restart`?

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,113 @@
...
+		lappend gpucore_list $core
+	}
+	}
+
+	set filtered_hostcore_list {}
+	foreach core $hostcore_list {
+	if {![string match "*.gpu" $core]} {
+		lappend filtered_hostcore_list $core
+	}
+	}
```
aktemur: These two loops can be combined.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,113 @@
...
+	set hostcore_list $filtered_hostcore_list
+
+	if {[llength $gpucore_list] != 1} {
+	unsupported "GPU core not found for $fault (found [llength $gpucore_list])"
```
aktemur: What if there were more than one core files found? The message would be misleading. GPU core was found, but was too many. If the length of the list is 0, then core was not found. An in these cases, do we want UNSUPPORTED or FAIL?

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,113 @@
...
+	return
+	}
+
+	if {[llength $hostcore_list] != 1} {
+	unsupported "Host core not found for $fault (found [llength $hostcore_list])"
```
aktemur: Similar comment here.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,113 @@
...
+	set coremerge [gdb_find_coremerge]
+	set merged_core "$coredir/merged-$fault.core"
+	remote_exec build "$coremerge $merged_core $hostcore $gpucore"
+
+	if {![remote_file build exists $merged_core]} {
+	fail "Failed to merge cores for $fault"
+	return
+	}
+
+	pass "Merged cores for $fault"
```
aktemur: In case of a problem, the FAILed test name will be different than the PASSed test name. I think we can do something like
```
gdb_assert {[remote_file build exists $merged_core]} "merged cores"
```
It is not too important to exit the test if there is a failure, IMHO. The tests below would continue emiting FAILs.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,113 @@
...
+
+	if {[llength $hostcore_list] != 1} {
+	unsupported "Host core not found for $fault (found [llength $hostcore_list])"
+	return
+	}
+
+	set gpucore [lindex $gpucore_list 0]
+	set hostcore [lindex $hostcore_list 0]
+
+	pass "Generated both cores for $fault while GDB attached"
```
aktemur: We don't need to include `$fault` in the test names. There is already the test prefix.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,113 @@
...
+
+	if {![remote_file build exists $merged_core]} {
+	fail "Failed to merge cores for $fault"
+	return
+	}
+
+	pass "Merged cores for $fault"
+
+	clean_restart
+	gdb_load $binfile
```
aktemur: Nit: `clean_restart $::testfile`

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,113 @@
...
+	gdb_test "core-file $merged_core" \
```
aktemur:
```
(gdb) help core-file
Use FILE as core dump for examining memory and registers.
Usage: core-file FILE
No arg means have no core file.  This command has been superseded by the
`target core' and `detach' commands.
```

We probably shall use the `target core` command instead.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
aktemur: > What if there were more than one core files found? The message would be misleading. GPU core was found, but was too many. If the length of the list is 0, then core was not found. An in these cases, do we want UNSUPPORTED or FAIL?

I saw @lancesix's comment about this. Hmm, if the setup does not create a core file in the way we expect, by emitting UNSUPPORTED, would we miss some problems that would require attention? Maybe we can emit UNTESTED instead, and define ways in the CI to catch UNTESTED cases. UNSUPPORTED may go completely undetected, I'm afraid. @lancesix, @lumachad, what do you think?

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
lancesix: I second that. I prefer the $:: syntax.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
lancesix: The main issue we could have with `clean_restart` is it does not allow us to set `ulimit`.   I expect this needs to be executed in the shell running GDB.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
lancesix: I am not sure I have a strong opinion. The intent was: we do not control core dump creation in the testsuite, so the best we can do is ensure that if a core is produced, it is correct. If no core is produced, then, there is not much we can say.

For CI environment, if we can have a guarantee that it should create a coredump, then yes, we should be more strict, and tread the UNTESTED / UNSUPPORTED (I would need to check what is the established pattern to use one or the other) as a hard FAIL.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
aktemur: > The main issue we could have with `clean_restart` is it does not allow us to set `ulimit`. I expect this needs to be executed in the shell running GDB.

For `ulimit`, yes, but the part starting with `$GDB` should be doable in regular manner, I believe.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: "For CI environment, if we can have a guarantee (...)". 
Don't like the sound of that. If you don't mind, I would leave this one as it is.

lumachad: corefile test are usually a bit finicky, and I'd say the PR's author should check they ran fine. Relying on a specific state for CI to notify us tends to be fragile, unless it is ERROR/FAIL etc.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,103 @@
...
+    return -1
+}
+
+proc do_attached_test { fault } {
+
+    set coredir [standard_output_file "coredir-$fault"]
+    remote_exec build "mkdir -p $coredir"
+    remote_exec build "rm -f $coredir/core.*"
+
+    set cmd "(cd $coredir && ulimit -c unlimited && $::GDB -batch -ex 'handle SIGABRT nostop pass' -ex 'handle SIGSEGV nostop pass' -ex 'run' -args $::binfile $fault; true)"
```
lumachad: I don't think we should be setting ulimit in gdb's testsuite. This should be set by whoever is invoking the testsuite.

lumachad (commented):
Could we please have a better description of what we're trying to test here? The "test core dump while gdb is attached" sounded to me like we were controlling rocgdb when generating the core file, but it sounds like what we want is for some unrelated process to be generating a core file while we test something else in the runtime.

Is that the case? If so, should this be a ROCR test instead?

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
lancesix: Upstream GDB is already this in `core_find`: https://sourceware.org/cgit/binutils-gdb/tree/gdb/testsuite/lib/gdb.exp?id=c116c67be27fad91095a2bb13bc0bd48f4e4d623#n9806.  It looks to me this particular patch is re-using this pattern.

lumachad: I see. That makes the code in gdb.exp a bit more annoying, as it insists on...

	warning "can't generate a core file - core tests suppressed - check ulimit -c""

... even if it sets ulimit -c. So it should probably give a better message instead of something misleading like that.

I think it would be more beneficial to check ulimit -c and make sure it is sane before we run our tests. If we can't set it, then we can't test it either.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: I tried refactoring this to use gdb_spawn + remote_spawn instead of the raw shell command, but ran into issues. The problem is we need ulimit to affect the shell that spawns GDB (not just be a GDB flag), and since we're running in batch mode where GDB exits after the crash, using expect + eof gets messy. The spawned process exits with non-zero status (expected since the program crashes), which remote_spawn treats as an error.

Anyways, this instance of gdb is spawned only for it to crash and trigger coredump creation, we don't care about any of it's outputs, so I think we don't need to run it regularly, it seems simpler to do it this way.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: Changed the description so it's more clear.
Added a check to see if we can actually set ulimit to unlimited and exit if we can not.

aktemur: I see the problem
```
Python Exception <class 'ModuleNotFoundError'>: No module named 'gdb'
/home/taktemur/out/gdb-roc/gdb/testsuite/../../gdb/gdb: warning:
Could not load the Python gdb module from `/usr/local/share/gdb/python'.
Limited Python support is available from the _gdb module.                                                                                                                                          Suggest passing --data-directory=/path/to/gdb/data-directory.
```
This is because we shouldn't be running GDB in a shell command. We can use `clean_restart` instead and submit the debugger commands. Like this:

```
    clean_restart $::testfile
    gdb_test "handle SIGABRT nostop pass" ...
    gdb_test "handle SIGSEGV nostop pass" ...
    gdb_run_cmd $fault

    gdb_test_multiple "" "...." {
       -re "..."
```

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,113 @@
...
+    catch {system "$cmd"} result
+
+    verbose "GDB batch command: $cmd" 2
+    verbose "GDB batch result: $result" 2
```
aktemur: This prints logs to the screen. Please use `verbose -log "..."` to have the logs printed to the log file only. Otherwise the screen will be polluted.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,113 @@
...
+    set gpucore [lindex $gpucore_list 0]
+    set hostcore [lindex $hostcore_list 0]
+
+    pass "Generated both cores while GDB attached"
+
+    set coremerge [gdb_find_coremerge]
+    set merged_core "$coredir/merged-$fault.core"
+    remote_exec build "$coremerge $merged_core $hostcore $gpucore"
+
+	gdb_assert {[remote_file build exists $merged_core]} "Merged cores"
```
aktemur: Indentation is off.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,113 @@
...
+
+    clean_restart $::testfile
+    gdb_test "target core $merged_core" \
+    [multi_line \
+	".*" \
+	"Core was generated by .*" \
+	"Program terminated with signal SIGABRT.*" \
+	"Thread $::decimal \\(AMDGPU Wave \[^\r\n\]*\\).*"] \
+    "load $fault merged corefile"
```
aktemur: Please indent for 4 spaces more. 8 spaces will need to be converted to tabs.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,113 @@
...
+
+    gdb_test "info threads" ".*AMDGPU Wave.*" "verify GPU threads"
+
+    gdb_exit
```
aktemur: This shouldn't be necessary.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: Ok changed it to `clean_restart` approach. 

Also changed the test so it doesn't set ulimit to unlimited, and instead rely on testing enviroment to have it set. Same as it `runtime-core.exp` which was the test that this test is based on.
@aktemur

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+
+    set coredir [standard_output_file "coredir-$fault"]
+    remote_exec build "mkdir -p $coredir"
+    remote_exec build "rm -f $coredir/core*"
+
+    clean_restart $::testfile
+    gdb_test "handle SIGABRT nostop pass" ".*"
+    gdb_test "handle SIGSEGV nostop pass" ".*"
+    gdb_test_no_output "set cwd $coredir"
+    gdb_test "set args $fault" ".*"
```
aktemur: `gdb_test_no_output`

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+
+    gdb_run_cmd
+
+    # Wait for program to crash and cores to be written.  This can take
+    # a while for large core dumps.
+    with_timeout_factor 3 {
+	gdb_test_multiple "" "wait for program exit" {
+	    -re "Program terminated with signal" {
```
aktemur, line 53: Also `-wrap`?

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+		pass "program crashed"
+	    }
+	    -re "The program no longer exists" {
```
aktemur, line 56: Also `-wrap`?

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+    gdb_test "set args $fault" ".*"
+
+    gdb_run_cmd
+
+    # Wait for program to crash and cores to be written.  This can take
+    # a while for large core dumps.
+    with_timeout_factor 3 {
+	gdb_test_multiple "" "wait for program exit" {
+	    -re "Program terminated with signal" {
+		pass "program crashed"
```
aktemur: Let's please use `$gdb_test_name`. Also in the occurrences below.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+	    }
+	    -re "The program no longer exists" {
+		pass "program crashed"
+	    }
+	    -re "$::gdb_prompt $" {
+		pass "program crashed"
+	    }
```
aktemur: We should use `-wrap` with `""` being the regexp. 

But the case doesn't feel right. Any kind of previous output (except those two above) are accepted?

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+	    -re "\r\n" {
+		exp_continue
+	    }
```
aktemur: Why do we need this? Can you give sample outputs of the program? I think we can improve the cases here.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+    set gpucore_list {}
+
+    set filtered_hostcore_list {}
+
+    foreach core $hostcore_list {
+	if {[string match "*.gpu" $core]} {
+	    lappend gpucore_list $core
+	}
+
+	if {![string match "*.gpu" $core]} {
```
aktemur: `else`?

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+		exp_continue
+	    }
+	    timeout {
+		fail "wait for program exit (timeout)"
+		return
+	    }
+	}
+    }
+
+    set hostcore_list [glob -nocomplain -directory $coredir "core*"]
```
aktemur: Maybe call this variable `allcore_list`, and ....

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+    set gpucore_list {}
+
+    set filtered_hostcore_list {}
```
aktemur: ... name this `hostcore_list`, so that ...

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+	if {[string match "*.gpu" $core]} {
+	    lappend gpucore_list $core
+	}
+
+	if {![string match "*.gpu" $core]} {
+	    lappend filtered_hostcore_list $core
+	}
+    }
+
+    set hostcore_list $filtered_hostcore_list
```
aktemur: ... this is not necessary.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+
+    if {[llength $gpucore_list] != 1} {
+	unsupported "GPU core not found (found [llength $gpucore_list])"
```
aktemur: Being pedantic here, if the value is larger than 1, we would still pring "GPU core not found" although technically we found "too many". Maybe say "single GPU core file must exist" or something similar.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+	return
+    }
+
+    if {[llength $hostcore_list] != 1} {
+	unsupported "Host core not found (found [llength $hostcore_list])"
```
aktemur: Similar comment here: "single host core file...

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+	return
+    }
+
+    set gpucore [lindex $gpucore_list 0]
+    set hostcore [lindex $hostcore_list 0]
+
+    pass "Generated both cores while GDB attached"
```
aktemur: Lowercase "G" please.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+
+    set coremerge [gdb_find_coremerge]
+    set merged_core "$coredir/merged-$fault.core"
+    remote_exec build "$coremerge $merged_core $hostcore $gpucore"
+
+    gdb_assert {[remote_file build exists $merged_core]} "Merged cores"
```
aktemur: Lowercase "M" please.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+
+    clean_restart $::testfile
+    gdb_test "target core $merged_core" \
+	[multi_line \
+	    ".*" \
+	    "Core was generated by .*" \
```
aktemur, line 106: I feel uneasy about `.*`'s in such cases, because they can also match newline. Can we use `\[\r\n\]*` instead. Also in the line below.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+	    "Program terminated with signal SIGABRT.*" \
+	    "Thread $::decimal \\(AMDGPU Wave \[^\r\n\]*\\).*"] \
+	"load $fault merged corefile"
+
+    gdb_test "info threads" ".*AMDGPU Wave.*" "verify GPU threads"
```
aktemur: The leading `.*` is not needed. It is implicit.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+# Test core dump generation while GDB is attached.
+# Verify that running a program under GDB with signal passthrough
+# generates both host and GPU core dumps.
+
+load_lib rocm.exp
+
+require allow_rocm_core_tests
+require allow_hip_tests
+
+standard_testfile runtime-core.cpp
```
lancesix: We usually try to have a 1:1 mapping between .ccp and .exp.  Either what is tested here fits in runtime-core.exp, or this test should have its own test program.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+      {debug hip}] } {
+    return -1
+}
+
+proc do_attached_test { fault } {
+
+    # Check that ulimit allows core dumps.  The test environment must
+    # have ulimit -c set to unlimited before running this test.
+    set ulimit_value [exec sh -c "ulimit -c"]
+    if {$ulimit_value != "unlimited"} {
```
lancesix: I am not sure checking ulimit that way is sufficient.  For example, if building GDB with ASAN, the ASAN runtime implicitly turns ulimit to 0 (the ASAN has a heavy impact on core size otherwise):

```
$ ulimit -c
unlimited
$ ASAN_OPTIONS=detect_leaks=0 ./gdb/gdb -data-directory ./gdb/data-directory -ex "shell ulimit -c" -q -batch
0
```

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,134 @@
...
+	    }
+	    timeout {
+		fail "$gdb_test_name (timeout)"
+		return
+	    }
+	}
+    }
+
+    set hostcore_list [glob -nocomplain -directory $coredir "core*"]
+    set gpucore_list {}
```
lancesix, line 70: isn't the runtime pringing the gpu core name?  Could we find this in the `gdb_test_multiple`?

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: Done. Changed to `gdb_test_no_output "set args $fault"` on line 53.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: Done. Changed to `-wrap -re ""` (lines 67-68) which now properly handles any program exit output that doesn't match the two specific patterns above. This fallback ensures we don't fail if the program terminates with slightly different output variations.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: done

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: done

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: The pattern (lines 120-124) follows the established pattern from `runtime-core.exp` for matching core file loading output. The leading `.*` on line 121 is necessary to skip the library loading messages that precede "Core was generated by". The trailing `.*` patterns allow matching variable content (command arguments, signal details, wave coordinates). The multi_line structure with these patterns has been proven to work correctly in the sibling test.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: Sure, added runtime-core-attached.cpp

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: I've removed the check, since `runtime-core.exp` does not check for it either.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: We'll go with "Single gpu/cpu core must exist"

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: done

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: Multiple messages arrive before the prompt (both "Program terminated" and "The program no longer exists"), so
 -wrap on each pattern causes race conditions; using exp_continue for messages and matching the final prompt
 directly works correctly.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: No it does not print it.

gdb/testsuite/gdb.rocm/runtime-core-attached.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+#include <hip/hip_runtime.h>
+#include <thread>
+#include <chrono>
+#include <iostream>
+#include <stdio.h>
+#include <stdlib.h>
+#include "gdb_watchdog.h"
+
+#define CHECK(cmd)                                                           \
```
lancesix: Please use the version from `rocm-test-utils.h`.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,118 @@
...
+require allow_hip_tests
+
+standard_testfile .cpp
+
+if { [build_executable "failed to prepare" ${testfile} ${srcfile} \
+      {debug hip}] } {
+    return -1
+}
+
+proc do_attached_test { fault } {
```
lancesix: attached_coredump_test?

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,118 @@
...
+	    }
+	    -re "$::gdb_prompt $" {
+		pass $gdb_test_name
+	    }
+	    timeout {
+		fail "$gdb_test_name (timeout)"
+		return
+	    }
+	}
+    }
```
lancesix, line 55: This gdb_test_multiple seems much.

The "timeout" should be handled by default by gdb_test_multiple (or gdb_test), and the other cases only use exp_continue.  All this matches is that we got the prompt back, which we could as well do with `gdb_test "run" "The program no longer exists"`, couldn't we?.

gdb/testsuite/gdb.rocm/runtime-core-attached.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+  while (true)
+    __builtin_amdgcn_s_sleep (1);
+}
+
+enum testcase_t
+{
+  memfault,
+  abort
+};
```
lumachad, line 60: `abort` as an enum member shadows the global `::abort()` called by `abort_kernel` a few lines above.  This compiles in C++ because the enum member is only in scope where the type is used as `testcase_t::abort`, but it is confusing and fragile.  Rename it, e.g. `kabort` or `do_abort`, as is idiomatic in GDB test sources — see `runtime-core.cpp` which has the same pattern and the same risk.

gdb/testsuite/gdb.rocm/runtime-core-attached.cpp
```diff
@@ -0,0 +1,130 @@
...
+    } while (0)
+
+/* Pagefault kernel.  In this testcase, OUT contains an address not reachable
+   by the GPU, triggering a page fault.  */
+
+__global__ void
+pagefault_kernel (int *out)
+{
+  int local = 42;
+  *out = 8;
```
lumachad, line 32: `int local = 42` is unused by the test — unlike `runtime-core.exp`, this test never runs `p local`.  Either remove it, or add a comment explaining it is intentionally left for future frame-variable checks (as `runtime-core.cpp` uses it).  Same applies to `abort_kernel` (line 61) and `aux_kernel` (line 73).

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,118 @@
...
+
+proc do_attached_test { fault } {
+
+    set coredir [standard_output_file "coredir-$fault"]
+    remote_exec build "mkdir -p $coredir"
+    remote_exec build "rm -f $coredir/core*"
+
+    clean_restart $::testfile
+    gdb_test "handle SIGABRT nostop pass" ".*"
+    gdb_test "handle SIGSEGV nostop pass" ".*"
```
lumachad: `".*"` as the expected pattern records no meaningful output check and produces a generic pass/fail.  Use `gdb_test_no_output` if no output check is intended, or supply a real pattern such as `"Signal.*SIGABRT.*No.*Yes.*"` so the signal-handler table is at least spot-checked.  Same for the `SIGSEGV` handle command on the next line.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,118 @@
...
+    if {[llength $hostcore_list] != 1} {
+	unsupported "single host core file must exist (found [llength $hostcore_list])"
+	return
+    }
+
+    set gpucore [lindex $gpucore_list 0]
+    set hostcore [lindex $hostcore_list 0]
+
+    pass "generated both cores while gdb attached"
+
```
lumachad: This `pass` fires before coremerge is invoked (line 100) and before `gdb_assert` checks that the merged core exists (line 101).  A coremerge failure will still leave this pass on record.  Move the `pass` to after the `gdb_assert`, or fold it into the assert message.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,118 @@
...
+    set coremerge [gdb_find_coremerge]
+    set merged_core "$coredir/merged-$fault.core"
+    remote_exec build "$coremerge $merged_core $hostcore $gpucore"
+
```
lumachad, line 92: The return value of `remote_exec build "$coremerge ..."` is ignored.  If coremerge is not found or exits non-zero, the subsequent `gdb_assert` fires with a confusing message.  Follow the pattern in `rocm.exp:599-604`: check `[remote_file build exists $merged_core]` immediately and emit a `warning` with the coremerge stderr if it is absent, then return early.

lumachad (commented):
Automated comments:

gdb/testsuite/gdb.rocm/runtime-core-attached.cpp
amd-bfilipov: Done - using CHECK from rocm-test-utils.h

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: Renamed to attached_core_test

gdb/testsuite/gdb.rocm/runtime-core-attached.cpp
amd-bfilipov: Renamed to kabort to avoid shadowing ::abort().

gdb/testsuite/gdb.rocm/runtime-core-attached.cpp
amd-bfilipov: Removed unused local variables from all kernels

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: Added proper signal handler table patterns: Signal\s+Stop\s+Print\s+Pass.*SIGABRT\s+No\s+Yes\s+Yes.* and similar
  for SIGSEGV to verify the table headers and values

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: Pass moved to after successful coremerge verification

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: Added coremerge return value check with warning and early return on failure, following rocm.exp pattern

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: Simplified to gdb_test "" ".*The program no longer exists.*" as suggested

gdb/testsuite/gdb.rocm/runtime-core-attached.cpp
```diff
@@ -0,0 +1,113 @@
...
+aux_kernel ()
+{
+  while (true)
+    __builtin_amdgcn_s_sleep (1);
+}
+
+enum testcase_t
+{
+  memfault,
+  kabort
```
lumachad, line 59: The enum value is `memfault` but the command-line argument accepted by `main()` (and passed by the `.exp`) is `"pagefault"`.  The mismatch is a readability hazard when adding further fault types.  Consider renaming the enum value to `pagefault` to match the string.

gdb/testsuite/gdb.rocm/runtime-core-attached.cpp
```diff
@@ -0,0 +1,113 @@
...
+  else if (teststr == "abort")
+    test = testcase_t::kabort;
+  else
+    {
+      std::cerr << "Invalid test name \"" << teststr << "\"" << std::endl;
+      return EXIT_FAILURE;
+    }
+
+  hipStream_t st1;
+  hipStream_t st2;
```
lumachad: `st1` and `st2` are declared here, well before they are used.  Declaring them immediately before `hipStreamCreate` keeps their scope narrow and makes the control flow easier to follow.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    set gpucore [lindex $gpucore_list 0]
+    set hostcore [lindex $hostcore_list 0]
+
+    set coremerge [gdb_find_coremerge]
+    set merged_core "$coredir/merged-$fault.core"
+    set coremerge_result [remote_exec build "$coremerge $merged_core $hostcore $gpucore"]
+    set coremerge_status [lindex $coremerge_result 0]
+    set coremerge_output [lindex $coremerge_result 1]
```
lumachad, line 86: `coremerge_status` is captured here but is not used in the guard below, which only checks whether the output file exists.  A `coremerge` run that exits non-zero but still writes a (possibly corrupt) file will pass the guard silently.  Either add `$coremerge_status != 0` as a second condition, or drop the capture if the status is not going to be checked.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,111 @@
...
+
+    if {![remote_file build exists $merged_core]} {
+	warning "coremerge failed (status $coremerge_status): $coremerge_output"
+	return
```
lumachad, line 90: When `coremerge` produces no merged core the proc calls `warning` and returns without emitting any test result (PASS, FAIL, or UNSUPPORTED).  The caller (`foreach_with_prefix`) cannot distinguish a skipped sub-test from a passed one, so the failure is invisible in the log summary.  Should emit `unsupported` or `fail` before returning, consistent with the pattern already used for the missing-core cases above.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    pass "generated and merged cores while gdb attached"
+
```
lumachad, line 94: `pass "generated and merged cores while gdb attached"` fires here, before the merged core is loaded in GDB and GPU threads are verified.  If the `target core` or `info threads` step below fails, the test still has a PASS in the log.  Consider either removing this call (the `gdb_test` results below record success) or renaming it to something narrower like `"coremerge succeeded"`.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,111 @@
...
+    gdb_test "target core $merged_core" \
+	[multi_line \
+	    ".*" \
+	    "Core was generated by .*" \
+	    "Program terminated with signal SIG(ABRT|SEGV).*" \
+	    "Thread $::decimal \\(AMDGPU Wave \[^\r\n\]*\\).*"] \
+	"load $fault merged corefile"
+
+    gdb_test "info threads" "AMDGPU Wave.*" "verify GPU threads"
+}
```
lumachad, line 140: Two related gaps here:

1. The `aux_kernel` comment in the `.cpp` file states its purpose explicitly: to ensure that GDB can load waves "which have not raised an exception" alongside the faulting wave.  That is a key correctness property of the merged core, but this test never verifies the aux wave is present.  `runtime-core.exp` handles this with per-wave identification; at minimum a check that an `aux_kernel` wave appears in `info threads` would confirm the intent.

2. More broadly, `runtime-core.exp` also accesses GPU-side variables (`p local`, `p out`, `p some_global`, `p data[12]`) and unwinds frames from the loaded core.  This test only runs `info threads`.  Since the goal is to prove that a core produced under an attached debugger is *valid*, at least one variable read from the faulting wave would meaningfully strengthen the assertion.

lumachad (commented):
Additional automated comments.

lumachad:
We should probably try to validate in as many gfx architectures as we can. It needs manual validation though, as CI only covers gfx942 for corefiles.

gdb/testsuite/gdb.rocm/runtime-core-attached.cpp
amd-bfilipov: Renamed to pagefault to match the CLI argument

gdb/testsuite/gdb.rocm/runtime-core-attached.cpp
amd-bfilipov: Moved declarations to immediately before their respective hipStreamCreate calls

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: Added $coremerge_status != 0 as the first condition: `if {$coremerge_status != 0 || ![remote_file build
  exists $merged_core]}`

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: Changed warning to unsupported to emit a proper test result, matching the pattern used for missing
  GPU/host core cases above

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: Renamed to pass "coremerge succeeded" to accurately reflect what is verified at that point

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
amd-bfilipov: Added comprehensive wave identification following the runtime-core.exp pattern: gdb_test_multiple parses
  info threads output to identify both the faulty wave (marked with *) and the aux wave (matching aux_kernel ()),
  with gdb_assert verifying both are found. Also added a with_test_prefix "faulty_wave" block that switches to the
  faulty wave, unwinds to the kernel frame, and verifies variable access (p out == 0 for pagefault).

lumachad:
> [!NOTE]
> The following review comments were generated with automated assistance. Please use your own judgement on each point.

## Review

### Mandatory

**`runtime-core-attached.exp:1`** — Copyright year is `2026` only, but the companion `.cpp` correctly uses `2023-2026`. Fix to `2023-2026`.

---

**`runtime-core-attached.exp:20-22`** — The file-level comment says "Verify that running a program under GDB with signal passthrough generates both host and GPU core dumps." The test also verifies that the merged core can be loaded and GPU waves are visible. The comment understates what is checked; expand it.

---

**`runtime-core-attached.exp:46`** — The test message `"set core directory"` is misleading:
```tcl
gdb_test_no_output "set cwd $coredir" "set core directory"
```
`set cwd` sets the inferior's working directory, not a GDB core-dump directory. The kernel writes the core to that path only because it becomes the process's cwd. Suggested: `"set inferior working directory"`.

---

**`runtime-core-attached.exp:39`** — `remote_exec build "rm -f $coredir/core*"` is fragile: leftover cores from a previous partial run are removed, but any core written by a concurrent run sharing the same output directory will pollute the glob count. Using `file delete -force $coredir` then re-creating it, or per-invocation subdirectories, would give a truly clean state.

---

**`runtime-core-attached.exp:49`** — `HSA_COREDUMP_PATTERN` is not set. On ROCm 7.3+ the default names are `core.<pid>` (host) and `core.<pid>.gpu` (GPU), and `[string match "*.gpu" $core]` handles that correctly. However, if the environment already has `HSA_COREDUMP_PATTERN` set to the `gpucore.%p` pattern used elsewhere (e.g. by `rocm_core_find_1`), the `.gpu` suffix classification will fail silently. Wrap the run in a `save_vars { ::env(HSA_COREDUMP_PATTERN) }` block with a known explicit pattern, or document the assumption.

---

**`runtime-core-attached.exp:69-75`** — When `[llength $gpucore_list] != 1` or `[llength $hostcore_list] != 1` the test calls `unsupported` and returns rather than `fail`. If the runtime produces two GPU cores (one per stream) this becomes a silent `UNSUPPORTED` instead of flagging a real problem. At minimum, add a comment explaining the `unsupported` policy here.

---

**`runtime-core-attached.exp:49`** — `gdb_run_cmd` followed by:
```tcl
gdb_test "" ".*The program no longer exists.*" "wait for program exit"
```
With `handle SIGABRT/SIGSEGV nostop pass`, GDB prints `"Program received signal SIGABRT, Aborted."`, resumes, then prints `"The program no longer exists."` all in one burst. The `.*` prefix should absorb the signal line, but the interaction between `gdb_run_cmd` (which sends `run\n`) and the subsequent blank `gdb_test` is fragile. Consider using `gdb_test_multiple` with separate `-re` arms for the signal notification and the exit message.

---

**`runtime-core-attached.exp:93`** — Explicit `pass` calls are discouraged; passes should come from `gdb_test*` or `gdb_assert`. Suggested:
```tcl
gdb_assert {$coremerge_status == 0} "coremerge succeeded"
```

---

**`runtime-core-attached.exp:128-137`** — The `with_test_prefix "faulty_wave"` block only runs `gdb_test "p out == 0"` for the pagefault case. For the abort case no variable is read from the faulty wave. The abort kernel has no local variables, unlike `runtime-core.cpp` which adds `int local = 42`. This is a silent coverage gap. Add a kernel-local variable to `abort_kernel` and test it, or add a comment explaining why no variable check is possible.

---

**`runtime-core-attached.cpp:59`** — The enum member `kabort` is inconsistent with the kernel name `abort_kernel` and the accepted command-line string `"abort"`. The mismatch is confusing; rename to `tabort` or pick a consistent scheme across the enum, kernel name, and argv string.

---

**Tests** — No `require no_optimized_code` guard. `runtime-core.exp` includes it with the explicit rationale that debug behaviour is unreliable on optimized device code. Add the guard for now, with a note that ideally this test should be made to work with optimization as well.

---

**Tests** — The new `.cpp` provides no device-side data for the abort case (no locals, no globals). Add at least a `__device__` global or a kernel-local so the abort path can prove core content is accessible, matching the coverage level of `runtime-core.cpp`. This also enables the missing variable check in the `faulty_wave` block noted above.

---

### Optional

**PR description** — The description says "Uses batch mode" and "in batch mode" multiple times, but the implementation uses normal interactive GDB mode (`clean_restart` + `gdb_run_cmd`). Update the description to avoid confusion for future readers.

amd-bfilipov:
Comment 1: Copyright year
-------------------------
Fix copyright from 2026 to 2023-2026.

Response:
Fixed. Both copyright lines now show 2023-2026 to match the .cpp file.


Comment 2: File-level comment expansion
----------------------------------------                                                                            
Expand comment to mention merged core loading and GPU wave visibility.                                                                                                                                                                  Response:                                                                                                           Expanded the file-level comment to read: "Verify that running a program under GDB with signal passthrough generates both host and GPU core dumps, that the merged core can be loaded, and that GPU waves are visible and accessible."


Comment 3: Test message clarity
--------------------------------                                                                                    
Change "set core directory" to "set inferior working directory".                                                                                                                                                                        Response:                                                                                                           Fixed. The test message now correctly describes what `set cwd` does - it sets the inferior's working directory, not a GDB core directory.

Comment 4: Core cleanup fragility
----------------------------------
Use `file delete -force` instead of `rm -f` for truly clean state.

Response:
Replaced `rm -f $coredir/core*` with `file delete -force $coredir` followed by `mkdir -p`. This ensures a truly clean state by removing the entire directory and recreating it, eliminating any potential pollution from concurrent runs.


Comment 5: HSA_COREDUMP_PATTERN not set
----------------------------------------
Wrap execution in save_vars or document assumption.

Response:
Wrapped the entire test in `save_vars { ::env(HSA_COREDUMP_PATTERN) }` with explicit unset to ensure default ROCm naming (core.<pid> and core.<pid>.gpu) is used regardless of the environment.


Comment 6: Unsupported vs fail policy
--------------------------------------
Add comment explaining unsupported policy for core count mismatches.

Response:
Added explanatory comment: "If the core count is wrong, mark as unsupported rather than fail. This handles environments where core dump generation is not properly configured. If multiple cores appear (e.g., one per stream), that indicates a runtime issue worth investigating separately."


Comment 7: gdb_run_cmd fragility
---------------------------------
Replace gdb_run_cmd + gdb_test with gdb_test_multiple with separate arms.

Response:
Replaced the fragile pattern with `gdb_test_multiple` using separate `-re` arms for signal notification and exit message. The signal arm uses `exp_continue`, the exit arm uses `pass`. Also wrapped in `with_timeout_factor` for large core dumps.


Comment 8: Explicit pass discouraged
-------------------------------------
Replace explicit pass with gdb_assert.

Response:
Changed `pass "coremerge succeeded"` to `gdb_assert {$coremerge_status == 0} "coremerge succeeded"` following testsuite best practices.


Comment 9: Missing abort variable check
----------------------------------------
Add kernel-local variable to abort_kernel and test it.

Response:
Added `int local = 42` to abort_kernel and added test case `gdb_test "p local" " = 42" "read variable from faulty wave"` for the abort path. This closes the coverage gap between abort and pagefault tests.


Comment 10: Enum naming inconsistency
--------------------------------------
Rename kabort to tabort or pick consistent scheme.

Response:
Skipping this change. Earlier, you requested renaming `abort` to `kabort` to avoid shadowing `::abort()`. Since we reference it as `testcase_t::kabort`, there's no actual shadowing issue, and changing it again would contradict the earlier guidance. The current naming is intentional per earlier feedback.


Comment 11: Missing no_optimized_code guard
--------------------------------------------
Add require guard like runtime-core.exp.

Response:
Added `require no_optimized_code` with comment: "Debug behavior with optimized device code is unreliable; ideally this test should be made to work with optimization as well." This matches runtime-core.exp's pattern and rationale.


Comment 12: Missing device-side data for abort
-----------------------------------------------
Add __device__ global or kernel-local for abort path.

Response:
Same fix as comment 9 - added `int local = 42` to abort_kernel. Combined with the test case in comment 9, this enables verification that core content is accessible for the abort path.

lumachad:
I'll do another pass through this one.

lancesix (commented):
It really feels we could simplify the test.

General testing of coredump generation is already covered by runitme-core.exp, we do not want to duplicate that here.

What this test is looking at is the ability for the runitme to generate a coredump when under debug.

A testcase as simple as

```c++
#include <hip/hip_runtime.h>
__global__ void
kern ()
{
  abort ();
}

int
main ()
{
  kern<<<1, 1>>> ();
  return hipDeviceSynchronize () != hipSuccess;
}
```

would be more than enough.  We run this under the debugger (forwarding the abort to the runtime), then gather the cordumps, merge them and ensure we see the SIGABRT from a wave.

Everything else would be duplicate logic.

amd-bfilipov:
> It really feels we could simplify the test.
> 
> General testing of coredump generation is already covered by runitme-core.exp, we do not want to duplicate that here.
> 
> What this test is looking at is the ability for the runitme to generate a coredump when under debug.
> 
> A testcase as simple as
> 
> ```c++
> #include <hip/hip_runtime.h>
> __global__ void
> kern ()
> {
>   abort ();
> }
> 
> int
> main ()
> {
>   kern<<<1, 1>>> ();
>   return hipDeviceSynchronize () != hipSuccess;
> }
> ```
> 
> would be more than enough. We run this under the debugger (forwarding the abort to the runtime), then gather the cordumps, merge them and ensure we see the SIGABRT from a wave.
> 
> Everything else would be duplicate logic.

Ok. Reduced test to minimum.

lumachad:

## Review summary
Solid, well-motivated test that fills a real gap (core dump generation while GDB is attached and passing signals through). Main issues: the coremerge/merge-verify logic duplicates existing helper functionality in `lib/rocm.exp`, one assertion is tautological, plus a couple of line-length and naming/comment nits.

## GNU coding standards
- `gdb/testsuite/gdb.rocm/runtime-core-attached.exp:70-107` — the glob-for-cores / `coremerge` invocation / existence-check sequence duplicates the tail of `rocm_core_find_1` in `gdb/testsuite/lib/rocm.exp:651-693`: both locate a host core and a matching GPU core, invoke `coremerge` via `gdb_find_coremerge`, and check the merged file exists before proceeding. Suggest factoring this into a shared proc in `lib/rocm.exp`, e.g.:

  ```tcl
  # Merge HOSTCORE and GPUCORE into MERGED_CORE using coremerge.  Returns 1
  # and calls "unsupported" if merging isn't possible (missing coremerge,
  # merge failure, etc), returns 0 on success.
  proc rocm_core_merge { hostcore gpucore merged_core } {
      set coremerge [gdb_find_coremerge]
      set coremerge_result [remote_exec build \
	  "$coremerge $merged_core $hostcore $gpucore"]
      set coremerge_status [lindex $coremerge_result 0]
      set coremerge_output [lindex $coremerge_result 1]

      if {$coremerge_status != 0 || ![remote_file build exists $merged_core]} {
	  unsupported "coremerge failed (status $coremerge_status): $coremerge_output"
	  return 1
      }

      return 0
  }
  ```

  Both `rocm_core_find_1` and `runtime-core-attached.exp` could then call this instead of duplicating the merge/verify steps inline.

- `gdb/testsuite/gdb.rocm/runtime-core-attached.exp:100,105` — exceed the 80-column limit. Suggested wraps:
  ```tcl
  set coremerge_result \
      [remote_exec build "$coremerge $merged_core $hostcore $gpucore"]
  ```
  ```tcl
  if {$coremerge_status != 0
      || ![remote_file build exists $merged_core]} {
  ```

## Cosmetic / style
- `gdb/testsuite/gdb.rocm/runtime-core-attached.exp:21` — "generation. This test verifies..." uses a single space after the sentence-ending period; GNU convention wants two spaces between sentences.
- `gdb/testsuite/gdb.rocm/runtime-core-attached.cpp:22` — kernel named `kern`, whereas the sibling test uses descriptive names (`abort_kernel`, `pagefault_kernel`, `assert_kernel`). Consider `abort_kernel` for consistency.

## Tests
- `gdb/testsuite/gdb.rocm/runtime-core-attached.exp:109` — `gdb_assert {$coremerge_status == 0} "coremerge succeeded"` is unreachable-false: the `if {$coremerge_status != 0 ...} { unsupported ...; return }` right above it already guarantees this, so the assert can never fail and provides no real coverage. Either drop it or replace it with a check that verifies something meaningful (e.g. content of `$coremerge_output`).

amd-bfilipov:
> 
> ## Review summary
> Solid, well-motivated test that fills a real gap (core dump generation while GDB is attached and passing signals through). Main issues: the coremerge/merge-verify logic duplicates existing helper functionality in `lib/rocm.exp`, one assertion is tautological, plus a couple of line-length and naming/comment nits.
> 
> ## GNU coding standards
> * `gdb/testsuite/gdb.rocm/runtime-core-attached.exp:70-107` — the glob-for-cores / `coremerge` invocation / existence-check sequence duplicates the tail of `rocm_core_find_1` in `gdb/testsuite/lib/rocm.exp:651-693`: both locate a host core and a matching GPU core, invoke `coremerge` via `gdb_find_coremerge`, and check the merged file exists before proceeding. Suggest factoring this into a shared proc in `lib/rocm.exp`, e.g.:
>   ```tcl
>   # Merge HOSTCORE and GPUCORE into MERGED_CORE using coremerge.  Returns 1
>   # and calls "unsupported" if merging isn't possible (missing coremerge,
>   # merge failure, etc), returns 0 on success.
>   proc rocm_core_merge { hostcore gpucore merged_core } {
>       set coremerge [gdb_find_coremerge]
>       set coremerge_result [remote_exec build \
>       "$coremerge $merged_core $hostcore $gpucore"]
>       set coremerge_status [lindex $coremerge_result 0]
>       set coremerge_output [lindex $coremerge_result 1]
>   
>       if {$coremerge_status != 0 || ![remote_file build exists $merged_core]} {
>       unsupported "coremerge failed (status $coremerge_status): $coremerge_output"
>       return 1
>       }
>   
>       return 0
>   }
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   Both `rocm_core_find_1` and `runtime-core-attached.exp` could then call this instead of duplicating the merge/verify steps inline.
> * `gdb/testsuite/gdb.rocm/runtime-core-attached.exp:100,105` — exceed the 80-column limit. Suggested wraps:
>   ```tcl
>   set coremerge_result \
>       [remote_exec build "$coremerge $merged_core $hostcore $gpucore"]
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   ```tcl
>   if {$coremerge_status != 0
>       || ![remote_file build exists $merged_core]} {
>   ```
> 
> ## Cosmetic / style
> * `gdb/testsuite/gdb.rocm/runtime-core-attached.exp:21` — "generation. This test verifies..." uses a single space after the sentence-ending period; GNU convention wants two spaces between sentences.
> * `gdb/testsuite/gdb.rocm/runtime-core-attached.cpp:22` — kernel named `kern`, whereas the sibling test uses descriptive names (`abort_kernel`, `pagefault_kernel`, `assert_kernel`). Consider `abort_kernel` for consistency.
> 
> ## Tests
> * `gdb/testsuite/gdb.rocm/runtime-core-attached.exp:109` — `gdb_assert {$coremerge_status == 0} "coremerge succeeded"` is unreachable-false: the `if {$coremerge_status != 0 ...} { unsupported ...; return }` right above it already guarantees this, so the assert can never fail and provides no real coverage. Either drop it or replace it with a check that verifies something meaningful (e.g. content of `$coremerge_output`).

I skipped the comment about editing `gdb/testsuite/lib/rocm.exp:651-693`. Seems unnecessary and it changes an unrelated file. Keeping the scope of pr on this one file only. Other comments are addresed.

gdb/testsuite/gdb.rocm/runtime-core-attached.exp
```diff
@@ -0,0 +1,121 @@
...
+	set merged_core "$coredir/merged.core"
+	set coremerge_result \
+	    [remote_exec build "$coremerge $merged_core $hostcore $gpucore"]
+	set coremerge_status [lindex $coremerge_result 0]
+	set coremerge_output [lindex $coremerge_result 1]
+
+	if {$coremerge_status != 0 \
+		|| ![remote_file build exists $merged_core]} {
+	    unsupported \
+		"coremerge failed (status $coremerge_status): $coremerge_output"
```
lancesix: This form is quite unusual.

It is usually "test name (optionnal reason)".  The parenthesis should be last.

If we want t olog the errors, use `verbose` instead.

amd-bfilipov: Fixed

lancesix:
FYI, running against a GDB built with asan (`-fsanitize=address`), I have a failure:

```
(gdb) PASS: gdb.rocm/runtime-core-attached.exp: set inferior working directory
run
Starting program: /home/lancesix/src/rocm-gdb/_build/gdb/testsuite/outputs/gdb.rocm/runtime-core-attached/runtime-core-attached 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Secondary CUID not available: AMDCUID support not enabled.
[New Thread 0x7fffe89ff6c0 (LWP 264064) (id 2)]
[New Thread 0x7fffe81fe6c0 (LWP 264065) (id 3)]
[New Thread 0x7fffe76ff6c0 (LWP 264066) (id 4)]
[Thread 0x7fffe76ff6c0 (LWP 264066) (id 4) exited]
[New Thread 0x7fffdf3ff6c0 (LWP 264067) (id 5)]

Thread 6 "abort_kernel" received signal SIGABRT, Aborted.
Warning: Queue error: HSA_STATUS_ERROR_EXCEPTION: An HSAIL operation resulted in a hardware exception.
core file size: 0
VGPU(0x55555565fcc0) hang analysis:
SWq=0x7ffff6080000, HWq=0x7fffe5800000, id=1, priority=1, Dispatch Header = 0x1502 (type=2, barrier=1, acquire=2, release=2), setup=3, grid=[1, 1, 1], workgroup=[1, 1, 1], private_seg_size=1024, group_seg_size=0, kernel_obj=0x7ffff7fa8980, kernarg_address=0x7fffe4400080, completion_signal=0x0, rptr=1, wptr=3
:0:rocdevice.cpp            :4457: 9843428485 us:  Callback: Queue 0x7fffe5800000 aborting with error : HSA_STATUS_ERROR_EXCEPTION: An HSAIL operation resulted in a hardware exception. code: 0x1016 [host: khazad-dum, GPU index: 0, kernel: abort_kernel()]

Thread 6 received signal SIGABRT, Aborted.
[Thread 0x7fffdf3ff6c0 (LWP 264067) (id 5) exited]
[Thread 0x7fffe89ff6c0 (LWP 264064) (id 2) exited]
[Thread 0x7fffe81fe6c0 (LWP 264065) (id 3) exited]
[Inferior 1 (process 264056) exited with code 01]
(gdb) FAIL: gdb.rocm/runtime-core-attached.exp: wait for program exit (the program exited)
```

With ASAN, I do not expect full build (ASAN implicitly sets `ulimit -c 0`, which is inherited by the inferior), but the FAIL should be avoided.

amd-bfilipov:
> FYI, running against a GDB built with asan (`-fsanitize=address`), I have a failure:
> 
> ```
> (gdb) PASS: gdb.rocm/runtime-core-attached.exp: set inferior working directory
> run
> Starting program: /home/lancesix/src/rocm-gdb/_build/gdb/testsuite/outputs/gdb.rocm/runtime-core-attached/runtime-core-attached 
> [Thread debugging using libthread_db enabled]
> Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
> Secondary CUID not available: AMDCUID support not enabled.
> [New Thread 0x7fffe89ff6c0 (LWP 264064) (id 2)]
> [New Thread 0x7fffe81fe6c0 (LWP 264065) (id 3)]
> [New Thread 0x7fffe76ff6c0 (LWP 264066) (id 4)]
> [Thread 0x7fffe76ff6c0 (LWP 264066) (id 4) exited]
> [New Thread 0x7fffdf3ff6c0 (LWP 264067) (id 5)]
> 
> Thread 6 "abort_kernel" received signal SIGABRT, Aborted.
> Warning: Queue error: HSA_STATUS_ERROR_EXCEPTION: An HSAIL operation resulted in a hardware exception.
> core file size: 0
> VGPU(0x55555565fcc0) hang analysis:
> SWq=0x7ffff6080000, HWq=0x7fffe5800000, id=1, priority=1, Dispatch Header = 0x1502 (type=2, barrier=1, acquire=2, release=2), setup=3, grid=[1, 1, 1], workgroup=[1, 1, 1], private_seg_size=1024, group_seg_size=0, kernel_obj=0x7ffff7fa8980, kernarg_address=0x7fffe4400080, completion_signal=0x0, rptr=1, wptr=3
> :0:rocdevice.cpp            :4457: 9843428485 us:  Callback: Queue 0x7fffe5800000 aborting with error : HSA_STATUS_ERROR_EXCEPTION: An HSAIL operation resulted in a hardware exception. code: 0x1016 [host: khazad-dum, GPU index: 0, kernel: abort_kernel()]
> 
> Thread 6 received signal SIGABRT, Aborted.
> [Thread 0x7fffdf3ff6c0 (LWP 264067) (id 5) exited]
> [Thread 0x7fffe89ff6c0 (LWP 264064) (id 2) exited]
> [Thread 0x7fffe81fe6c0 (LWP 264065) (id 3) exited]
> [Inferior 1 (process 264056) exited with code 01]
> (gdb) FAIL: gdb.rocm/runtime-core-attached.exp: wait for program exit (the program exited)
> ```
> 
> With ASAN, I do not expect full build (ASAN implicitly sets `ulimit -c 0`, which is inherited by the inferior), but the FAIL should be avoided.

Updated. Test will be marked as UNSUPPORTED if `ulimit` is 0

lumachad:

## Tests

- The PR description says: "Uses batch mode: ... 'handle SIGABRT/SIGSEGV nostop pass'" and "Tests pagefault and abort scenarios," but the actual `.exp`/`.cpp` only configure/exercise `SIGABRT` via `abort()`, no pagefault scenario is present. Either the description overstates the scope, or a pagefault variant was intended but not included. Worth reconciling the PR description with what's actually tested (and clarifying if pagefault coverage is planned as a follow-up).

lumachad:
Not ideal, but we're still not testing anything other than gdb.rocm/runtime-core.exp here. We need this sequencing to happen:

- Merge https://github.com/ROCm/ROCgdb/pull/358
- Bump ROCgdb
- Update TheRock's rocgdb-corefile test to use --gpu-corefile-tests
- Update ROCgdb's TheRock ref to pick that change up. 
- Confirm gdb.rocm/runtime-core-attached.exp gets tested and merge this PR.

PR #217, opened 2026-07-16
ci: add gfx125x to TheRock CI matrix

dan-garvey:
## Summary
- add gfx125X-dcgpu to the ROCgdb TheRock CI build matrix

This keeps gfx125X-dcgpu build-only because the existing test job remains gated to gfx94X-dcgpu.

Depends on ROCm/therock-ci-config#14.

## Testing
- parsed .github/workflows/therock-ci.yml and .github/workflows/therock-ci-linux.yml with PyYAML
- git diff --check

### DIFF
--- a/.github/workflows/therock-ci.yml
+++ b/.github/workflows/therock-ci.yml
@@ -46,7 +46,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        amdgpu_family: [gfx94X-dcgpu]
+        amdgpu_family: [gfx94X-dcgpu, gfx125X-dcgpu]
     uses: ./.github/workflows/therock-ci-linux.yml
     secrets: inherit
     with:


lumachad:
@dan-garvey Do you have an ETA on when we can switch to also running tests on gfx1250? And could we enable gfx950 testing by default?

lumachad:
@dan-garvey We use rebase and merge for ROCgdb. Due to that, could you please add a commit message, as the PR description will be discarded when merging.

lumachad:
@dan-garvey ping?

lumachad:
Converting to draft as we don't seem to have any activity.

dan-garvey:
I dont have any context on 950 capacity/usage/roadmap, so I'll leave that as out of scope for this PR.

lumachad:
I don't think this is needed here. We need it for rocr-debug-agent, as we build gfx-specific code for their tests. For ROCgdb, we're arch-independent, so it wouldn't make sense to do a "gfx1250-specific build".

lumachad:
What would make sense is running tests on gfx1250. But since that's not happening, I'll close this one.

lumachad:
@dan-garvey If you see another reason why we need to do another build of ROCgdb for gfx1250, please let me know and reopen this.

PR #218, opened 2026-07-17
gdb: extend maint print address-spaces to show all dbgapi available info

czidev-amd:
## Motivation

The maint print address-spaces command previously printed only a flat list of address-space names.  Rewrite address_spaces_dump to query dbgapi for five properties per address space and emit them as a structured ui_out table: name, DWARF number, address size in bits, null address, and access class.

## Technical Details

A new helper, address_space_access_to_string, maps the amd_dbgapi_address_space_access_t enum to a human-readable string (all, program-constant, dispatch-constant).  The command help text and the maint print address-spaces section in gdb.texinfo are updated to document the new columns.

## Test Plan

A new test, gdb.rocm/maint-print-address-spaces.exp, runs the command, and checks that a header row and at least one data row are printed without error.

## Submission Checklist

### DIFF
--- a/gdb/amd-dbgapi-target.c
+++ b/gdb/amd-dbgapi-target.c
@@ -4811,21 +4811,153 @@ info_dispatches_command (const char *args, int from_tty)
   gdb_flush (gdb_stdout);
 }
 
+
+/* Helper for decoding and pretty printing the address space access
+   modifiers.  */
+
+static std::string
+address_space_access_to_string (amd_dbgapi_address_space_access_t access)
+{
+  switch (access)
+    {
+    case AMD_DBGAPI_ADDRESS_SPACE_ACCESS_ALL:
+      return "all";
+    case AMD_DBGAPI_ADDRESS_SPACE_ACCESS_PROGRAM_CONSTANT:
+      return "program-constant";
+    case AMD_DBGAPI_ADDRESS_SPACE_ACCESS_DISPATCH_CONSTANT:
+      return "dispatch-constant";
+    }
+  return string_printf ("unknown (0x%x)", access);
+}
+
 /* Dump out a table of address spaces for the current architecture.  */
 
 static void
-address_spaces_dump (struct gdbarch *gdbarch, struct ui_file *file)
+address_spaces_dump (struct gdbarch *gdbarch)
 {
   if (!gdbarch_address_spaces_p (gdbarch))
     return;
 
-  auto address_spaces = gdbarch_address_spaces (gdbarch);
+  /* Get the architecture id.  */
+  amd_dbgapi_architecture_id_t architecture_id;
+  amd_dbgapi_status_t status = amd_dbgapi_get_architecture
+    (gdbarch_bfd_arch_info (gdbarch)->mach, &architecture_id);
+  if (status != AMD_DBGAPI_STATUS_SUCCESS)
+    error (_("amd_dbgapi_get_architecture failed (%s)"),
+	   get_status_string (status));
 
-  gdb_printf (file, " Name\n");
+  /* Get info about address spaces.  */
+  size_t address_space_count;
+  amd_dbgapi_address_space_id_t *address_spaces;
+  status = amd_dbgapi_architecture_address_space_list (architecture_id,
+						       &address_space_count,
+						       &address_spaces);
+  if (status != AMD_DBGAPI_STATUS_SUCCESS)
+    error (_("amd_dbgapi_architecture_address_space_list failed (%s)"),
+	   get_status_string (status));
 
-  for (const auto &address_space : address_spaces)
+  /* Wrap the returned array.  */
+  gdb::unique_xmalloc_ptr<amd_dbgapi_address_space_id_t[]> address_spaces_holder
+    (address_spaces);
+
+  /* "0x" (2) + two hex digits per byte of pointer (ptr_bit/4) - 1 space
+     that ui_out inserts between columns automatically.  */
+  const int addr_width = 4 + (gdbarch_ptr_bit (gdbarch) / 4) - 1;
+
+  /* Collect all address-space properties in one pass so each dbgapi
+     property is queried exactly once.  */
+  struct aspace_info
+  {
+    std::string name;
+    arch_addr_space_id dwarf_num;
+    amd_dbgapi_size_t addr_size;
+    amd_dbgapi_segment_address_t null_address;
+    std::string access;
+  };
+  std::vector<aspace_info> aspaces;
+  aspaces.reserve (address_space_count);
+
+  for (size_t i = 0; i < address_space_count; i++)
     {
-      gdb_printf (file, " %-10s\n", address_space.name.get ());
+      aspace_info info;
+
+      char *raw_name;
+      status = amd_dbgapi_address_space_get_info
+	(address_spaces[i], AMD_DBGAPI_ADDRESS_SPACE_INFO_NAME,
+	 sizeof (raw_name), &raw_name);
+      if (status != AMD_DBGAPI_STATUS_SUCCESS)
+	error (_("amd_dbgapi_address_space_get_info (name) failed (%s)"),
+	       get_status_string (status));
+      if (raw_name == nullptr)
+	error (_("amd_dbgapi_address_space_get_info (name) returned "
+		 "a null pointer"));
+      gdb::unique_xmalloc_ptr<char> name_holder (raw_name);
+      info.name = raw_name;
+
+      status = amd_dbgapi_address_space_get_info
+	(address_spaces[i], AMD_DBGAPI_ADDRESS_SPACE_INFO_DWARF,
+	 sizeof (info.dwarf_num), &info.dwarf_num);
+      if (status != AMD_DBGAPI_STATUS_SUCCESS)
+	error (_("amd_dbgapi_address_space_get_info (dwarf) failed (%s)"),
+	       get_status_string (status));
+
+      status = amd_dbgapi_address_space_get_info
+	  (address_spaces[i], AMD_DBGAPI_ADDRESS_SPACE_INFO_ADDRESS_SIZE,
+	   sizeof (info.addr_size), &info.addr_size);
+      if (status != AMD_DBGAPI_STATUS_SUCCESS)
+	error (_("amd_dbgapi_address_space_get_info (address_size) failed (%s)"),
+	       get_status_string (status));
+
+      status = amd_dbgapi_address_space_get_info
+	  (address_spaces[i], AMD_DBGAPI_ADDRESS_SPACE_INFO_NULL_ADDRESS,
+	   sizeof (info.null_address), &info.null_address);
+      if (status != AMD_DBGAPI_STATUS_SUCCESS)
+	error (_("amd_dbgapi_address_space_get_info (null_address) failed (%s)"),
+	       get_status_string (status));
+
+      amd_dbgapi_address_space_access_t address_access;
+      status = amd_dbgapi_address_space_get_info
+	  (address_spaces[i], AMD_DBGAPI_ADDRESS_SPACE_INFO_ACCESS,
+	   sizeof (address_access), &address_access);
+      if (status != AMD_DBGAPI_STATUS_SUCCESS)
+	error (_("amd_dbgapi_address_space_get_info (access) failed (%s)"),
+	       get_status_string (status));
+      info.access = address_space_access_to_string (address_access);
+
+      aspaces.push_back (std::move (info));
+    }
+
+  /* Compute column widths from the collected data.  */
+  size_t max_name_len = strlen (_("Name"));
+  size_t max_access_len = strlen (_("Access"));
+  for (const auto &info : aspaces)
+    {
+      max_name_len = std::max (max_name_len, info.name.size ());
+      max_access_len = std::max (max_access_len, info.access.size ());
+    }
+
+  struct ui_out *ui_out = current_uiout;
+  ui_out_emit_table table_emitter (ui_out, 5, address_space_count,
+				   "AddressSpacesTable");
+
+  ui_out->table_header ((int) max_name_len, ui_left, "name", _("Name"));
+  ui_out->table_header (5, ui_right, "dwarf", _("DWARF"));
+  ui_out->table_header (9, ui_right, "addr", _("Addr Bits"));
+  ui_out->table_header (addr_width, ui_right, "nulladdr",
+			_("Null Address"));
+  ui_out->table_header ((int) max_access_len, ui_right, "access", _("Access"));
+  ui_out->table_body ();
+
+  for (const auto &info : aspaces)
+    {
+      ui_out_emit_tuple tuple_emitter (ui_out, "space");
+
+      ui_out->field_string ("name", info.name);
+      ui_out->field_unsigned ("dwarf", info.dwarf_num);
+      ui_out->field_unsigned ("addr", info.addr_size);
+      ui_out->field_fmt ("nulladdr", "0x%s", phex_nz (info.null_address, 0));
+      ui_out->field_string ("access", info.access);
+      ui_out->text ("\n");
     }
 }
 
@@ -4834,7 +4966,7 @@ address_spaces_dump (struct gdbarch *gdbarch, struct ui_file *file)
 static void
 maintenance_print_address_spaces (const char *args, int from_tty)
 {
-  address_spaces_dump (get_current_arch (), gdb_stdout);
+  address_spaces_dump (get_current_arch ());
 }
 
 static void
@@ -5113,6 +5245,8 @@ Otherwise, all dispatches are displayed."),
 
   add_cmd ("address-spaces", class_maintenance,
 	   maintenance_print_address_spaces, _("\
-Displays the address space names supported by the current architecture."),
+Display properties of address-spaces supported by the current architecture.\n\
+For each address space, print its name, DWARF id, address size,\n\
+null address, and access class."),
 	   &maintenanceprintlist);
 }
--- a/gdb/doc/gdb.texinfo
+++ b/gdb/doc/gdb.texinfo
@@ -25242,19 +25242,49 @@ lanes are inactive.
 @c the @code{maint print reggroups} command which also should not be a
 @c maintenance command for the same reason.
 @item maint print address-spaces
-@code{maint print address-spaces} displays the address space names
-supported by the current architecture.
+@code{maint print address-spaces} displays the properties of all address
+spaces supported by the current architecture.  For each address space the
+following columns are shown:
 
-The address spaces info looks like this:
+@table @asis
+@item Name
+The address-space name as returned by the architecture support library.
+
+@item DWARF
+The DWARF address-space number used as the address space identifier in
+DWARF debug information.
+
+@item Addr Bits
+The size of an address in this address space, in bits.
+
+@item Null Address
+The null (invalid) segment address for this address space, printed in
+hexadecimal.
+
+@item Access
+The access class for this address space.  One of:
+@table @code
+@item all
+The address space supports all accesses (reads and writes).
+@item program-constant
+The address space is read-only; its contents are constant for the lifetime
+of the program execution.
+@item dispatch-constant
+The address space is read-only for the waves of a kernel dispatch; its
+contents are constant for the lifetime of a single dispatch.
+@end table
+@end table
+
+The address spaces table looks like this:
 
 @smallexample
 (@value{GDBP}) @kbd{maint print address-spaces}
- Name
- global
- generic
- local
- private_lane
- private_wave
+Name              DWARF  Addr Bits    Null Address  Access
+global                0         64             0x0     all
+generic               1         64             0x0     all
+local                 3         32      0xffffffff     all
+private_lane          5         32      0xffffffff     all
+private_wave          6         32      0xffffffff     all
 @end smallexample
 
 Any expression that evaluates to an integral value, can be used
--- a/gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
+++ b/gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
@@ -1,0 +1,93 @@
+# Copyright (C) 2026 Free Software Foundation, Inc.
+# Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+# This file is part of GDB.
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Baseline test for "maintenance print address-spaces".
+# For each known amdgcn architecture, verify the command runs without error
+# and produces a header row and at least one data row.
+
+clean_restart
+
+# Collect all amdgcn:gfx* architecture names that GDB knows about.
+set all_archs {}
+gdb_test_multiple "complete set architecture amdgcn:" "" {
+    -re "set architecture (amdgcn:\[^\r\n\]+)\r\n" {
+	lappend all_archs $expect_out(1,string)
+	exp_continue
+    }
+    -re "$::gdb_prompt $" {
+	pass $gdb_test_name
+    }
+}
+
+set all_archs [lsort -unique $all_archs]
+
+if {[llength $all_archs] == 0} {
+    unsupported "no amdgcn architectures found"
+    return
+}
+
+set asname "\[^ \t\r\n\]+\[ \t\]+"
+set asdec "$::decimal\[ \t\]+"
+set ashex "$::hex\[ \t\]+"
+set asaccess "(all|program-constant|dispatch-constant|unknown \\($::hex\\))"
+set trailer "\[^\r\n\]*(?=\r\n)"
+
+foreach_with_prefix arch $all_archs {
+    clean_restart
+
+    set arch_supported 0
+
+    gdb_test_multiple "set architecture $arch" "set arch" {
+	-re -wrap "The target architecture is set to \"$arch\"\\." {
+	    pass $gdb_test_name
+	    set arch_supported 1
+	}
+	-re -wrap "Undefined item: \"$arch\"\\." {
+	    untested $gdb_test_name
+	}
+    }
+
+    if {!$arch_supported} {
+	continue
+    }
+
+    set seen_header 0
+    set seen_address 0
+
+    # The actual values aren't very important here; we want to verify that GDB
+    # doesn't crash or error on any supported architecture.
+    gdb_test_multiple "maintenance print address-spaces" "" -lbl {
+	-re "Name\[ \t\]+DWARF\[ \t\]+Addr Bits\[ \t\]+Null Address\[ \t\]+Access${trailer}" {
+	    # Match the header line (anchored to start of line via -lbl).
+	    incr seen_header
+	    exp_continue
+	}
+
+	-re "\r\n${asname}${asdec}${asdec}${ashex}${asaccess}${trailer}" {
+	    # Match a data row.
+	    incr seen_address
+	    exp_continue
+	}
+
+	-re -wrap "" {
+	    # Done.
+	    gdb_assert { $seen_header == 1 } "$gdb_test_name: seen header"
+	    gdb_assert { $seen_address > 0 } "$gdb_test_name: seen address"
+	}
+    }
+}


gdb/doc/gdb.texinfo
```diff
@@ -25042,19 +25242,49 @@ lanes are inactive.
...
 
 @smallexample
 (@value{GDBP}) @kbd{maint print address-spaces}
- Name
- global
- generic
- local
- private_lane
- private_wave
+Name              DWARF  Addr Bits    Null Address  Access
```
aktemur, line 25282: The values in the "Addr Bits" column look like bytes.

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
```diff
@@ -0,0 +1,57 @@
...
+# Baseline test for "maintenance print address-spaces".
+
+clean_restart
+
+gdb_test_multiple "set architecture amdgcn:gfx90a" "set arch" {
+    -re -wrap "The target architecture is set to \"amdgcn:gfx90a\"." {
+	pass $gdb_test_name
+    }
+
+    -re -wrap "Undefined item: \"amdgcngfx90a\"." {
```
aktemur: I think there needs to be a colon `:` between amdgcn and gfx90a.

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
```diff
@@ -0,0 +1,57 @@
...
+    }
+}
+
+set seen_header 0
+set seen_address 0
+
+# The actual output isn't very important, we want to verify that GDB
+# doesn't have any latent error.  Morevover, the number of address
+# spaces can vary, if one is setting a different architecture.
+gdb_test_multiple "maintenance print address-spaces" "list all address spaces" -lbl {
```
aktemur: Nit: Test name can be left empty, in which case GDB will use the command as the test name and the command itself here is just fine as a test name, I think. The line would be shortened, too.

.gitignore
```diff
@@ -75,3 +75,4 @@ stamp-*
 /gmp*
 /isl*
 /gettext*
+.worktrees/
```
aktemur: This seems to have sneaked in.

gdb/amd-dbgapi-target.c
```diff
@@ -5113,6 +5234,6 @@ Otherwise, all dispatches are displayed."),
 
   add_cmd ("address-spaces", class_maintenance,
 	   maintenance_print_address_spaces, _("\
-Displays the address space names supported by the current architecture."),
+Displays address-space properties (name, DWARF id, address size, null address, and access) supported by the current architecture."),
```
aktemur: Let's please break the line.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,142 @@ info_dispatches_command (const char *args, int from_tty)
...
+      != AMD_DBGAPI_STATUS_SUCCESS)
+    error (_("amd_dbgapi_architecture_address_space_list failed"));
 
-  gdb_printf (file, " Name\n");
+  /* Wrap the returned array.  */
+  gdb::unique_xmalloc_ptr<amd_dbgapi_address_space_id_t[]> address_spaces_holder
+    (address_spaces);
 
-  for (const auto &address_space : address_spaces)
+   /* "0x", a little whitespace, and two hex digits per byte of pointers.  */
```
aktemur: Indentation is one space too much.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,142 @@ info_dispatches_command (const char *args, int from_tty)
...
+    }
+
+  /* Use UI to output.  */
+  struct ui_out *ui_out = current_uiout;
+  ui_out_emit_table table_emitter (ui_out, 5, address_space_count,
+				   "AddressSpacesTable");
+
+  ui_out->table_header (max_name_len, ui_left, "name", _("Name"));
+  /* The "- 1" is because ui_out adds one space between columns.  */
+  ui_out->table_header (addr_width - 1, ui_right, "dwarf", _("DWARF"));
```
aktemur: I don't follow why we use "addr_width - 1" here for the width. Can we just use 5, which is the length of "DWARF". The DWARF id wouldn't be larger than that.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,142 @@ info_dispatches_command (const char *args, int from_tty)
...
+  ui_out->table_header (10, ui_right, "addr", _("Addr Bits"));
```
aktemur: The length of "Addr Bits" is 9.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,142 @@ info_dispatches_command (const char *args, int from_tty)
...
+  ui_out->table_header (addr_width - 1, ui_right, "nulladdr",
+			_("Null Address"));
+  ui_out->table_header (17, ui_right, "access", _("Access"));
```
aktemur: Nit: In the first pass above where we find out max name length, we can calculate max access length, too.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,142 @@ info_dispatches_command (const char *args, int from_tty)
...
+    {
+      char *address_space_name;
+
+      if (amd_dbgapi_address_space_get_info
+	    (address_spaces[i], AMD_DBGAPI_ADDRESS_SPACE_INFO_NAME,
+	     sizeof (address_space_name), &address_space_name)
+	  != AMD_DBGAPI_STATUS_SUCCESS)
+	error (_("amd_dbgapi_address_space_get_info (name) failed"));
+
+      gdb_assert (address_space_name != nullptr);
```
aktemur: Given that the name is set by the library, which is an external component to GDB, I'd suggest not asserting. We can print `<null>` instead.

gdb/doc/gdb.texinfo
czidev-amd: In the gdbapi documentation we have "Return the byte size of an address in the address space." However, the tests results show me something different:

```
 Name                       DWARF  Addr Bits        Null Address            Access
 global                         0         64                 0x0               all
 generic                        1         64                 0x0               all
 private_wave                   6         32          0xffffffff               all
 private_lane                   5         32          0xffffffff               all
 local                          3         32          0xffffffff               all
```
I need to fix the example and fix the documentation in dbgapi.

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
czidev-amd: Thank you.

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
```diff
@@ -0,0 +1,57 @@
...
+}
+
+set seen_header 0
+set seen_address 0
+
+# The actual output isn't very important, we want to verify that GDB
+# doesn't have any latent error.  Morevover, the number of address
+# spaces can vary, if one is setting a different architecture.
+gdb_test_multiple "maintenance print address-spaces" "" -lbl {
+    -re "Name\[ \t\]+DWARF\[ \t\]+Addr Bits\[ \t\]+Null Address\[ \t\]+Access\[ \t]*" {
```
lancesix: Might be worth anchoring this so it needs to read from the start of the buffer (start of a line thanks to -lbl) until the end of that line.

See #196 for an example.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,156 @@ info_dispatches_command (const char *args, int from_tty)
...
+  switch (access)
+    {
+    case AMD_DBGAPI_ADDRESS_SPACE_ACCESS_ALL:
+      return "all";
+    case AMD_DBGAPI_ADDRESS_SPACE_ACCESS_PROGRAM_CONSTANT:
+      return "program-constant";
+    case AMD_DBGAPI_ADDRESS_SPACE_ACCESS_DISPATCH_CONSTANT:
+      return "dispatch-constant";
+    default:
+      return "unknown";
```
lancesix: should we print the raw value in this case? Could help debug.

Something like "unknown (0x42)"

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,156 @@ info_dispatches_command (const char *args, int from_tty)
...
 
-  gdb_printf (file, " Name\n");
+  /* "0x", a little whitespace, and two hex digits per byte of pointers.  */
+  int addr_width = 4 + (gdbarch_ptr_bit (gdbarch) / 4);
 
-  for (const auto &address_space : address_spaces)
+  /* First pass: compute the maximum name length for column sizing.
+     Initialise with the header string width so the column is never
+     narrower than the title.  */
+  size_t max_name_len = strlen ("Name");
```
lancesix: The code below will use `_("Name")`, not `"Name"`.  I don't think translations have been kept for a long time, but this should be the string given by gettext, not just the key.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,156 @@ info_dispatches_command (const char *args, int from_tty)
...
+  size_t max_access_len = strlen ("Access");
```
lancesix: Same.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,156 @@ info_dispatches_command (const char *args, int from_tty)
...
+	  gdb::unique_xmalloc_ptr<char> name_holder (address_space_name);
+	  max_name_len = std::max (max_name_len, strlen (address_space_name));
+	}
+
+      amd_dbgapi_address_space_access_t address_access;
+      if (amd_dbgapi_address_space_get_info
+	  (address_spaces[i], AMD_DBGAPI_ADDRESS_SPACE_INFO_ACCESS,
+	   sizeof (address_access), &address_access)
+	  == AMD_DBGAPI_STATUS_SUCCESS)
+	{
```
lancesix: in pure GNU style, you only have one statement in this block, it sholud not have {}.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,156 @@ info_dispatches_command (const char *args, int from_tty)
...
+	  max_access_len = std::max
+	    (max_access_len,
+	     strlen (address_space_access_to_string (address_access)));
+	}
+
```
lancesix, line 4926: extra empty line.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,156 @@ info_dispatches_command (const char *args, int from_tty)
...
+     NAME, ADDRESS_SIZE, NULL_ADDRESS, ACCESS, and DWARF.  */
+  for (size_t i = 0; i < address_space_count; i++)
+    {
+      char *address_space_name;
+
+      if (amd_dbgapi_address_space_get_info
+	    (address_spaces[i], AMD_DBGAPI_ADDRESS_SPACE_INFO_NAME,
+	     sizeof (address_space_name), &address_space_name)
+	  != AMD_DBGAPI_STATUS_SUCCESS)
+	error (_("amd_dbgapi_address_space_get_info (name) failed"));
```
lancesix: Just noting: you call this twice.  The first time, errors are ignored, but not the second.

It also feels a bit of overkill to call into dbgapi twice per property.  Shouldn't we remember thew for he duration of this function?

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,156 @@ info_dispatches_command (const char *args, int from_tty)
...
+
+      address_space_name =
+	address_space_name != nullptr ? address_space_name : _("<null>");
```
lancesix: the width calculation did not account for the `_("<null>")` size.  Is it valid for dbgapi to give us a null name?  Seems to me this should not be valid, and could result in an error.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,156 @@ info_dispatches_command (const char *args, int from_tty)
...
+	  (address_spaces[i], AMD_DBGAPI_ADDRESS_SPACE_INFO_ACCESS,
+	   sizeof (address_access), &address_access)
+	  != AMD_DBGAPI_STATUS_SUCCESS)
+	error (_("amd_dbgapi_address_space_get_info (access) failed"));
+
+      /* We have all the info, print it.  */
+      ui_out_emit_tuple tuple_emitter (ui_out, "space");
+
+      ui_out->field_string ("name", address_space_name_holder.get ());
+      ui_out->field_fmt ("dwarf", "%lu", (unsigned long) address_space_dwarf_num);
```
lancesix: The address space number is a uint64_t.  This works on Linux as "unsigned long" is compatible, but on Windows this would loose precision.  Prefer `"0x%s"` with `phex`-style function instead.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,156 @@ info_dispatches_command (const char *args, int from_tty)
...
+      ui_out->field_fmt ("addr", "%lu", (unsigned long) address_byte_size);
```
lancesix: same.

gdb/amd-dbgapi-target.c
czidev-amd: So, Shall I revert back to a gdb_assert?

lancesix: gdb_assert would probably not be the right tool. assertions are to ensure internal invariant, this would be validating external data.  So either this (after thinking about is probably ok) or use

```
if (address_space_name == nullptr)
    error (_("Invalid data from dbgapi...."));
```

czidev-amd: I'll use the error, as I am very curious if someone will bump in this issue, and an error is easier to spot and report.

czidev-amd:
Pushed a new version with the next changes:

- address_space_access_to_string: change return type to std::string and print the raw value in hex for the unknown case, e.g. "unknown (0x42)",to aid debugging.

- address_spaces_dump: collect all dbgapi properties in a single pass into an aspace_info vector, so each property is queried exactly once. Width calculation and printing both iterate the cached vector.  This eliminates the double-query-per-property pattern and the single-statement braced if blocks in the old first-pass loop.

- Width sentinels: use strlen(_("Name")) / strlen(_("Access")) instead of the bare string literals so the gettext result is measured, not the key.

- Numeric fields: switch dwarf and addr columns from field_fmt("%lu") to field_unsigned(), which is portable (no LLP64 truncation on Windows) and preserves correct right-alignment.

- Test: anchor the header regex so -lbl matching fires on a full header line, not a substring buried elsewhere in the output.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,147 @@ info_dispatches_command (const char *args, int from_tty)
...
   if (!gdbarch_address_spaces_p (gdbarch))
     return;
 
-  auto address_spaces = gdbarch_address_spaces (gdbarch);
+  /* Get the architecture id.  */
+  amd_dbgapi_architecture_id_t architecture_id;
+  if (amd_dbgapi_get_architecture
+      (gdbarch_bfd_arch_info (gdbarch)->mach, &architecture_id)
+      != AMD_DBGAPI_STATUS_SUCCESS)
+    error (_("amd_dbgapi_get_architecture failed"));
```
amd-shahab: new line here after `error ...` line.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,147 @@ info_dispatches_command (const char *args, int from_tty)
...
+  /* Get info about address spaces.  */
+  size_t address_space_count;
+  amd_dbgapi_address_space_id_t *address_spaces;
+
```
amd-shahab: remove this empty line.
following the same construct/form for "architecture_id".

amd-shahab (approved):

czidev-amd:
Thank you @amd-shahab for your input, I have fix that and each error message will contain the returned error status too (as Pedro requested in https://github.com/ROCm/ROCgdb/pull/225)

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,154 @@ info_dispatches_command (const char *args, int from_tty)
...
+      char *raw_name;
+      status = amd_dbgapi_address_space_get_info
+	(address_spaces[i], AMD_DBGAPI_ADDRESS_SPACE_INFO_NAME,
+	 sizeof (raw_name), &raw_name);
+      if (status != AMD_DBGAPI_STATUS_SUCCESS)
+	error (_("amd_dbgapi_address_space_get_info (name) failed (%s)"),
+	       get_status_string (status));
+      if (raw_name == nullptr)
+	error (_("Invalid data from dbgapi...."));
+      gdb::unique_xmalloc_ptr<char> name_holder (raw_name);
```
lumachad, line 4894: `"Invalid data from dbgapi...."` has four dots. Use three for an ellipsis, or write a more informative message such as `"amd_dbgapi_address_space_get_info (name) returned a null pointer"`.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,154 @@ info_dispatches_command (const char *args, int from_tty)
...
+  size_t max_access_len = strlen (_("Access"));
+  for (const auto &info : aspaces)
+    {
+      max_name_len = std::max (max_name_len, info.name.size ());
+      max_access_len = std::max (max_access_len, info.access.size ());
+    }
 
-  for (const auto &address_space : address_spaces)
+  /* Use UI to output.  */
+  struct ui_out *ui_out = current_uiout;
```
lumachad, line 4939: `/* Use UI to output.  */` states the obvious. Drop it.

lumachad (commented):
Just nits. Take it or not, your choice.

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
```diff
@@ -0,0 +1,63 @@
...
+
+set asname "^\r\n\[a-z_\]+\[ \t\]+"
+set asdec "$::decimal\[ \t\]+"
+set ashex "$::hex\[ \t\]+"
+set asaccess "(all|program-constant|dispatch-constant)"
+set trailer "\[^\r\n\]*(?=\r\n)"
+
+# The actual output isn't very important, we want to verify that GDB
+# doesn't have any latent error.  Moreover, the number of address
+# spaces can vary, if one is setting a different architecture.
```
palves: What does "setting a different architecture" mean?

Did you mean "testing"?

We do "set architecture amdgcn:gfx90a" above, though, so the architecture is fixed.  So how can the premise that the output varies be true?

czidev-amd: Trailing message from previous refactoring, removing.

palves: Well, that still leaves the whole premise that lead to not parsing actual specific address spaces incorrect -- if we're hard coding one specific architecture, then we could make the testcase look closer to the address spaces that came out.  OTOH, maybe we want instead to loop over all architectures and issue the command, to confirm that we (gdb, dbgapi, etc.) don't do anything bad in any.

czidev-amd: That (having a strict testcase for a specific architecture) was my initial draft, however, after speaking with Lancelot, I've simplified the test case to what we have right now. However, I can loop over a list of known architectures, and check if we crash or something of a sort.

gdb/amd-dbgapi-target.c
czidev-amd: Removed the dot.

gdb/amd-dbgapi-target.c
czidev-amd: done. thank you

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
lancesix: The issue with testing the values is: aperture configuration are not part of the ABI.  Those are not expected to change often, but they could, and are different on Windows and Linux.

We could have "probe" kernels which we use to figure out the configuration and deduce valid conversions we want to test for if we really wanted to test conversion.

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
```diff
@@ -0,0 +1,62 @@
...
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Baseline test for "maintenance print address-spaces".
+
+clean_restart
+
+gdb_test_multiple "set architecture amdgcn:gfx90a" "set arch" {
+    -re -wrap "The target architecture is set to \"amdgcn:gfx90a\"." {
```
lancesix: while at it, should we test for all amdgcn:* architectures.

```
complete set architecture amdgcn:
```

should give us the full list.

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
```diff
@@ -0,0 +1,62 @@
...
+    -re -wrap "Undefined item: \"amdgcn:gfx90a\"." {
+	untested $gdb_test_name
+	return
+    }
+}
+
+set seen_header 0
+set seen_address 0
+
+set asname "^\r\n\[a-z_\]+\[ \t\]+"
```
lancesix: minor thing, but `a-z_` might be a bit restrictive.  I do not think we have any definition of what is a valid address space (no formal grammar for this).  This, although valid today, seems more restrictive than needed.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,153 @@ info_dispatches_command (const char *args, int from_tty)
...
+{
+  switch (access)
+    {
+    case AMD_DBGAPI_ADDRESS_SPACE_ACCESS_ALL:
+      return "all";
+    case AMD_DBGAPI_ADDRESS_SPACE_ACCESS_PROGRAM_CONSTANT:
+      return "program-constant";
+    case AMD_DBGAPI_ADDRESS_SPACE_ACCESS_DISPATCH_CONSTANT:
+      return "dispatch-constant";
+    default:
```
lancesix: I would remove the `default` out of here:

```c++
  switch (access)
    {
    case AMD_DBGAPI_ADDRESS_SPACE_ACCESS_ALL:
      return "all";
    case AMD_DBGAPI_ADDRESS_SPACE_ACCESS_PROGRAM_CONSTANT:
      return "program-constant";
    case AMD_DBGAPI_ADDRESS_SPACE_ACCESS_DISPATCH_CONSTANT:
      return "dispatch-constant";
  }
return string_printf ("unknown (0x%x)", access);
```

This way, if the `amd_dbgapi_address_space_access_t` is extended in the future, the compiler can warn us, but we still support the fallback if the library returns an unexpected value.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,153 @@ info_dispatches_command (const char *args, int from_tty)
...
+
+      char *raw_name;
+      status = amd_dbgapi_address_space_get_info
+	(address_spaces[i], AMD_DBGAPI_ADDRESS_SPACE_INFO_NAME,
+	 sizeof (raw_name), &raw_name);
+      if (status != AMD_DBGAPI_STATUS_SUCCESS)
+	error (_("amd_dbgapi_address_space_get_info (name) failed (%s)"),
+	       get_status_string (status));
+      if (raw_name == nullptr)
+	error (_("Invalid data from dbgapi..."));
```
lancesix: might be worth giving a bit more info, so one could read the error and know where to go and look.

gdb/amd-dbgapi-target.c
```diff
@@ -5113,6 +5245,7 @@ Otherwise, all dispatches are displayed."),
 
   add_cmd ("address-spaces", class_maintenance,
 	   maintenance_print_address_spaces, _("\
-Displays the address space names supported by the current architecture."),
+Displays address-space properties (name, DWARF id, address size, \n\
+null address, and access) supported by the current architecture."),
```
palves: Removing the parens bit to make the following point clearer:

> Displays address-space properties supported by the current architecture."

This reads as:

> Gives the name of the set of address-space properties that are supported by the current architecture."

It should instead be something like:

> Displays properties of address-spaces supported by the current architecture."

BTW, GDB command help uses the imperative -- "Display", "Use", "Set", etc., not third-person present.  Let's fix that while we're here.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,153 @@ info_dispatches_command (const char *args, int from_tty)
...
+  struct ui_out *ui_out = current_uiout;
+  ui_out_emit_table table_emitter (ui_out, 5, address_space_count,
+				   "AddressSpacesTable");
 
-  gdb_printf (file, " Name\n");
+  ui_out->table_header ((int) max_name_len, ui_left, "name", _("Name"));
+  ui_out->table_header (5, ui_right, "dwarf", _("DWARF"));
+  ui_out->table_header (9, ui_right, "addr", _("Addr Bits"));
+  /* The "- 1" is because ui_out adds one space between columns.  */
+  ui_out->table_header (addr_width - 1, ui_right, "nulladdr",
```
lancesix: `addr_width` is statically computed above, and only used here.  This -1 could be included in the declaration above.

gdb/amd-dbgapi-target.c
```diff
@@ -4811,21 +4811,153 @@ info_dispatches_command (const char *args, int from_tty)
...
+  if (status != AMD_DBGAPI_STATUS_SUCCESS)
+    error (_("amd_dbgapi_architecture_address_space_list failed (%s)"),
+	   get_status_string (status));
+
+  /* Wrap the returned array.  */
+  gdb::unique_xmalloc_ptr<amd_dbgapi_address_space_id_t[]> address_spaces_holder
+    (address_spaces);
+
+  /* "0x", a little whitespace, and two hex digits per byte of pointers.  */
+  int addr_width = 4 + (gdbarch_ptr_bit (gdbarch) / 4);
```
lancesix: could be const.  This is never modified.

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
palves: You are thinking of the address alias PR instead?  This PR is about listing this:

```
Name              DWARF  Addr Bits    Null Address  Access
global                0         64             0x0     all
generic               1         64             0x0     all
local                 3         32      0xffffffff     all
private_lane          5         32      0xffffffff     all
private_wave          6         32      0xffffffff     all
```

lancesix: ah, sorry for the confusion.  Yes, I was thinking about the alias PR.

The list of address space supported is a per architecture, but we could teach GDB what to expect for here, and match the actual output.

gdb/amd-dbgapi-target.c
czidev-amd: fixed

gdb/amd-dbgapi-target.c
czidev-amd: fixed

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
czidev-amd: I have relaxed it and I accept anything but `\r\n`

gdb/amd-dbgapi-target.c
czidev-amd: Done

gdb/amd-dbgapi-target.c
czidev-amd: Done

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
czidev-amd: Done

gdb/amd-dbgapi-target.c
czidev-amd: Done

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
czidev-amd: I have updated the test, now it loops over the gdb's known amdgpu arches and executes for each one the maintenance command. Please let me know if you want to strength more up the test. For example, null address can be either 0x0+ or 0xf+

lancesix (approved):
Thanks, this LGTM.

Maybe give some time to other people who contributed reviews before merging see if they are happy to.

gdb/amd-dbgapi-target.c
```diff
@@ -5113,6 +5245,8 @@ Otherwise, all dispatches are displayed."),
 
   add_cmd ("address-spaces", class_maintenance,
 	   maintenance_print_address_spaces, _("\
-Displays the address space names supported by the current architecture."),
+Display address-space properties supported by the current architecture.\n\
+For each address space, print its name, DWARF id, address size,\n\
+null address, and access class."),
```
palves, line 5250: This still has the same problem:

> Display address-space properties supported by the current architecture.\n\

The set of properties is fixed!  What varies is the address spaces.

> For each address space, print its name, DWARF id, address size,\n\
> null address, and access class."),

So this would be more correct:

> Display properties of address spaces supported by the current architecture.\n\
> For each address space, print its name, DWARF id, address size,\n\
> null address, and access class."),

gdb/doc/gdb.texinfo
```diff
@@ -25068,19 +25242,49 @@ lanes are inactive.
...
+spaces supported by the current architecture.  For each address space the
+following columns are shown:
+
+@table @asis
+@item Name
+The address-space name as returned by the architecture support library.
+
+@item DWARF
+The DWARF address-space number used to qualify addresses in DWARF debug
+information.
```
aktemur: Nit: I suggest the following rewording:
The DWARF address-space number used as the address space identifier in DWARF debug information.

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
```diff
@@ -0,0 +1,92 @@
...
+}
+
+set asname "\r\n\[^ \t\r\n\]+\[ \t\]+"
+set asdec "$::decimal\[ \t\]+"
+set ashex "$::hex\[ \t\]+"
+set asaccess "(all|program-constant|dispatch-constant|unknown \\($::hex\\))"
+set trailer "\[^\r\n\]*(?=\r\n)"
+
+foreach_with_prefix arch $all_archs {
+    clean_restart
```
aktemur, line 51: Do we really need to restart for each arch?

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
```diff
@@ -0,0 +1,92 @@
...
+}
+
+set all_archs [lsort -unique $all_archs]
+
+if {[llength $all_archs] == 0} {
+    unsupported "no amdgcn architectures found"
+    return
+}
+
+set asname "\r\n\[^ \t\r\n\]+\[ \t\]+"
```
aktemur: To be more precise, the leading `\r\n` is not part of the name. I suggest we remove it from the regexp here and instead...

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
```diff
@@ -0,0 +1,92 @@
...
+    # The actual values aren't very important here; we want to verify that GDB
+    # doesn't crash or error on any supported architecture.
+    gdb_test_multiple "maintenance print address-spaces" "" -lbl {
+	-re "Name\[ \t\]+DWARF\[ \t\]+Addr Bits\[ \t\]+Null Address\[ \t\]+Access${trailer}" {
+	    # Match the header line (anchored to start of line via -lbl).
+	    incr seen_header
+	    exp_continue
+	}
+
+	-re "${asname}${asdec}${asdec}${ashex}${asaccess}${trailer}" {
```
aktemur: ...and instead we begin this regexp with `\r\n`. This would make the test here more readable, considering the use of `-lbl`.

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
```diff
@@ -0,0 +1,92 @@
...
+	    # Match a data row.
+	    incr seen_address
+	    exp_continue
+	}
+
+	-re -wrap "" {
+	    # Done.
+	    gdb_assert { $seen_header == 1 && $seen_address > 0 } $gdb_test_name
```
aktemur: Nit: We can separate the asserts. In case one fails while the other passes, it will be easier to see which case failed.
```
gdb_assert { $seen_header == 1 } "$gdb_test_name: seen header"
gdb_assert { $seen_address > 0 } "$gdb_test_name: seen address"
```

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
czidev-amd: As far as I understand it, yes, to ensure gdbarch selection is clean per iteration.

gdb/amd-dbgapi-target.c
czidev-amd: Done.

gdb/doc/gdb.texinfo
czidev-amd: done

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
czidev-amd: done

gdb/testsuite/gdb.rocm/maint-print-address-spaces.exp
czidev-amd: done, we have 85 new test now.

aktemur (approved):
LGTM, thanks.

PR #219, opened 2026-07-17
gdb/amd-dbgapi-target: add trap-on-entry support

sebdar:
## Motivation

This commit introduces interface to the new 'trap-on-entry' feature. Users may now ask the debugger to halt any new wavefront executing on an AMD GPU using:

  (gdb) set amdgpu trap-on-entry on

As soon as the inferior runs a compute workload, control is returned to the debugger:
```
  Thread 15 "kernel" received signal SIGTRAP, Trace/breakpoint trap.
  [Switching to thread 15, lane 0 (AMDGPU Lane 1:2:2:9/0 (0,0,0)[0,0,0])]
  kernel () at /home/sdarche/test-dbg/src/app/long_wave.cpp:12
  12	__global__ void kernel() {
  (gdb)
```

This feature can be similarly disabled using `set amdgpu trap-on-entry off`. It is disabled by default.

## Tracking

JIRA ID : AIROCGDB-647

## Test Plan

New testcases are provided in gdb.rocm/trap-on-entry.exp. These testcases exercise the feature and verify that it disables cleanly.

## Test Result

Tests should pass with the new dbgapi. They are unsupported for version < 0.81.

## Submission Checklist

### DIFF
--- a/gdb/amd-dbgapi-target.c
+++ b/gdb/amd-dbgapi-target.c
@@ -220,11 +220,13 @@ struct amd_dbgapi_inferior_info
 {
   explicit amd_dbgapi_inferior_info
     (inferior *inf, bool precise_memory_requested = false,
-     bool precise_alu_exceptions_requested = false)
+     bool precise_alu_exceptions_requested = false,
+     bool trap_on_entry_requested = false)
     : inf (inf)
   {
     precise_memory.requested = precise_memory_requested;
     precise_alu_exceptions.requested = precise_alu_exceptions_requested;
+    trap_on_entry.requested = trap_on_entry_requested;
   }
 
   /* Backlink to inferior.  */
@@ -263,6 +265,9 @@ struct amd_dbgapi_inferior_info
   /* Track the status of precise ALU exception reporting request and
      enablement.  */
   struct dbgapi_feature_request precise_alu_exceptions;
+
+  /* Track the status of trap-on-entry request and enablement.  */
+  struct dbgapi_feature_request trap_on_entry;
 
   gdb::unordered_map<decltype (amd_dbgapi_breakpoint_id_t::handle),
 		     struct breakpoint *>
@@ -2447,6 +2452,31 @@ set_process_alu_exceptions_precision (amd_dbgapi_inferior_info &info)
 	   get_status_string (status));
 }
 
+/* Set the process' wave launch trap-on-entry mode.  */
+
+static void
+set_process_trap_on_entry (amd_dbgapi_inferior_info &info)
+{
+#if HAVE_DECL_AMD_DBGAPI_PROCESS_ENABLE_TRAPS
+  auto value = (info.trap_on_entry.requested 
+                ? AMD_DBGAPI_WAVE_ENABLE_TRAP_ON_ENTRY
+                : AMD_DBGAPI_WAVE_ENABLE_TRAP_NONE);
+
+  amd_dbgapi_status_t status
+    = amd_dbgapi_process_enable_traps (info.process_id, value, AMD_DBGAPI_WAVE_ENABLE_TRAP_ON_ENTRY);
+#else
+  amd_dbgapi_status_t status = AMD_DBGAPI_STATUS_ERROR_NOT_SUPPORTED;
+#endif
+
+  if (status == AMD_DBGAPI_STATUS_SUCCESS)
+    info.trap_on_entry.enabled = info.trap_on_entry.requested;
+  else if (status == AMD_DBGAPI_STATUS_ERROR_NOT_SUPPORTED)
+	warning (_("AMDGPU enabled traps could not be set."));
+  else
+    error (_("amd_dbgapi_process_enable_traps failed (%s)"),
+	   get_status_string (status));
+}
+
 /* Handle extra initialisation after we have attached to a AMDGPU corefile.  */
 
 static void
@@ -2570,6 +2600,7 @@ attach_amd_dbgapi (inferior *inf)
 
   set_process_memory_precision (info);
   set_process_alu_exceptions_precision (info);
+  set_process_trap_on_entry (info);
 
   /* If GDB is attaching to a process that has the runtime loaded, there will
      already be a "runtime loaded" event available.  Consume it and push the
@@ -2619,10 +2650,11 @@ detach_amd_dbgapi (inferior *inf)
   for (auto &&value : info.breakpoint_map)
     delete_breakpoint (value.second);
 
-  /* Reset the amd_dbgapi_inferior_info, except for precise_memory_mode and
-     precise_alu_exceptions.  */
+  /* Reset the amd_dbgapi_inferior_info, except for precise_memory_mode,
+     precise_alu_exceptions and trap_on_entry.  */
   info = amd_dbgapi_inferior_info (inf, info.precise_memory.requested,
-				   info.precise_alu_exceptions.requested);
+				   info.precise_alu_exceptions.requested,
+				   info.trap_on_entry.requested);
 
   maybe_reset_amd_dbgapi ();
 }
@@ -3058,6 +3090,7 @@ amd_dbgapi_target_inferior_cloned (inferior *original_inferior,
   new_info.precise_memory.requested = orig_info.precise_memory.requested;
   new_info.precise_alu_exceptions.requested
     = orig_info.precise_alu_exceptions.requested;
+  new_info.trap_on_entry.requested = orig_info.trap_on_entry.requested;
 }
 
 /* inferior_execd observer.  */
@@ -3078,6 +3111,8 @@ amd_dbgapi_inferior_execd (inferior *exec_inf, inferior *follow_inf)
   get_amd_dbgapi_inferior_info (follow_inf).precise_alu_exceptions.requested
     = get_amd_dbgapi_inferior_info (exec_inf)
 	.precise_alu_exceptions.requested;
+  get_amd_dbgapi_inferior_info (follow_inf).trap_on_entry.requested
+    = get_amd_dbgapi_inferior_info (exec_inf).trap_on_entry.requested;
 
   attach_amd_dbgapi (follow_inf);
 }
@@ -3099,6 +3134,8 @@ amd_dbgapi_inferior_forked (inferior *parent_inf, inferior *child_inf,
 	= parent_info.precise_memory.requested;
       child_info.precise_alu_exceptions.requested
 	= parent_info.precise_alu_exceptions.requested;
+      child_info.trap_on_entry.requested
+	= parent_info.trap_on_entry.requested;
 
       if (fork_kind != TARGET_WAITKIND_VFORKED)
 	{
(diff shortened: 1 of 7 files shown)


sebdar:
Corresponding dbgapi PR : https://github.com/ROCm/rocm-systems/pull/8762

gdb/amd-dbgapi-target.c
```diff
@@ -2447,6 +2452,31 @@ set_process_alu_exceptions_precision (amd_dbgapi_inferior_info &info)
...
+  amd_dbgapi_status_t status
+    = amd_dbgapi_process_set_wave_launch_trap_on_entry (info.process_id, mode);
+#else
+  amd_dbgapi_status_t status = AMD_DBGAPI_STATUS_ERROR_NOT_SUPPORTED;
+#endif
+
+  if (status == AMD_DBGAPI_STATUS_SUCCESS)
+    info.trap_on_entry.enabled = info.trap_on_entry.requested;
+  else if (status == AMD_DBGAPI_STATUS_ERROR_NOT_SUPPORTED)
+    warning (_("AMDGPU wave launch trap-on-entry could not be enabled."));
```
aktemur: Nit: "could not be enabled." -> "could not be set."

gdb/amd-dbgapi-target.c
```diff
@@ -2447,6 +2452,31 @@ set_process_alu_exceptions_precision (amd_dbgapi_inferior_info &info)
...
+  else if (status != AMD_DBGAPI_STATUS_SUCCESS)
+    error (_("amd_dbgapi_process_set_wave_launch_trap_on_entry failed (%s)"),
```
aktemur: Raising an error would cut the execution flow but not being able to set the property is only a side problem for debugging in general. We don't need to quit the attach flow and leave the state half-baked without adding the file handler. I'd suggest we emit a warning here, too, instead of an error. Also, the string would be visible to the user. So, a more user-friendly warning message like the one above may be better.

gdb/testsuite/gdb.rocm/trap-on-entry.exp
```diff
@@ -0,0 +1,101 @@
+# Copyright 2022-2026 Free Software Foundation, Inc.
```
aktemur: Are the years correct?

gdb/testsuite/gdb.rocm/trap-on-entry.exp
```diff
@@ -0,0 +1,101 @@
...
+    }
+}
+
+# Not every architecture supports trap-on-entry.  Request the feature before
+# the process starts, then let the process attach apply it and report through
+# "show" whether it actually became effective.
+set trap_on_entry_supported false
+with_test_prefix "check_trap_on_entry_support" {
+    clean_restart
+    gdb_load $::binfile
```
aktemur: Nit: `clean_restart $::testfile`

gdb/testsuite/gdb.rocm/trap-on-entry.exp
```diff
@@ -0,0 +1,101 @@
...
+
+    with_rocm_gpu_lock {
```
aktemur, line 75: Do we need this lock?

gdb/testsuite/gdb.rocm/trap-on-entry.exp
```diff
@@ -0,0 +1,101 @@
...
+
+standard_testfile .cpp
+
+if {[build_executable "failed to prepare" $testfile $srcfile {debug hip}]} {
+    return
+}
+
+proc do_test {} {
+    clean_restart
+    gdb_load $::binfile
```
aktemur: Nit: `clean_restart $::testfile`
Can we continue from the existing session, which we started below to check for support, so that we don't need to restart?

gdb/testsuite/gdb.rocm/trap-on-entry.exp
```diff
@@ -0,0 +1,101 @@
...
+	# Disable the feature while stopped in the trapped kernel.
+	gdb_test_no_output "set amdgpu trap-on-entry off" "trap-on-entry disabled at runtime"
+	gdb_test "show amdgpu trap-on-entry" \
+	    "AMDGPU wave launch trap-on-entry is off \\(currently disabled\\)\\." \
+	    "trap-on-entry shown disabled"
+
+	# The remaining dispatch must not trap; the program runs to completion.
+	gdb_test "continue" \
+	    "Inferior 1 .* exited normally.*" \
+	    "continue to end"
```
aktemur: Can we use `gdb_continue_to_end`?

lumachad:
@sebdar Is there a ticket tracking this work? If so, it would be nice to add it to the PR/commit.

gdb/testsuite/gdb.rocm/trap-on-entry.exp
sebdar: They're not :) I tend to forget to change this after copy-pasting this from another test case. Fixed.

gdb/amd-dbgapi-target.c
sebdar: This follows the pattern that set_process_memory_precision / set_process_alu_exceptions_precision above use. Getting something other than NOT_SUPPORTED means that something wrong happened in dbgapi (or lower in the stack, driver/device). So I think this is appropriate?

gdb/testsuite/gdb.rocm/trap-on-entry.exp
sebdar: Fixed (i'll admit this was copy-pasted from a similar test, we might want to update those as well ?)

gdb/testsuite/gdb.rocm/trap-on-entry.exp
sebdar: On the driver side this is supposed to be per-VMID so it shouldn't (?) affect other processes. But I'm not sure if the API should guarantee this.

gdb/testsuite/gdb.rocm/trap-on-entry.exp
sebdar: The feature would be enabled. Similarly, this is the approach in `alu-exceptions.exp`

gdb/testsuite/gdb.rocm/trap-on-entry.exp
sebdar: Yes. Fixed.

gdb/amd-dbgapi-target.c
sebdar: I'll agree this sounds more correct so I changed it. But again this is pretty similar to the two above features. Leaving unresolved for now.

sebdar:
Thanks for the review @aktemur .

In the last push : 
- Addressed some of Baris' comments. Left a few unresolved for further discussion

gdb/testsuite/gdb.rocm/trap-on-entry.exp
```diff
@@ -0,0 +1,99 @@
...
+
+    with_rocm_gpu_lock {
+	gdb_test_no_output "set amdgpu trap-on-entry on"
+	if {![runto_main]} {
+	    return
+	}
+
+	gdb_test_multiple "show amdgpu trap-on-entry" "" {
+	    -re -wrap "AMDGPU wave launch trap-on-entry is on \\(currently (enabled|disabled)\\).*" {
+		if {$expect_out(1,string) == "enabled"} {
```
aktemur: Let's please use `eq` here as the operator.

gdb/testsuite/gdb.rocm/trap-on-entry.exp
aktemur: Can you check if you pushed the change? I don't see it.

aktemur (approved):

gdb/testsuite/gdb.rocm/trap-on-entry.exp
sebdar: Done.

gdb/testsuite/gdb.rocm/trap-on-entry.exp
lancesix: I think we need it.  Older architectures only support debugging one process at a time.  When we do `runto_main`, we start debugging the process from KFD's perspective.  We need to make sure this is not conflicting with other testcases running in parllel.

gdb/testsuite/gdb.rocm/trap-on-entry.exp
```diff
@@ -0,0 +1,97 @@
...
+
+	# The remaining dispatch must not trap; the program runs to completion.
+	gdb_continue_to_end "" continue 1
+    }
+}
+
+# Not every architecture supports trap-on-entry.  Request the feature before
+# the process starts, then let the process attach apply it and report through
+# "show" whether it actually became effective.
+set trap_on_entry_supported false
```
lancesix, line 71: nit: I think I would make `supports_trap_on_entry` a function which can be called, rather than inlining it.  But this is mostly a me thing.

gdb/amd-dbgapi-target.c
lancesix: Does dbgapi give a guarantee regarding its state if it returns an error? Usually STATUS_ERROR are said to leave dbgapi in an unknown state, which might not be usable to continue an attaching sequence.

gdb/amd-dbgapi-target.c
```diff
@@ -2447,6 +2452,31 @@ set_process_alu_exceptions_precision (amd_dbgapi_inferior_info &info)
 	   get_status_string (status));
 }
 
+/* Set the process' wave launch trap-on-entry mode.  */
+
+static void
+set_process_trap_on_entry (amd_dbgapi_inferior_info &info)
+{
+#if AMD_DBGAPI_VERSION_MAJOR > 0 || AMD_DBGAPI_VERSION_MINOR >= 81
```
lancesix: Ideally, we might want to have a configure test checking if `AMD_DBGAPI_WAVE_LAUNCH_TRAP_ON_ENTRY_ENABLED` is defined for example, and set a `AMD_DBGAPI_SUPPORTS_TRAP_ON_ENTRY`.  This would remove some ordering issues, as we have quite a few dbgapi things in the pipeline comming, ordering might not guarantee that this feature will be 0.81.

lancesix (changes requested):
Documentation is missing.

Other bits and bobs comments below.

sebdar:
In the last push : 
- Added documentation
- Reworked integration with dbgapi to check for the symbol at configure-time instead of checking for a dbgapi version.
- Misc modifications:
  - copyright year in the testcase
  - forward state to cloned inferiors, just like precise-memory and alu-exceptions

gdb/doc/gdb.texinfo
```diff
@@ -29895,6 +30074,79 @@ Display the currently requested and effective @acronym{AMD}
...
+
+@item off
+The program executes normally.  This is the default.
+
+@item on
+Request that the program breaks at every kernel entry on the devices.
+
+The trap-on-entry mode can be changed at any time in the life of the
+program, but remains effective for all waves that were created until it
+was disabled.  If the trap-on-entry mode is updated before the inferior
```
lancesix: "but remains effective for all waves that were created until it was disabled" sounds odd here.  By definition, this only affects we have never seen.  Once the setting is disabled and execution is resumed, we should not see any more wave on entry trap.

gdb/testsuite/gdb.rocm/trap-on-entry.cpp
```diff
@@ -0,0 +1,63 @@
...
+  *out = a + b;
+}
+
+static void
+run_addition (int *result_ptr)
+{
+  do_an_addition<<<dim3(1), dim3(1), 0, 0>>> (1, 2, result_ptr);
+
+  hipError_t error = hipDeviceSynchronize ();
+  assert (error == hipSuccess);
```
lancesix: might be worth using `CHECK`, which would print the hip error code in case of error.

lancesix (commented):
some minor comments, otherwise LGTM.

gdb/doc/gdb.texinfo
sebdar: I agree the wording feels a bit awkward.  What I meant was that this would affect all waves in the same dispatch, so if the user sees a first wave and decides they want to disable the feature, they'll have to `continue` through all the remaining waves that were created at the same time as the first one.

gdb/testsuite/gdb.rocm/trap-on-entry.cpp
sebdar: Done.

gdb/doc/gdb.texinfo
sebdar: Removed for now.

sebdar:
In the last push : 
- Used `CHECK` in the test case
- Removed awkward sentence in doc.

gdb/doc/gdb.texinfo
lancesix: so what you meant as that one needs to drain the list of already created but still pending events.

lumachad:
New PR is https://github.com/ROCm/rocm-systems/pull/9925. Pre-commit seems unhappy about whitespaces.

PR #225, opened 2026-07-21
Implements the "maintenance info amdgpu address-aliases" command

czidev-amd:
## Motivation

Given an expression that evaluates to an address in some address space, show for each address space supported by the current AMDGPU architecture the aliasing address in that space, if any, along with the number of consecutive bytes for which the aliasing is valid.

This is purely a diagnostic aid: it queries amd_dbgapi_convert_address_space for every address space known to the architecture and reports "<not reachable>" for the ones the given address does not alias into.

## Test Plan

New `show_address_aliases.exp` test added.

### DIFF
--- a/gdb/amdgpu-tdep.c
+++ b/gdb/amdgpu-tdep.c
@@ -35,6 +35,8 @@
 #include "reggroups.h"
 #include "extract-store-integer.h"
 #include <optional>
+#include "cli/cli-cmds.h"
+#include "cli/cli-style.h"
 
 struct amdgpu_per_inferior
 {
@@ -93,6 +95,9 @@ amdgpu_get_segment_address_significant_bits (inferior *inf)
 
   return *per_inf.significant_bits;
 }
+
+/* List of maint info amdgpu commands.  */
+static struct cmd_list_element *maintenance_info_amdgpu_list;
 
 /* Return true if INFO is of an AMDGPU architecture.  */
 
@@ -1930,6 +1935,143 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
   return aliases;
 }
 
+/* Show for each address space supported by the current AMDGPU
+   architecture the aliases of a given ADDR, and the number of
+   consecutive bytes for which each aliasing is valid.
+
+   This is purely a diagnostic aid: it queries
+   amd_dbgapi_convert_address_space for every address space known to
+   the architecture and reports "<not reachable>" for the ones the
+   given address does not alias into.  */
+
+static void
+amdgpu_info_address_aliases_command (const char *exp, int from_tty)
+{
+  /* A series of sanity checks.  */
+  if (exp == nullptr || *exp == '\0')
+    error (_("Argument required."));
+
+  expression_up expr = parse_expression (exp);
+  struct gdbarch *gdbarch = expr->gdbarch;
+  if (!is_amdgpu_arch (gdbarch))
+    error (_("Current architecture is not AMDGPU."));
+
+  if (inferior_ptid == null_ptid)
+    error (_("No thread selected."));
+
+  if (!ptid_is_gpu (inferior_ptid))
+    error (_("Current thread is not an AMDGPU GPU thread."));
+
+  struct value *val = expr->evaluate ();
+  if (TYPE_IS_REFERENCE (val->type ()))
+    val = coerce_ref (val);
+  CORE_ADDR addr = value_as_address (val);
+
+  amd_dbgapi_architecture_id_t architecture_id;
+  amd_dbgapi_status_t status = amd_dbgapi_get_architecture
+    (gdbarch_bfd_arch_info (gdbarch)->mach, &architecture_id);
+  if (status != AMD_DBGAPI_STATUS_SUCCESS)
+    error (_("amd_dbgapi_get_architecture failed (%s)"),
+	   get_status_string (status));
+
+  arch_addr_space_id src_aspace_num
+    = amdgpu_address_space_id_from_core_address (addr);
+
+  amd_dbgapi_address_space_id_t src_aspace;
+  status = amd_dbgapi_dwarf_address_space_to_address_space (architecture_id,
+							    src_aspace_num,
+							    &src_aspace);
+  if (status != AMD_DBGAPI_STATUS_SUCCESS)
+    error (_("amd_dbgapi_dwarf_address_space_to_address_space failed (%s)"),
+	   get_status_string (status));
+
+  CORE_ADDR offset = amdgpu_segment_address_from_core_address (addr);
+
+  /* amd_dbgapi_convert_address_space special-cases the NULL segment
+     address of the source space: it succeeds and returns the NULL
+     segment address of the destination space, without this being a
+     genuine alias.  Detect that case up front so it isn't reported
+     as a real hit below.  */
+  amd_dbgapi_segment_address_t src_null_address;
+  status = amd_dbgapi_address_space_get_info
+    (src_aspace, AMD_DBGAPI_ADDRESS_SPACE_INFO_NULL_ADDRESS,
+     sizeof (src_null_address), &src_null_address);
+  if (status != AMD_DBGAPI_STATUS_SUCCESS)
+    error (_("amd_dbgapi_address_space_get_info (NULL_ADDRESS) failed (%s)"),
+	   get_status_string (status));
+  bool src_is_null_address = (offset == src_null_address);
+
+  const amd_dbgapi_wave_id_t wave_id
+    = get_amd_dbgapi_wave_id (inferior_thread ()->ptid);
+  const amd_dbgapi_lane_id_t lane_id
+    = inferior_thread ()->current_simd_lane ();
+  const gdb::array_view<const arch_addr_space> aspaces
+    = amdgpu_address_spaces (gdbarch);
+
+  /* First pass: size the "Address space" column to the longest name.
+     Initialize with the header title width so the column is never
+     narrower than the label.  */
+  size_t max_aspace_name_len = strlen (_("Address space"));
+  for (const auto &address_space : aspaces)
+    max_aspace_name_len = std::max (max_aspace_name_len,
+				    strlen (address_space.name.get ()));
+
+  int addr_width = 4 + (gdbarch_ptr_bit (gdbarch) / 4);
+
+  struct ui_out *ui_out = current_uiout;
+  ui_out_emit_table table_emitter (ui_out, 3, aspaces.size (), "aliases");
+
+  ui_out->table_header (max_aspace_name_len, ui_left, "aspace",
+			_("Address space"));
+  /* Address form is: <name>#0xNN..NNN for a named space,
+     or just 0xNN..NNN for the default (global) space.  */
+  ui_out->table_header (max_aspace_name_len + addr_width,
+			ui_right, "addr", _("Address"));
+  ui_out->table_header (addr_width, ui_right, "valid", _("Valid for"));
+  ui_out->table_body ();
+
+  for (const auto &address_space : aspaces)
+    {
+      amd_dbgapi_address_space_id_t dst_aspace;
+      status = amd_dbgapi_dwarf_address_space_to_address_space
+	(architecture_id, address_space.id, &dst_aspace);
+      if (status != AMD_DBGAPI_STATUS_SUCCESS)
+	error (_("amd_dbgapi_dwarf_address_space_to_address_space failed (%s)"),
+	       get_status_string (status));
+
+      ui_out_emit_tuple tuple_emitter (ui_out, "alias");
+
+      amd_dbgapi_segment_address_t dst_offset;
+      amd_dbgapi_size_t cont_bytes;
+      status = amd_dbgapi_convert_address_space (wave_id, lane_id, src_aspace,
+						 offset, dst_aspace,
+						 &dst_offset, &cont_bytes);
+      ui_out->field_string ("aspace", address_space.name.get ());
+      if (status != AMD_DBGAPI_STATUS_SUCCESS)
+	{
+	  ui_out->field_string ("addr", "<not reachable>");
+	  ui_out->field_string ("valid", "0x0");
+	}
+      else
+	{
+	  CORE_ADDR dst_address
+	    = amdgpu_segment_address_to_core_address (address_space.id,
+						      dst_offset);
+	  ui_out->field_string ("addr", paspace_and_addr (gdbarch,
+							  dst_address),
+				address_style.style ());
+
+	  /* Not a real alias, NULL just maps to NULL in every
+	     space.  */
+	  if (src_is_null_address)
+	    ui_out->field_string ("valid", "<null>");
+	  else
+	    ui_out->field_fmt ("valid", "0x%s", phex_nz (cont_bytes));
+	}
+      ui_out->text ("\n");
+    }
+}
+
 static simd_lanes_mask_t
 amdgpu_active_lanes_mask (struct gdbarch *gdbarch, thread_info *tp)
 {
@@ -2434,6 +2576,18 @@ INIT_GDB_FILE (amdgpu_tdep)
 		    amdgpu_supports_arch_info);
   gdb::observers::inferior_appeared.attach (amdgpu_observer_inferior_appeared,
 					    "amdgpu_tdep");
+
+  add_basic_prefix_cmd ("amdgpu", class_maintenance, _("\
+Commands for showing internal info about the AMDGPU target."),
+			&maintenance_info_amdgpu_list, 0,
+			&maintenanceinfolist);
+
+  add_cmd ("address-aliases", class_maintenance,
+	   amdgpu_info_address_aliases_command, _("\
+Show ADDRESS aliases across AMDGPU address spaces.\n\
+Usage: maintenance info amdgpu address-aliases ADDRESS.\n"),
+	   &maintenance_info_amdgpu_list);
+
 #if defined GDB_SELF_TEST
   selftests::register_test ("amdgpu-register-type-parse-flags-fields",
 			    amdgpu_register_type_parse_test);
--- a/gdb/doc/gdb.texinfo
+++ b/gdb/doc/gdb.texinfo
@@ -45345,6 +45345,31 @@ packet history.
 @item maint info jit
 Print information about JIT code objects loaded in the current inferior.
 
+@kindex maint info amdgpu address-aliases
+@item maint info amdgpu address-aliases @var{expression}
+Evaluate @var{expression} to an address, and show the address it
+aliases to in every address space known to the current @acronym{AMD}
+@acronym{GPU} architecture.  The @samp{Valid for} column shows the
+number of consecutive bytes for which the aliasing is valid.
+
+For example:
+
+@smallexample
+(gdb) maintenance info amdgpu address-aliases &idx
+Address space                      Address         Valid for
+global                      0x7ff7f4600c00               0x4
+generic            generic#0x2000000000030             0x3d0
+private_wave            private_wave#0xc00               0x4
+private_lane             private_lane#0x30             0x3d0
+local                      <not reachable>               0x0
+@end smallexample
+
+This only works while stopped in an @acronym{AMD} @acronym{GPU} thread.
+Some address spaces are private to a wave or a lane of a wave, and
+@value{GDBN} converts using whichever wave and lane are currently
+selected.  Not every address is reachable from every space: if it isn't,
+the row for that space reads @samp{<not reachable>}.
+
 @anchor{maint info python-disassemblers}
 @kindex maint info python-disassemblers
 @item maint info python-disassemblers
--- a/gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.cpp
+++ b/gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.cpp
@@ -1,0 +1,60 @@
+/* Copyright (C) 2026 Free Software Foundation, Inc.
+   Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+   This file is part of GDB.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.  */
+
+#include "hip/hip_runtime.h"
+#include <vector>
+
+#include "rocm-test-utils.h"
+
+__global__ void __attribute__ ((optnone))
+kern (int *out)
+{
+  int idx = threadIdx.x;
+  __shared__ int gidx;
+  __shared__ int gdimx;
+
+  if (idx == 0)
+    {
+      int _gidx = blockIdx.x;
+      int _gdimx = blockDim.x;
+      gidx = _gidx;
+      gdimx = _gdimx;
+      volatile int *gidx_p = &gidx;
+      volatile int *gdimx_p = &gdimx;
+      volatile auto unused = gidx + gdimx; /* Break here.  */
+    }
+  __syncthreads ();
+
+  out[gidx * gdimx + idx] = threadIdx.x;
+}
+
+int
+main ()
+{
+  int *devOut;
+
+  CHECK (hipMalloc (&devOut, 64 * sizeof (int)));
+
+  kern<<<1, 1>>> (devOut);
+
+  std::vector<int> out (64);
+  CHECK (hipMemcpy (out.data (), devOut, 64 * sizeof (int),
+		    hipMemcpyDeviceToHost));
+
+  return 0;
+}
--- a/gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.exp
+++ b/gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.exp
@@ -1,0 +1,103 @@
+# Copyright (C) 2026 Free Software Foundation, Inc.
+# Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+# This file is part of GDB.
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test the maint info amdgpu address-aliases command.
+
+load_lib rocm.exp
+
+standard_testfile .cpp
+
+require allow_hip_tests
+
+if {[prepare_for_testing "failed to prepare ${testfile}" $testfile $srcfile {debug hip}]} {
+    return -1
+}
+
+with_rocm_gpu_lock {
+
+    # Test: command before any inferior is running (no thread selected).
+    gdb_test "maint info amdgpu address-aliases &main" \
+	"Current architecture is not AMDGPU\\." \
+	"address-aliases with no thread selected"
+
+    if {![runto_main]} {
+	return -1
+    }
+
+    # Test: command when focused on a host (CPU) thread, right after
+    # reaching main.
+    gdb_test "maint info amdgpu address-aliases &main" \
+	"Current architecture is not AMDGPU\\." \
+	"address-aliases on host thread"
+
+    gdb_breakpoint [gdb_get_line_number "Break here"] -allow-pending
+    gdb_continue_to_breakpoint "kern"
+
+    set asname "\[^ \t\r\n\]+"
+    set trailer "\[^\r\n\]*(?=\r\n)"
+
+    # Count address spaces known to the architecture.
+    set aspaces 0
+    gdb_test_multiple "maint print address-spaces" "" -lbl {
+	-re "\r\n${asname}\[ \t\]+$::decimal\[ \t\]+" {
+	    incr aspaces
+	    exp_continue
+	}
+	-re -wrap "" {
+	    # Done collecting.
+	}
+    }
+
+    set seen_header 0
+    set seen_address 0
+
+    # The actual output can vary, we want to verify if GDB doesn't
+    # have any latent error.
+    gdb_test_multiple "maint info amdgpu address-aliases &idx" "" -lbl {
+	-re "Address space\[ \t\]+Address\[ \t\]+Valid for${trailer}" {
+	    # Match the header.
+	    incr seen_header
+	    exp_continue
+	}
+
+	-re "^\r\n${asname}\[ \t\]+$::hex\[ \t\]+$::hex${trailer}" {
+	    # Match global address space.
+	    incr seen_address
+	    exp_continue
+	}
+
+	-re "^\r\n${asname}\[ \t\]+${asname}#$::hex\[ \t\]+($::hex|<null>)${trailer}" {
+	    # Match address space.
+	    incr seen_address
+	    exp_continue
+	}
+
+	-re "^\r\n${asname}\[ \t\]+<not reachable>\[ \t\]+($::hex|<null>)${trailer}" {
+	    # Match <not reachable> addresses.
+	    incr seen_address
+	    exp_continue
+	}
+
+	-re -wrap "" {
+	    # Done.  Every address space should have exactly one row.
+	    gdb_assert { $seen_header == 1 } "$gdb_test_name: seen header"
+	    gdb_assert { $seen_address == $aspaces } \
+		"$gdb_test_name: seen address space"
+	}
+    }
+}
--- a/CHANGELOG_AMD.md
+++ b/CHANGELOG_AMD.md
@@ -7,6 +7,11 @@ Full documentation for ROCgdb is available at
 
 ### Added
 
+- The "maint info amdgpu address-aliases" command evaluate a given
+  expression to an address, and show what the address aliases to in
+  every address space known to the current AMD GPU selected
+  architecture.
+
 - Improve "maint print address-spaces" command to display properties
   of address spaces supported by an architecture.  For each address
   space, print its name, DWARF id, address size, null address, and


gdb/amdgpu-tdep.c
```diff
@@ -94,6 +96,9 @@ amdgpu_get_segment_address_significant_bits (inferior *inf)
   return *per_inf.significant_bits;
 }
 
+/* List of maint info amdgpu commands.  */
+struct cmd_list_element *maintenance_info_amdgpu_list;
+
```
lumachad, line 101: This variable is never referenced from another translation unit and should be declared `static`.

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,140 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+  ui_out->table_header (max_aspace_name_len, ui_left, "aspace",
+			_("Address space"));
+  /* Address form is: <Name>0xNN..NNN.  */
+  ui_out->table_header (addr_width + max_aspace_name_len,
+			ui_right, "addr", _("Address"));
+  ui_out->table_header (addr_width, ui_right, "valid", _("Valid for"));
+  ui_out->table_body ();
+
+  for (const auto &address_space : amdgpu_address_spaces (gdbarch))
+    {
```
lumachad, line 2032: `aspaces` was already computed above for the first loop. Can we reuse it?

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,140 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+      amd_dbgapi_status_t err = amd_dbgapi_convert_address_space
+	(wave_id, lane_id, src_aspace, offset,
+	 dst_aspace, &dst_offset, &cont_bytes);
+
+      ui_out->field_string ("aspace", address_space.name.get ());
+      if (err != AMD_DBGAPI_STATUS_SUCCESS)
+	{
+	  ui_out->field_string ("addr", "<not reachable>");
+	  ui_out->field_fmt ("valid", "0x0");
+	}
```
lumachad, line 2052: `field_fmt` with no format arguments should be `field_string ("valid", "0x0")` instead.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <vector>
+
+#define CHECK(cmd)                                                           \
+    {                                                                        \
+	hipError_t error = cmd;                                              \
+	if (error != hipSuccess)                                             \
+	  {                                                                  \
+	    fprintf(stderr, "error: '%s'(%d) at %s:%d\n",                    \
+		    hipGetErrorString(error), error,                         \
+		    __FILE__, __LINE__);                                     \
```
lumachad: Formatting: Missing space before `(` in `fprintf(` and `hipGetErrorString(`.

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,140 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+   the number of consecutive bytes for which the aliasing is valid.
+
+   This is purely a diagnostic aid: it queries
+   amd_dbgapi_convert_address_space for every address space known to
+   the architecture and reports "<not reachable>" for the ones the
+   given address does not alias into.  */
+
+static void
+amdgpu_info_addressaliases_command (const char *exp, int from_tty)
+{
```
lumachad, line 1948: Typo: Function name `amdgpu_info_addressaliases_command` runs two words together without a separator.

gdb/doc/gdb.texinfo
```diff
@@ -45198,6 +45198,19 @@ packet history.
 @item maint info jit
 Print information about JIT code objects loaded in the current inferior.
 
+@kindex maint info amdgpu address-aliases
+@item maint info amdgpu address-aliases @var{expression}
+Evaluate @var{expression} to an address, and show the address it
+aliases to in every other address space known to the current
+@acronym{AMD} @acronym{GPU} architecture, along with how many
```
lumachad: "every other address space" is inaccurate. The command shows all address spaces including the source space itself. Drop "other".

gdb/testsuite/gdb.rocm/show_address_aliases.exp
```diff
@@ -0,0 +1,61 @@
...
+	     ".*Address space\[ \t\]*Address\[ \t\]*Valid for\[ \t\]*" \
+	     "global\[ \t\]*$::hex\[ \t\]*$::hex\[ \t\]*" \
+	     "generic\[ \t\]*generic\#$::hex\[ \t\]*$::hex\[ \t\]*" \
+	     "private_wave\[ \t\]*private_wave\#$::hex\[ \t\]*$::hex\[ \t\]*" \
+	     "private_lane\[ \t\]*private_lane\#$::hex\[ \t\]*$::hex\[ \t\]*" \
+	     "local\[ \t\]*<not reachable>\[ \t\]*0x0\[ \t\]*" ]
+
+    # Show aliases of a LDS address
+    gdb_test "maint info amdgpu address-aliases gidx_p" \
+	[multi_line \
```
lumachad: The comment says "LDS address" but `gidx_p` is a pointer variable whose value is an LDS address, not an LDS address directly. A clearer comment would be: `# Show aliases using a pointer to a shared variable (LDS address)`.

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,140 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+  if (amd_dbgapi_address_space_get_info
+      (src_aspace, AMD_DBGAPI_ADDRESS_SPACE_INFO_NULL_ADDRESS,
+       sizeof (src_null_address), &src_null_address)
+      != AMD_DBGAPI_STATUS_SUCCESS)
+    error (_("amd_dbgapi_address_space_get_info (NULL_ADDRESS) failed."));
+  bool src_is_null_address = (offset == src_null_address);
+
+  const amd_dbgapi_wave_id_t wave_id
+    = get_amd_dbgapi_wave_id (inferior_thread ()->ptid);
+  const amd_dbgapi_lane_id_t lane_id
```
lumachad, line 2005: `inferior_thread()` asserts `current_thread_ != nullptr` and will abort GDB if no thread is selected. Do we risk crashing GDB if we call this without inferior threads?

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,140 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+							  dst_address),
+				address_style.style ());
+
+	  /* Not a real alias, NULL just maps to NULL in every
+	     space.  */
+	  if (src_is_null_address)
+	    ui_out->field_string ("valid", "<null>");
+	  else
+	    ui_out->field_fmt ("valid", "0x%lx", cont_bytes);
+	}
```
lumachad, line 2068: `%lx` assumes `amd_dbgapi_size_t` is `unsigned long`, which is not guaranteed.

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,140 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+   amd_dbgapi_convert_address_space for every address space known to
+   the architecture and reports "<not reachable>" for the ones the
+   given address does not alias into.  */
+
+static void
+amdgpu_info_addressaliases_command (const char *exp, int from_tty)
+{
+  if (exp == nullptr || *exp == '\0')
+    error ("Argument required.");
+
```
lumachad, line 1952: Formatting: Missing `_()` wrapper for i18n.

gdb/testsuite/gdb.rocm/show_address_aliases.exp
```diff
@@ -0,0 +1,61 @@
...
+	     ".*Address space\[ \t\]*Address\[ \t\]*Valid for\[ \t\]*" \
+	     "global\[ \t\]*0x0\[ \t\]*<null>\[ \t\]*" \
+	     "generic\[ \t\]*generic\#0x0\[ \t\]*<null>\[ \t\]*" \
+	     "private_wave\[ \t\]*private_wave\#$::hex\[ \t\]*<null>\[ \t\]*" \
+	     "private_lane\[ \t\]*private_lane\#$::hex\[ \t\]*<null>\[ \t\]*" \
+	     "local\[ \t\]*local\#$::hex\[ \t\]*<null>\[ \t\]*" ]
```
lumachad: Do we know if this is expected to work across all supported gfx architectures? Will all of them show all of these? What if they show more or less?

gdb/amdgpu-tdep.c
czidev-amd: In the beginning of this function, I've added two checks, one to check if there is any active debugging thread, and the second checks if the current debugging thread is a GPU thread or not.

czidev-amd (commented):
Thank you @lumachad for your review, I have reworked a bit this PR to include your comments. One observation is the renaming of the tests to match other "maint" tests naming pattern. Also, I have relaxed the test as I cannot really guarantee the same output across different architectures. Now, the tests is testing if gdb implementation is sane rather than gdbabi output sanity.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.cpp
```diff
@@ -0,0 +1,68 @@
...
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.  */
+
+#include "hip/hip_runtime.h"
+#include <vector>
+
+#define CHECK(cmd)                                                           \
```
palves: Include rocm-test-utils.h instead of inlining CHECK here.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.cpp
```diff
@@ -0,0 +1,68 @@
...
+int
+main ()
+{
+  int *devOut;
+
+  CHECK (hipMalloc (&devOut, 64 * sizeof (int)));
+
+  kern<<<2, 32>>> (devOut);
+
+  std::vector<int> out(64);
```
palves: Space before parens.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
```diff
@@ -0,0 +1,64 @@
...
+require allow_hip_tests
+
+if {[prepare_for_testing "failed to prepare ${testfile}" $testfile $srcfile {debug hip}]} {
+    return -1
+}
+
+with_rocm_gpu_lock {
+
+    if {![runto_main]} {
+	fail "can't run to main"
```
palves: runto_main already emits a FAIL.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
```diff
@@ -0,0 +1,64 @@
...
+    gdb_continue_to_breakpoint "kern"
+
+    set seen_header 0
+    set seen_address 0
+
+    # The actual output can vary, we want to verify if GDB doesn't
+    # have any latent error.
+    gdb_test_multiple "maint info amdgpu address-aliases &idx" "" -lbl {
+	-re ".*Address space\[ \t\]*Address\[ \t\]*Valid for\[^\r\n\]*(?=\r\n)" {
+	    # Match the header
```
palves: Missing period.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
```diff
@@ -0,0 +1,64 @@
...
+    gdb_breakpoint [gdb_get_line_number "Break here"] -allow-pending
+    gdb_continue_to_breakpoint "kern"
+
+    set seen_header 0
+    set seen_address 0
+
+    # The actual output can vary, we want to verify if GDB doesn't
+    # have any latent error.
+    gdb_test_multiple "maint info amdgpu address-aliases &idx" "" -lbl {
+	-re ".*Address space\[ \t\]*Address\[ \t\]*Valid for\[^\r\n\]*(?=\r\n)" {
```
palves: Drop leading ".*", it's implicit / unnecessary.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
```diff
@@ -0,0 +1,64 @@
...
+	    # Match the header
+	    incr seen_header
+	    exp_continue
+	}
+
+	-re "^\r\n\[a-z_\]+\[ \t\]+\[a-z_<>\]*\#*$::hex\[ \t\]*($::hex|<null>)\[^\r\n\]*(?=\r\n)" {
+	    # Match address space
```
palves: Period.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
```diff
@@ -0,0 +1,64 @@
...
+	    incr seen_address
+	    exp_continue
+	}
+
+	-re -wrap "" {
+	    # Done
```
palves: Period.

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,148 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+amdgpu_info_address_aliases_command (const char *exp, int from_tty)
+{
+  /* A series of sanity checks.  */
+  if (exp == nullptr || *exp == '\0')
+    error (_("Argument required."));
+
+  expression_up expr = parse_expression (exp);
+  struct gdbarch *gdbarch = expr->gdbarch;
+  if (!is_amdgpu_arch (gdbarch))
+    error (_("Expression does not evaluate in the context of an AMDGPU device."));
```
palves: I find this sentence a bit ambiguous/confusing, as it could also be interpreted as:

  "Expression cannot be evaluated in the context of an AMDGPU device."

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,148 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+  struct value *val = expr->evaluate ();
+  if (TYPE_IS_REFERENCE (val->type ()))
+    val = coerce_ref (val);
+  CORE_ADDR addr = value_as_address (val);
+
+  amd_dbgapi_architecture_id_t architecture_id;
+  if (amd_dbgapi_get_architecture
+      (gdbarch_bfd_arch_info (gdbarch)->mach, &architecture_id)
+      != AMD_DBGAPI_STATUS_SUCCESS)
+    error (_("amd_dbgapi_get_architecture failed."));
```
palves: Please include the failing status in the error message.  The typical pattern is:

```
  if (status != AMD_DBGAPI_STATUS_SUCCESS)
    error (_("amd_dbgapi_get_architecture failed (%s)"), get_status_string (status));
```

This applies to other places in the patch as well.

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,148 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+
+  const amd_dbgapi_wave_id_t wave_id
+    = get_amd_dbgapi_wave_id (inferior_thread ()->ptid);
+  const amd_dbgapi_lane_id_t lane_id
+    = inferior_thread ()->current_simd_lane ();
+  const gdb::array_view<const arch_addr_space> aspaces
+    = amdgpu_address_spaces (gdbarch);
+
+  /* First pass: size the "Address space" column to the longest name.
+     Initialise with the header title width so the column is never
```
palves: GDB uses US spelling.  Initialise => Initialize

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,148 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+     narrower than the label.  */
+  size_t max_aspace_name_len = strlen ("Address space");
```
palves: Below, when the table is constructed, you run the string via gettext, so here you should too, otherwise if the column header is translate, your max length will be wrong.

gdb/amdgpu-tdep.c
```diff
@@ -2430,6 +2577,17 @@ INIT_GDB_FILE (amdgpu_tdep)
...
 					    "amdgpu_tdep");
+
+  add_basic_prefix_cmd ("amdgpu", class_maintenance, _("\
+Commands for showing internal info about the AMDGPU target."),
+			    &maintenance_info_amdgpu_list, 0,
+			    &maintenanceinfolist);
+
+  add_cmd ("address-aliases", class_maintenance,
+	   amdgpu_info_address_aliases_command,
+	   _("Show alias of address"),
```
palves: This could be expanded a bit more.  Also, there can be more than one alias, right?  So alias => aliases.

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,148 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+							  dst_address),
+				address_style.style ());
+
+	  /* Not a real alias, NULL just maps to NULL in every
+	     space.  */
+	  if (src_is_null_address)
+	    ui_out->field_string ("valid", "<null>");
+	  else
+	    ui_out->field_fmt ("valid", "0x%s",
+			       phex_nz ((uint64_t) cont_bytes));
```
palves: Do we need the cast?

Also, seems like it'd fit in one line.

gdb/doc/gdb.texinfo
```diff
@@ -45246,6 +45246,19 @@ packet history.
 @item maint info jit
 Print information about JIT code objects loaded in the current inferior.
 
+@kindex maint info amdgpu address-aliases
+@item maint info amdgpu address-aliases @var{expression}
+Evaluate @var{expression} to an address, and show the address it
+aliases to in every address space known to the current @acronym{AMD}
+@acronym{GPU} architecture, along with how many consecutive bytes each
+alias stays valid for.
```
palves: The "valid for" bit is a bit cryptic.  I think this could benefit from an example.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+  if (idx == 0)
+    {
+      int _gidx = blockIdx.x;
+      int _gdimx = blockDim.x;
+      gidx = _gidx;
+      gdimx = _gdimx;
+      int * gidx_p = &gidx;
+      int * gdimx_p = &gdimx;
+      auto unused = gidx + gdimx; /* Break here.  */
```
lancesix: some support landed in to try to support testing when building with optimisations.  I kind of expect this needs to be volatile so the compiler can't just optimise everything out.

gdb/doc/gdb.texinfo
czidev-amd: Done, thank you

czidev-amd:
I have spined a new version including Pedro's and Lancelot's inputs.

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,150 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+      if (status != AMD_DBGAPI_STATUS_SUCCESS)
+	error (_("amd_dbgapi_dwarf_address_space_to_address_space failed (%s)"),
+	       get_status_string (status));
+
+      ui_out_emit_tuple tuple_emitter (ui_out, "alias");
+
+      amd_dbgapi_segment_address_t dst_offset;
+      amd_dbgapi_size_t cont_bytes;
+      amd_dbgapi_status_t err = amd_dbgapi_convert_address_space
+	(wave_id, lane_id, src_aspace, offset,
```
lumachad: This variable is named `err` while every other `amd_dbgapi_status_t` variable in this function uses `status`. Please use `status` consistently (the outer `status` is no longer live at this point in the loop, so it can safely be reused).

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
```diff
@@ -0,0 +1,63 @@
...
+    gdb_test_multiple "maint info amdgpu address-aliases &idx" "" -lbl {
+	-re "Address space\[ \t\]*Address\[ \t\]*Valid for\[^\r\n\]*(?=\r\n)" {
+	    # Match the header.
+	    incr seen_header
+	    exp_continue
+	}
+
+	-re "^\r\n\[a-z_\]+\[ \t\]+\[a-z_<>\]*\#*$::hex\[ \t\]*($::hex|<null>)\[^\r\n\]*(?=\r\n)" {
+	    # Match address space.
+	    incr seen_address
```
lumachad, line 54: The data-row pattern requires a hex address (via `$::hex`) in the address column, so rows showing `<not reachable>` do not match it. In `-lbl` mode, a non-matching line falls through to `-re -wrap ""`, which consumes all remaining output and jumps immediately to the `done` arm. If the first address-space row in the output is unreachable, `seen_address` stays `0` and the assert fails. Please add a third pattern for the unreachable case:

```tcl
-re "^\r\n\[a-z_\]+\[ \t\]+<not reachable>\[ \t\]*0x0\[^\r\n\]*(?=\r\n)" {
    exp_continue
}
```

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
```diff
@@ -0,0 +1,63 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test the maint info amdgpu address-aliases
+
+load_lib rocm.exp
```
lumachad, line 21: Missing a period at the end of the comment. Other `.exp` files in this directory end their opening description comment with a period.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.cpp
```diff
@@ -0,0 +1,58 @@
...
+  __shared__ int gdimx;
+
+  if (idx == 0)
+    {
+      int _gidx = blockIdx.x;
+      int _gdimx = blockDim.x;
+      gidx = _gidx;
+      gdimx = _gdimx;
+      volatile int * gidx_p = &gidx;
+      volatile int * gdimx_p = &gdimx;
```
aktemur: Nit: No space after `*`.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.cpp
```diff
@@ -0,0 +1,58 @@
...
+  int *devOut;
+
+  CHECK (hipMalloc (&devOut, 64 * sizeof (int)));
+
+  kern<<<2, 32>>> (devOut);
+
+  std::vector<int> out (64);
+  CHECK (hipMemcpy (out.data (), devOut, 64 * sizeof (int),
+		    hipMemcpyDeviceToHost));
+}
```
aktemur, line 60: Return statement is missing.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+int
+main ()
+{
+  int *devOut;
+
+  CHECK (hipMalloc (&devOut, 64 * sizeof (int)));
+
+  kern<<<2, 32>>> (devOut);
```
aktemur: Can we use `<<<1, 1>>>` for simplicity?

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
```diff
@@ -0,0 +1,63 @@
...
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test the maint info amdgpu address-aliases
```
aktemur: Period at the end.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    gdb_breakpoint [gdb_get_line_number "Break here"] -allow-pending
+    gdb_continue_to_breakpoint "kern"
+
+    set seen_header 0
+    set seen_address 0
+
+    # The actual output can vary, we want to verify if GDB doesn't
+    # have any latent error.
```
aktemur, line 44: Have you considered using `pipe maintenance print address-spaces | wc -l` to count the number of address spaces and then requiring in the test below to match rows that many times?

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
```diff
@@ -0,0 +1,63 @@
...
+	    incr seen_address
+	    exp_continue
+	}
+
+	-re -wrap "" {
+	    # Done.
+	    gdb_assert { $seen_header == 1 && $seen_address > 0 } $gdb_test_name
+	}
+    }
+}
```
aktemur, line 69: Just an idea to consider: printing address aliases of nullptr (i.e. `0`) may be an interesting testcase.

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,150 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+	{
+	  ui_out->field_string ("addr", "<not reachable>");
+	  ui_out->field_string ("valid", "0x0");
+	}
+      else
+	{
+	  CORE_ADDR dst_address
+	    = amdgpu_segment_address_to_core_address (address_space.id,
+						      dst_offset);
+	  ui_out->field_string ("addr", paspace_and_addr (gdbarch,
```
aktemur, line 2058: Here we print the aspace as part of the address. It was already printed in the first column. This is a duplication. On the other hand, this is how aspace'd addresses are displayed to the user. So, I'm not asking for a change, but just making a remark about this. We will have an overall discussion about address spaces and the UI at some point in near future.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.cpp
palves: Formatting, no space between "* gidx_p".

https://sourceware.org/gdb/wiki/Internals%20GDB-C-Coding-Standards#Whitespaces

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
palves: Incomplete sentence.  Suggest:

```
Test the "maint info amdgpu address-aliases" command.
```

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,150 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+							  dst_address),
+				address_style.style ());
+
+	  /* Not a real alias, NULL just maps to NULL in every
+	     space.  */
+	  if (src_is_null_address)
+	    ui_out->field_string ("valid", "<null>");
```
aktemur, line 2065: IMHO it wouldn't be wrong to print 0 here.

gdb/amdgpu-tdep.c
```diff
@@ -2430,6 +2579,17 @@ INIT_GDB_FILE (amdgpu_tdep)
 		    amdgpu_supports_arch_info);
   gdb::observers::inferior_appeared.attach (amdgpu_observer_inferior_appeared,
 					    "amdgpu_tdep");
+
+  add_basic_prefix_cmd ("amdgpu", class_maintenance, _("\
+Commands for showing internal info about the AMDGPU target."),
+			    &maintenance_info_amdgpu_list, 0,
+			    &maintenanceinfolist);
```
aktemur: Indentation doesn't look right. Please check.

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,150 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+
+  int addr_width = 4 + (gdbarch_ptr_bit (gdbarch) / 4);
+
+  struct ui_out *ui_out = current_uiout;
+  ui_out_emit_table table_emitter (ui_out, 3, aspaces.size (), "aliases");
+
+  ui_out->table_header (max_aspace_name_len, ui_left, "aspace",
+			_("Address space"));
+  /* Address form is: <Name>0xNN..NNN.  */
+  ui_out->table_header (addr_width + max_aspace_name_len,
```
palves: I'd write as `max_aspace_name_len + addr_width` to match the order in the comment.

gdb/amdgpu-tdep.c
```diff
@@ -2430,6 +2579,17 @@ INIT_GDB_FILE (amdgpu_tdep)
...
 					    "amdgpu_tdep");
+
+  add_basic_prefix_cmd ("amdgpu", class_maintenance, _("\
+Commands for showing internal info about the AMDGPU target."),
+			    &maintenance_info_amdgpu_list, 0,
+			    &maintenanceinfolist);
+
+  add_cmd ("address-aliases", class_maintenance,
+	   amdgpu_info_address_aliases_command,
+	   _("Show address aliases across AMDGPU address spaces."),
```
palves: Sorry for pressing the same again -- this is still not saying enough.   Note it doesn't say that it takes an argument.

Help for commands nowadays shows a "Usage" block.  See `grep "Usage:" gdb/*.c` for many examples.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
czidev-amd: Thanks for catching it up! Indeed my regexp doesn't catch it.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
czidev-amd: The number of address spaces depends on the dbgabi implementation, AFAIK there is not an agreed minimum number of addresses spaces guaranteed.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
czidev-amd: One of the previous versions have it, but it does add little to the test.

gdb/doc/gdb.texinfo
```diff
@@ -45246,6 +45246,32 @@ packet history.
 @item maint info jit
 Print information about JIT code objects loaded in the current inferior.
 
+@kindex maint info amdgpu address-aliases
+@item maint info amdgpu address-aliases @var{expression}
+Evaluate @var{expression} to an address, and show the address it
+aliases to in every address space known to the current @acronym{AMD}
+@acronym{GPU} architecture, along with how many consecutive bytes each
+alias stays valid for.  The @samp{Valid for} column shows the number
```
lancesix: Seem that the end of the previous sentence ("along with how many consecutive bytes each alias stays valid for") and the last sentence ("The @samp{Valid for} column...") say the same thing.  Probably only keep the last one.

gdb/doc/gdb.texinfo
```diff
@@ -45246,6 +45246,32 @@ packet history.
...
+private_wave            private_wave#0xc00               0x4
+private_lane             private_lane#0x30             0x3d0
+local                      <not reachable>               0x0
+@end smallexample
+
+This only works while stopped in an @acronym{AMD} @acronym{GPU} thread.
+Some address spaces are private to a wave or a lane of a wave, and
+@value{GDBN} converts using whichever wave and lane are currently
+selected.  Not every address is reachable from every space: if it isn't,
+the row for that space reads @samp{<not reachable>}.
```
lancesix, line 45272: One nit, I think that we do an LDS -> global conversion, but this is only valid (ish) while the wave is stopped.  As soon as we resume, this address is not valid anymore.  Not sure if this should be described, or maybe this is just a dbgapi bug we should address.

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,150 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+  if (status != AMD_DBGAPI_STATUS_SUCCESS)
+    error (_("amd_dbgapi_get_architecture failed (%s)"),
+	   get_status_string (status));
+
+  arch_addr_space_id src_aspace_num
+    = amdgpu_address_space_id_from_core_address (addr);
+  /* Use GENERIC (a.k.a flat) address space as default.  Global is a subset
+     of generic, so this is safe to do.  */
+  if (src_aspace_num == DWARF_GLOBAL_ADDR_CLASS)
+    src_aspace_num = DWARF_GENERIC_ADDR_CLASS;
```
lancesix: Is this needed?   There is no other place where GDB has such assumption baked in, I would probably advise not doing this unless absolutely necessary.

gdb/amdgpu-tdep.c
lancesix: given the address are currently shown as "aspace#offset" everywhere, I would continue dong this here.  Refactor is indeed a broader question for another time.

gdb/amdgpu-tdep.c
czidev-amd: Removed.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.cpp
czidev-amd: Fixed, thank you

gdb/amdgpu-tdep.c
czidev-amd: Fixed, thank you

gdb/testsuite/gdb.rocm/maint-print-address-aliases.cpp
czidev-amd: Fixed, thank you

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
czidev-amd: Fixed, thank you

gdb/amdgpu-tdep.c
czidev-amd: indeed, the indentation was off.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
czidev-amd: Fixed, thank you.

gdb/amdgpu-tdep.c
czidev-amd: Updated with Usage block.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
palves: @czidev-amd I don't think Baris was suggesting to hardcode anything.  Instead, the count of address spaces shown by "maint print address-spaces" should match the count of address spaces shown by "maint info amdgpu address-aliases".  Seems reasonable to me.  

I'd probably do the counting in tcl, instead of relying on pipe + wc -- even though I was worried that wouldn't work on Windows, but it turns out it does (assuming MSYS2 environment).

gdb/doc/gdb.texinfo
```diff
@@ -45246,6 +45246,31 @@ packet history.
 @item maint info jit
 Print information about JIT code objects loaded in the current inferior.
 
+@kindex maint info amdgpu address-aliases
```
aktemur, line 45348: In commit message: "It also show" -> "It also shows"

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
```diff
@@ -0,0 +1,69 @@
...
+
+	-re "^\r\n\[a-z_\]+\[ \t\]+<not reachable>\[ \t\]*($::hex|<null>)\[^\r\n\]*(?=\r\n)" {
+	    # Match <not reachable> addresses.
+	    incr seen_address
+	    exp_continue
+	}
+
+	-re -wrap "" {
+	    # Done.
+	    gdb_assert { $seen_header == 1 && $seen_address > 0 } $gdb_test_name
```
aktemur: Nit: this could be split into two separate asserts, like done in the other PR.

gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.exp
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    gdb_breakpoint [gdb_get_line_number "Break here"] -allow-pending
+    gdb_continue_to_breakpoint "kern"
+
+    set seen_header 0
+    set seen_address 0
+
+    # The actual output can vary, we want to verify if GDB doesn't
+    # have any latent error.
```
czidev-amd: Added, thank you.

gdb/doc/gdb.texinfo
czidev-amd: done.

gdb/testsuite/gdb.rocm/maint-print-address-aliases.cpp
czidev-amd: solved

gdb/testsuite/gdb.rocm/maint-print-address-aliases.exp
czidev-amd: done.

gdb/amdgpu-tdep.c
czidev-amd: I can do that, let's see if anyone else has a different idea.

gdb/doc/gdb.texinfo
czidev-amd: Done.

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,145 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
   return aliases;
 }
 
+/* Implements the "maintenance info amdgpu address-aliases" command.
+
```
lumachad: GDB function comments should be written in third-person present tense describing what the function does, not "Implements the". Consider:

```
/* Show for each address space supported by the current AMDGPU
   architecture the aliases of a given ADDR, and the number of
   consecutive bytes for which each aliasing is valid.  */
```

"Implements the" reads more like a commit message than a source comment.

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,145 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+     as a real hit below.  */
+  amd_dbgapi_segment_address_t src_null_address;
+  status = amd_dbgapi_address_space_get_info
+    (src_aspace, AMD_DBGAPI_ADDRESS_SPACE_INFO_NULL_ADDRESS,
+     sizeof (src_null_address), &src_null_address);
+  if (status != AMD_DBGAPI_STATUS_SUCCESS)
+    error (_("amd_dbgapi_address_space_get_info (NULL_ADDRESS) failed (%s)"),
+	   get_status_string (status));
+  bool src_is_null_address = (offset == src_null_address);
+
```
lumachad, line 2003: The outer parentheses around the comparison are unnecessary in C++:

```c
bool src_is_null_address = offset == src_null_address;
```

gdb/amdgpu-tdep.c
```diff
@@ -1926,6 +1931,145 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+  int addr_width = 4 + (gdbarch_ptr_bit (gdbarch) / 4);
+
+  struct ui_out *ui_out = current_uiout;
+  ui_out_emit_table table_emitter (ui_out, 3, aspaces.size (), "aliases");
+
+  ui_out->table_header (max_aspace_name_len, ui_left, "aspace",
+			_("Address space"));
+  /* Address form is: <Name>0xNN..NNN.  */
+  ui_out->table_header (max_aspace_name_len + addr_width,
```
lumachad, line 2028: The comment on the next line (`/* Address form is: <Name>0xNN..NNN.  */`) is inaccurate.  `paspace()` emits a `#` separator between the address-space name and the hex address (e.g., `generic#0x2000000000030`), and for the default space it emits nothing at all.  Suggested wording:

```c
/* Address form is: <name>#0xNN..NNN for a named space,
   or just 0xNN..NNN for the default (global) space.  */
```

gdb/amdgpu-tdep.c
```diff
@@ -2430,6 +2574,18 @@ INIT_GDB_FILE (amdgpu_tdep)
...
 					    "amdgpu_tdep");
+
+  add_basic_prefix_cmd ("amdgpu", class_maintenance, _("\
+Commands for showing internal info about the AMDGPU target."),
+			&maintenance_info_amdgpu_list, 0,
+			&maintenanceinfolist);
+
+  add_cmd ("address-aliases", class_maintenance,
+	   amdgpu_info_address_aliases_command, _("\
+Show ADDRESS aliases across AMDGPU address spaces.\n\
```
lumachad, line 2587: The commit message headline says **"maint amdgpu info address-alias"** (singular, and with `amdgpu` before `info`), but the registered command is **"maintenance info amdgpu address-aliases"** (plural, correct order).  Please update the commit headline to match, e.g.:

    gdb: Add "maint info amdgpu address-aliases" command

gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.cpp
```diff
@@ -0,0 +1,60 @@
+/* Copyright (C) 2026 Free Software Foundation, Inc.
+   Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
```
lumachad, line 2: The filename says `maint-print-address-aliases` but the command under test is `maint info amdgpu address-aliases` (note `info`, not `print`).  Consider renaming both test files to `maint-info-amdgpu-address-aliases.{cpp,exp}` so they are discoverable by command name.

gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.exp
```diff
@@ -0,0 +1,92 @@
...
+    gdb_breakpoint [gdb_get_line_number "Break here"] -allow-pending
+    gdb_continue_to_breakpoint "kern"
+
+    set asname "\[^ \t\r\n\]+"
+    set trailer "\[^\r\n\]*(?=\r\n)"
+
+    # Count address spaces known to the architecture.
+    set aspaces 0
+    gdb_test_multiple "maint print address-spaces" "" -lbl {
+	-re "\r\n${asname}\[ \t\]+$::decimal\[ \t\]+" {
```
lumachad, line 57: The address-space count is obtained from `maint print address-spaces` and then cross-checked against the row count from `maint info amdgpu address-aliases`.  If those two commands ever diverge (e.g. one filters address spaces the other doesn't), the assertion gives a confusing failure with no obvious cause.  It would be more robust to count the rows produced by the `address-aliases` command itself.

gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.exp
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	-re "^\r\n${asname}\[ \t\]+$::hex\[ \t\]+$::hex${trailer}" {
+	    # Match global address space.
+	    incr seen_address
+	    exp_continue
+	}
+
+	-re "^\r\n${asname}\[ \t\]+${asname}#*$::hex\[ \t\]+($::hex|<null>)${trailer}" {
+	    # Match address space.
```
lumachad, line 85: `#*` matches zero or more `#` characters, so this branch can fire even when there is no `#` in the address field, potentially stealing matches from the `global` branch above.  The separator is always exactly `#` for named spaces, so use a literal `#`:

```tcl
-re "^\r\n${asname}\[ \t\]+${asname}#$::hex\[ \t\]+($::hex|<null>)${trailer}" {
```

lumachad (commented):
Some automated comments. Take it or leave it. They seem mostly nits to me.

gdb/amdgpu-tdep.c
czidev-amd: I like more the outer parentheses, make (in my opinion) the code easy to be read.

gdb/doc/gdb.texinfo
```diff
@@ -45345,6 +45345,31 @@ packet history.
...
+(gdb) maintenance info amdgpu address-aliases &idx
+Address space                      Address         Valid for
+global                      0x7ff7f4600c00               0x4
+generic            generic#0x2000000000030             0x3d0
+private_wave            private_wave#0xc00               0x4
+private_lane             private_lane#0x30             0x3d0
+local                      <not reachable>               0x0
+@end smallexample
+
+This only works while stopped in an @acronym{AMD} @acronym{GPU} thread.
```
palves, line 45367: Are we testing this (non-gpu thread cases)?  On a quick skim I didn't spot it.  It should be easy to cover:

  - running the command before running to main (i.e., test what happens when there's no thread)
  - running the command when focused on a host thread (e.g., right after running to main)

These will ensure that the command won't crash or something silly like that.

czidev-amd: Added two new tests as instructed. Please check.

lumachad:
@czidev-amd Is this waiting on some blocker? Otherwise if it's not ready, we might want to make it a draft while it's being worked on.

gdb/doc/gdb.texinfo
lancesix: Is "This only works while stopped in an @acronym{AMD} @acronym{GPU} thread." meant to address this particular comment?  It feels that this sentence says that the command can only be (meaningfully) used when stopped in a wave, but it does not really say that the conversion might be rendered invalid as soon as the wave resumes.

gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.exp
```diff
@@ -0,0 +1,103 @@
...
+if {[prepare_for_testing "failed to prepare ${testfile}" $testfile $srcfile {debug hip}]} {
+    return -1
+}
+
+with_rocm_gpu_lock {
+
+    # Test: command before any inferior is running (no thread selected).
+    gdb_test "maint info amdgpu address-aliases &main" \
+	"Current architecture is not AMDGPU\\." \
+	"address-aliases with no thread selected"
```
lancesix, line 36: not sure how things would react if we did `set architecture amdgcn:gfx900` or something here.  The current arch would be AMDGPU, but we do not have a process to do any dbgapi call.

gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.exp
```diff
@@ -0,0 +1,103 @@
...
+
```
lancesix, line 37: this particular command do not need to be in the `with_rocm_gpu_lock` block.

gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.exp
lancesix: I do expect that both commands must give the same number of lines.  Checking address-aliases against itself is not really testing anything.

gdb/amdgpu-tdep.c
```diff
@@ -1930,6 +1935,143 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+    error (_("Current thread is not an AMDGPU GPU thread."));
+
+  struct value *val = expr->evaluate ();
+  if (TYPE_IS_REFERENCE (val->type ()))
+    val = coerce_ref (val);
+  CORE_ADDR addr = value_as_address (val);
+
+  amd_dbgapi_architecture_id_t architecture_id;
+  amd_dbgapi_status_t status = amd_dbgapi_get_architecture
+    (gdbarch_bfd_arch_info (gdbarch)->mach, &architecture_id);
```
lancesix, line 1972: I suspect this could cause issues when dealing with generic code objects.

See https://github.com/ROCm/ROCgdb/pull/336/changes/0525aa4697836cc05a02a97a7389336f9e903773.

Might be worth testing with the generic board to check if this is needed.

gdb/amdgpu-tdep.c
```diff
@@ -1930,6 +1935,143 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+
+  /* First pass: size the "Address space" column to the longest name.
+     Initialize with the header title width so the column is never
+     narrower than the label.  */
+  size_t max_aspace_name_len = strlen (_("Address space"));
+  for (const auto &address_space : aspaces)
+    max_aspace_name_len = std::max (max_aspace_name_len,
+				    strlen (address_space.name.get ()));
+
+  int addr_width = 4 + (gdbarch_ptr_bit (gdbarch) / 4);
```
lancesix, line 2019: could be const.

gdb/amdgpu-tdep.c
```diff
@@ -1930,6 +1935,143 @@ amdgpu_get_watchable_aliases (struct gdbarch *gdbarch,
...
+
+      ui_out_emit_tuple tuple_emitter (ui_out, "alias");
+
+      amd_dbgapi_segment_address_t dst_offset;
+      amd_dbgapi_size_t cont_bytes;
+      status = amd_dbgapi_convert_address_space (wave_id, lane_id, src_aspace,
+						 offset, dst_aspace,
+						 &dst_offset, &cont_bytes);
+      ui_out->field_string ("aspace", address_space.name.get ());
+      if (status != AMD_DBGAPI_STATUS_SUCCESS)
```
lancesix, line 2050: This should be a test against `AMD_DBGAPI_STATUS_ERROR_INVALID_ADDRESS_SPACE_CONVERSION`.  We might want to error out for other cases (at least report an error, not the "not reachable").

gdb/amdgpu-tdep.c
lancesix: 0 is not always the null address.  For LDS, or private_lane, 0 is a valid address, different from the null one.  I think there is value highlighting the null address as such.

gdb/doc/gdb.texinfo
czidev-amd: New developments here: https://github.com/ROCm/ROCgdb/pull/225#discussion_r4015180654

gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.exp
czidev-amd: I have added 3 new tests: setting the architecture to gfx900, asking for the info, setting back the architecture to auto. Everything works as expected.

gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.exp
czidev-amd: done

gdb/amdgpu-tdep.c
czidev-amd: No issue detected on my gfx1201. I suppose I don't need to add the workaround as shown in the PR^

gdb/amdgpu-tdep.c
czidev-amd: Done

gdb/amdgpu-tdep.c
czidev-amd: Done

czidev-amd (commented):
New update with Lancelot requested changes:
- Updated the gdb.texinfo file to reflect when the conversion is valid,
- Update the test with 3 new tests to inspect what is going on when there is no dbgapi context.
- Update the implementation.

gdb/doc/gdb.texinfo
```diff
@@ -45376,6 +45376,34 @@ packet history.
...
+global                      0x7ff7f4600c00               0x4
+generic            generic#0x2000000000030             0x3d0
+private_wave            private_wave#0xc00               0x4
+private_lane             private_lane#0x30             0x3d0
+local                      <not reachable>               0x0
+@end smallexample
+
+This only works while stopped in an @acronym{AMD} @acronym{GPU}
+thread.  Note that addresses resulting from conversion to the global
+address space are only valid while the wave remains stopped; they
```
lancesix: Not sure this should be specific to "to the global address space".

> Note that addresses conversions might only be valid while the wave remains stopped and may become invalid as soon as execution resumes.

Then it is up to the person knowing which one is valid or not.

For example, private_wave, private_lane or generic to global will be stable, but local will not.

gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.cpp
```diff
@@ -0,0 +1,60 @@
...
+    {
+      int _gidx = blockIdx.x;
+      int _gdimx = blockDim.x;
+      gidx = _gidx;
+      gdimx = _gdimx;
+      volatile int *gidx_p = &gidx;
+      volatile int *gdimx_p = &gdimx;
+      volatile auto unused = gidx + gdimx; /* Break here.  */
+    }
+  __syncthreads ();
```
lancesix: the kernel is submited with just one lane active.  `__syncthreads` will be a no-op, and `if (idx == 0)` will always be true for that only wave running.  I suggest we should be able to remove this.

gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+  out[gidx * gdimx + idx] = threadIdx.x;
+}
+
+int
+main ()
+{
+  int *devOut;
+
+  CHECK (hipMalloc (&devOut, 64 * sizeof (int)));
```
lancesix: why 64?

gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.exp
```diff
@@ -0,0 +1,117 @@
...
+	    exp_continue
+	}
+
+	-re "^\r\n${asname}\[ \t\]+$::hex\[ \t\]+$::hex${trailer}" {
+	    # Match global address space.
+	    incr seen_address
+	    exp_continue
+	}
+
+	-re "^\r\n${asname}\[ \t\]+${asname}#$::hex\[ \t\]+($::hex|<null>)${trailer}" {
```
lancesix: Can't this be folded with the previous one (i.e. `(${aspace}#)?` to make the aspace and # optional.

gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.exp
```diff
@@ -0,0 +1,117 @@
...
+	    # Match address space.
+	    incr seen_address
+	    exp_continue
+	}
+
+	-re "^\r\n${asname}\[ \t\]+<not reachable>\[ \t\]+($::hex|<null>)${trailer}" {
```
lancesix: is <null> expected here?  The example in the manual shows 0x0.

gdb/testsuite/gdb.rocm/maint-info-amdgpu-address-aliases.exp
lancesix: also, can <null> appear in the "valid for" column?  Shouldn't it be in the address column?

PR #227, opened 2026-07-22
doc: add TESTING.md documenting testing strategy

spatrang:
## Summary

Adds a top-level `TESTING.md` that serves as the authoritative reference for how ROCgdb is validated and how engineers should write and run tests. This satisfies the org-wide requirement that every component repository document its testing strategy.

## What's included

The document is organized into 10 sections and covers the areas requested by the ticket:

- **Testing layers overview** — the three layers: selftests (unit), the DejaGnu integration testsuite, and performance tests.
- **Prerequisites & building for testing** — host toolchain, DejaGnu 1.5.3+, and the ROCm/GPU requirements (ROCdbgapi, HIP compiler, `--with-amd-dbgapi`), with an example development build.
- **Unit testing (selftests)** — the `maintenance selftest` mechanism, `--enable-unit-tests`/`GDB_SELF_TEST`, `gdbsupport/selftest.h`, and how selftests are run and wrapped by `gdb.gdb/unittest.exp`.
- **Integration / functional testing** — the DejaGnu testsuite layout, `make check-gdb` with `TESTS`/`RUNTESTFLAGS`, output files, result codes, and the ROCm/GPU slice (`gdb.rocm/`, `allow_hip_tests` skip logic, GPU lock, the `hip` board, and the `test_rocgdb.py` CI runner).
- **Performance testing** — the `gdb.perf/` framework, `check-perf`, and how baselines/regressions are handled.
- **When tests run (CI)** — the TheRock GitHub Actions workflows, per-PR/per-push cadence, and skip behavior.
- **How to write tests** — layer selection, `.exp` conventions, common helper procs, and what constitutes sufficient coverage.
- **Debugging test failures** and a **quick-command reference**.

## Notes

- Documentation only — no code or test behavior changes.
- Content was verified against the actual repository (testsuite Makefiles, `lib/rocm.exp`, `boards/hip.exp`, `.github/scripts/test_rocgdb.py`, and the CI workflows).

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -1,0 +1,144 @@
+# Testing ROCgdb
+
+**ROCgdb is AMD's fork of the [GNU Debugger (GDB)](https://www.gnu.org/software/gdb)
+that adds heterogeneous debugging support for AMD GPUs.** It inherits GDB's
+testing infrastructure and adds a GPU-specific test suite (`gdb.rocm/`) and CI on
+top of it. The goal is to keep ROCgdb correct for both CPU and GPU debugging as
+it tracks upstream GDB and successive ROCm releases.
+
+**Testing spans a broad surface.** ROCgdb must work across many AMD GPU
+architectures, host operating systems, and toolchains, on top of everything
+upstream GDB already supports. Testing every combination on every change is not
+practical, so tests are layered by cost: fast, high-signal suites run on every
+change, while configurations CI cannot cover per PR - for example GPU
+architectures without dedicated per-PR runner capacity - run nightly or on
+demand.
+
+**Testing should be accessible to contributors.** Wherever possible the same
+`make check` targets run on a developer's machine and in CI, so behavior can be
+validated locally first, with CI providing consistent, representative
+environments.
+
+> **Upstream is the source of truth.** GDB's testing conventions are defined
+> upstream, and where this document diverges, upstream wins. Read these first:
+> - `gdb/testsuite/README` - the reference for the GDB testsuite.
+> - <https://sourceware.org/gdb/wiki/TestingGDB> - the GDB testing wiki.
+>
+> This page summarizes and extends those conventions for ROCgdb; it is a
+> practical starting point, not a complete manual.
+
+## Testing principles
+
+### Make tests accessible during development
+
+Every fix or feature should come with a test that verifies the specific behavior
+being changed, and that a contributor can run quickly while developing. Selftests
+need no GPU and run in-process; the functional testsuite runs against a locally
+built debugger, so most behavior can be validated locally before CI.
+
+### Use layered validation
+
+ROCgdb is validated in three layers, cheapest first. Prefer the cheapest layer
+that can catch a given problem:
+
+- **Unit tests (selftests)** - in-process C++ checks compiled into GDB
+  (`gdb/unittests/`); run with `maintenance selftest` and also exercised by the
+  functional testsuite.
+- **Functional tests** - end-to-end DejaGnu tests under `gdb/testsuite/gdb.*/`,
+  including the GPU suite `gdb.rocm/`. This is where most ROCgdb testing happens.
+- **Performance tests** - the perftest harness under `gdb/testsuite/gdb.perf/`
+  (currently exercised only rarely).
+
+### Scale coverage to available resources
+
+High-signal suites run on every change; less common configurations run nightly or
+on demand. Some GPUs cannot debug multiple processes concurrently, so GPU tests
+serialize device access with a lock (`with_rocm_gpu_lock`); `-j$(nproc)` therefore
+speeds up compilation more than GPU test execution.
+
+### Use static analysis for mechanical checks
+
+Formatting and other mechanical checks run as pre-commit hooks in CI
+(`.github/workflows/pre-commit.yml`), keeping human review focused on substance.
+
+### Add reliable tests to required CI
+
+CI runs the testsuite through `.github/scripts/test_rocgdb.py` and compares
+results against a known-failures list. Only add tests that are deterministic, so
+required CI stays trustworthy and failures are actionable.
+
+## Testing changes to ROCgdb
+
+### Core, target-independent changes
+
+Cover self-contained logic (utilities, containers, parsers) with a **selftest**
+next to the code, and observable debugger behavior with a **DejaGnu** test in the
+matching `gdb.*/` directory, chosen by area: core debugger features (`gdb.base`),
+architecture (`gdb.arch`), language support such as C++ (`gdb.cp`) and Python
+(`gdb.python`), the machine interface (`gdb.mi`), and so on. See
+[Run the same tests locally and in CI](#run-the-same-tests-locally-and-in-ci) for
+how to run a chosen subset.
+
+### AMDGPU and ROCm-specific changes
+
+Add a test under `gdb.rocm/`. A GPU test loads `rocm.exp`, gates on a capability
+check (`require allow_hip_tests`), compiles its program as HIP device code, and
+serializes device access with `with_rocm_gpu_lock`. For the mechanics of writing
+`.exp` tests, follow `gdb/testsuite/README` and copy an existing `gdb.rocm/` test
+as a template.
+
+### Performance-sensitive changes
+
+Add or run a **perf** test under `gdb.perf/`. Performance tests are not part of
+`make check`; run them on demand and compare measurements before and after on the
+same hardware. ROCgdb does not currently enforce automated regression thresholds
+in CI. See `gdb/testsuite/gdb.perf/README`.
+
+## Testing ROCgdb against ROCm
+
+### Build with amd-dbgapi
+
+GPU debugging requires ROCgdb built `--with-amd-dbgapi` and a working ROCm stack
+(the amd-dbgapi library and a HIP compiler). Without these the GPU capability
+gate reports the `gdb.rocm/` tests as unsupported and they do not run.
+
+### Keep test environments reproducible
+
+The testsuite reads the ROCm environment it needs (e.g. `ROCM_PATH`) and manages
+any GPU-test-specific variables where required, so a correct ROCm install and an
+amd-dbgapi-enabled build are generally all that is needed. Avoid depending on ad
+hoc environment tweaks in individual tests.
+
+### Run the same tests locally and in CI
+
+The entry point is the GDB testsuite in both cases. CI wraps it with
+`.github/scripts/test_rocgdb.py`, which runs the CPU and GPU test sets across the
+supported toolchains, retries flaky failures, and applies the known-failures
+list. It is also the easiest way to reproduce a CI run locally; run it with
+`--help` for the current options.
+
+## Validating ROCgdb on hardware and in CI
+
+### Test on supported GPU hardware
+
+GPU results are only meaningful on supported devices. On unsupported or mixed
+hardware, expect real failures rather than clean skips: ensure every visible
+device is supported, or restrict the run to a supported device with
+`ROCR_VISIBLE_DEVICES`.
+
+### When tests run
+
+CI is implemented with GitHub Actions under `.github/workflows/`. On pull
+requests and pushes to the staging branches, ROCgdb is built and tested on both
+CPU and GPU runners. At present each PR runs the GPU suite (`gdb.rocm`, excluding
+corefile tests) on GPU runners and `gdb.dwarf2` on CPU runners, with GPU corefile
+tests being added on a dedicated runner. Documentation-only changes and PRs
+labeled to skip CI do not trigger full runs. The workflow files are the source of
+truth for exact triggers and cadence.
+
+### Read results and triage failures
+
+Results are written in the build tree as `gdb.sum` (summary) and `gdb.log`
+(detail), with per-test artifacts under `outputs/` for replaying a failure. They
+use the standard DejaGnu codes (`PASS`, `FAIL`, `UNTESTED`, `UNSUPPORTED`,
+`UNRESOLVED`, `XFAIL`, `KFAIL`); see `gdb/testsuite/README` for their meanings.


TESTING.md
```diff
@@ -0,0 +1,580 @@
...
+---
+
+## 2. Prerequisites
+
+### Host toolchain (all builds)
+
+- A C++17 compiler (e.g. GCC 9+ or Clang 5+).
+- `bison`, `flex`, `make`, `texinfo`, and development headers for `ncurses`,
+  `zlib`, `expat`, `python3`, `lzma`, `gmp`, and `mpfr`. See `README-ROCM.md`
+  for the exact package lists per distribution.
```
lancesix: This is for building gdb, I would rather we just refer to `README-ROCM.md` or `gdb/README`.  Otherwise we risk having this information getting out of sync.

TESTING.md
```diff
@@ -0,0 +1,580 @@
...
+- [DejaGnu](https://www.gnu.org/software/dejagnu/) **1.5.3 or newer**, plus Tcl
+  and Expect, to run the functional testsuite (see `README-maintainer-mode`).
+
+### ROCm / GPU testing
+
+- **ROCm** installed (default root `/opt/rocm`, overridable via `ROCM_PATH`).
```
lancesix: in the "new world", is `/opt/rocm` still that special?

TESTING.md
```diff
@@ -0,0 +1,580 @@
...
+- **AMD Debugger API Library (ROCdbgapi)** and Comgr, provided by the
+  `rocm-dbgapi` package (`librocm-dbgapi.so`, `libamd_comgr.so`).
```
lancesix: This is a GDB dependency, nothing specific to testing. I would reference to other docs to have GDB running.

TESTING.md
```diff
@@ -0,0 +1,580 @@
...
+- A **HIP compiler** (`amdclang++`, discovered automatically) for compiling GPU
+  test programs. OpenMP-offload tests additionally need `amdclang`/`amdflang`.
+- Supported **AMD GPU hardware** with a supported `gfxNNN` architecture. GPU
+  tests are automatically skipped (reported `UNSUPPORTED`) when no compatible
+  device is present.
```
lancesix: I do not think this is true.  I expect a lot of failures for unsupported devices.  Also, to test, we need all the devices on the system to be supported, or isolate the supported one to run on.

TESTING.md
```diff
@@ -0,0 +1,580 @@
...
+# Parallel run
+make -j8 check-gdb FORCE_PARALLEL=1
+```
+
+Or directly from the testsuite directory:
+
+```shell
+cd build/gdb/testsuite
+make site.exp
+runtest GDB=../gdb gdb.rocm/alu-exceptions.exp
```
lancesix: that, or more usual us `make check TESTS...`.

Running `runtest GDB=../gdb` might have issues with `-data-directory` being missing.

TESTING.md
```diff
@@ -0,0 +1,580 @@
...
+| Variable | Meaning |
+|----------|---------|
+| `TESTS` | Space-separated `.exp` paths to run (wildcards allowed with GNU make) |
+| `RUNTESTFLAGS` | Extra flags for `runtest` (e.g. `GDB=...`, `--target_board=hip`, `-debug`) |
+| `GDB` | Debugger binary under test — a `runtest` variable, supplied via `RUNTESTFLAGS` (e.g. `RUNTESTFLAGS="GDB=..."`) rather than a make variable |
+| `FORCE_PARALLEL` | Force parallel mode even when `RUNTESTFLAGS` is set |
+| `RACY_ITER` | Repeat the run N times to surface flaky tests |
+
+`check-gdb` is the top-level build-tree target (it accepts `TESTS=` and
+`RUNTESTFLAGS=`). The other testsuite targets — `check`, `check-parallel`,
```
lancesix: It could be clearer that those targets are available when running from the test build directory `make -C build/gdb/testsuite`.

TESTING.md
```diff
@@ -0,0 +1,580 @@
...
+`hip_devices_support_debug_multi_process`. Because CI and developers often share
+GPUs, GPU access is serialized in parallel runs via `with_rocm_gpu_lock { ... }`.
+
+**Relevant environment variables.**
+
+| Variable | Effect |
+|----------|--------|
+| `ROCM_PATH` | ROCm install root (compiler, bundler, LLVM tools). Default `/opt/rocm`. |
+| `HCC_AMDGPU_TARGET` | Comma-separated `gfx` targets, overriding auto-detection. |
+| `HSA_COREDUMP_PATTERN` | GPU core-dump filename pattern (core-dump tests). |
```
lancesix: This env variable (as of today) will not impact the debugger testsuite.  The testsuite overrides it when needed, and restores it at the end.

TESTING.md
```diff
@@ -0,0 +1,580 @@
...
+require allow_hip_tests    ;# skip -> UNSUPPORTED if no suitable GPU/build
+
+standard_testfile .cpp     ;# derive testfile/binfile/srcfile from the .exp name
+
+if {[build_executable "failed to prepare $testfile" $testfile $srcfile \
+        {debug hip}]} {
+    return                 ;# compile failure -> UNTESTED
+}
+
+with_rocm_gpu_lock {
```
lancesix: Would be good to discuss this with_rocm_gpu_lock.

TESTING.md
```diff
@@ -0,0 +1,580 @@
+# Testing ROCgdb
+
+This document is the authoritative reference for how ROCgdb is validated and how
```
lancesix: I am not sure this is.  Any upstream requirement should take precedence.  This document can summarize upstream policy and extend on them, but in case of a divergence, the upstream policy should prevail.

Should be good to link early here the upstream documentations `gdb/testsuite/README`, https://sourceware.org/gdb/wiki/TestingGDB

lancesix (commented):
This is a good start, but I feel that this document gets into some details that are bound to get out of sync rapidly.  It discusses details of some function or script parameters, but I am afraid that those scripts / functions can end-up changing and causing discontinuity here.

lumachad:
On a first skim, I think the document is too verbose. We need to make it cover the points but be succint. We don't need to talk about dependencies required to test but rather focus on what test strategies we have and a quick summary of what they are, what they look like etc.

TESTING.md
spatrang: Done — removed the build/toolchain dependency list; the doc now defers to `README-ROCM.md` / `gdb/README` for build and prerequisites.

TESTING.md
spatrang: Done — dropped the `/opt/rocm` canonical-path assertion.

TESTING.md
spatrang: Done — removed; this is a GDB runtime dependency, so the doc now points to the build docs instead of restating it.

TESTING.md
spatrang: Done — corrected. The doc now states that on unsupported or mixed hardware you should expect real failures rather than clean skips, and that every visible device must be supported (or the run isolated to a supported one).

TESTING.md
spatrang: Done — the doc now recommends `make check`/`check-gdb TESTS=...` and notes that invoking `runtest` directly needs the environment (e.g. `-data-directory`) the make targets set up.

TESTING.md
spatrang: Done — clarified that these targets are run from the testsuite build directory (`make -C build/gdb/testsuite ...`).

TESTING.md
spatrang: Done — removed `HSA_COREDUMP_PATTERN` (the whole env-var table is gone); it doesn't affect the debugger testsuite.

TESTING.md
spatrang: Done — added a short explanation of `with_rocm_gpu_lock` (serializing shared-GPU access), and it appears in the illustrative test skeleton.

TESTING.md
spatrang: Done — reframed so upstream is the source of truth (upstream wins on divergence), and linked `gdb/testsuite/README` and the TestingGDB wiki up front.

spatrang:
### Update: rewrote TESTING.md per review feedback

Thanks for the reviews. I've reworked the document to address the two overarching points — that it was too verbose (@lumachad) and got into specifics that would drift out of sync (@lancesix) — and pushed the update 
Key changes:
- **Much shorter and strategy-focused** (~580 → ~180 lines): a "test strategies at a glance" table plus a brief section per strategy (selftests, DejaGnu functional incl. GPU, performance), when tests run, and how to write tests.
- **Defers to upstream as the source of truth**: no longer claims to be "authoritative"; links `gdb/testsuite/README` and the [TestingGDB wiki](https://sourceware.org/gdb/wiki/TestingGDB) up front, and points to `README-ROCM.md` / `gdb/README` for build/dependency details instead of duplicating them.
- **Dropped churn-prone specifics** (helper-proc listings, script parameter/flag tables, env-var tables); the CI runner now just points at `test_rocgdb.py --help`.
- **Fixed the accuracy issues** raised inline: GPU tests are not cleanly skipped on unsupported/mixed hardware (expect real failures), and `HSA_COREDUMP_PATTERN` was removed.
- Added a short, clearly **illustrative** `.exp` skeleton in "How to write tests" so the document still shows what a test looks like, pointing at the real `gdb.rocm/` tests as templates.

All individual review comments have been addressed and replied to inline. Please take another look.

TESTING.md
```diff
@@ -0,0 +1,181 @@
+# Testing ROCgdb
```
czidev-amd, line 1: I am not sure if we need a copyright banner here, please check with one of the maintainers.

TESTING.md
```diff
@@ -0,0 +1,181 @@
...
+They also run as part of the functional suite (`gdb.gdb/unittest.exp`), so a
+failing selftest is a normal test regression. New standalone, inferior-free
+logic should come with a selftest.
+
+## 2. Functional tests (DejaGnu)
+
+The bulk of ROCgdb's validation — including all GPU debugging — is end-to-end
+functional testing through the real debugger interface, using the DejaGnu
+framework under `gdb/testsuite/`. Tests are grouped by area in `gdb.*/`
+directories (e.g. `gdb.base/`, `gdb.cp/`, `gdb.python/`, `gdb.mi/`, and the
```
czidev-amd: That enumeration doesn't add anything to the grouped idea. I would suggest instead of enumerating the folders, give them a meaning:
- base tests (gdb.base)
- assembly tests (gdb,asm)
- cxx tests (gdb.cp)
- backend  specific tests (e.g. gdb.rocm)
etc.

TESTING.md
```diff
@@ -0,0 +1,181 @@
...
+GPU-specific `gdb.rocm/`). A test is an `.exp` script, usually paired with a
+source program of the same basename that DejaGnu compiles on the fly.
+
+### Running
+
+Use the `check-gdb` target from the top of the build tree, selecting tests with
+`TESTS`:
```
czidev-amd: I would add that is possible to run the gdb tests in parallel by passing `-j $(nproc)` to the command line

TESTING.md
```diff
@@ -0,0 +1,181 @@
...
+exact triggers and cadence.
+
+## 5. How to write tests
+
+Pick the layer that fits:
+
+- **No running inferior needed** (a pure helper, container, parser) → add a
+  **selftest** next to the code it covers.
+- **Observable debugger behavior** (breakpoints, stepping, printing, GPU/wave
+  handling) → add a **DejaGnu `.exp`** test in the matching `gdb.*/` directory
```
czidev-amd: Please replace the `→` symbol with something more friendly :)

TESTING.md
```diff
@@ -0,0 +1,181 @@
...
+  (`gdb.rocm/` for GPU behavior).
+- **Cost of an operation** → add a **perf** test under `gdb.perf/`.
```
czidev-amd: likewise

TESTING.md
```diff
@@ -0,0 +1,181 @@
...
+
+For functional tests, follow the upstream conventions in `gdb/testsuite/README`:
+an `.exp` driver plus a same-named source program, one focused feature per file,
+and precise output matching so real regressions are caught. Gate
+environment-dependent tests so they are skipped (not failed) where the feature
+genuinely cannot run, and keep tests parallel-safe. A GPU test additionally loads
+`rocm.exp`, requires the GPU capability gate, compiles its program as HIP device
+code, and serializes hardware access with `with_rocm_gpu_lock`.
+
+A GPU functional test looks roughly like this (illustrative skeleton — copy an
```
czidev-amd: likewise but for `—` symbol

czidev-amd (changes requested):
Please also check all the symbols in this file, use standard ones, do not use the one from extended ascii set.
Otherwise looks good to me.

lumachad:
You might want to keep an eye on https://github.com/ROCm/TheRock/issues/6711.

TESTING.md
spatrang: Good catch. I checked the emerging ROCm testing-doc standard (ROCm/TheRock#6711) and its Markdown docs (its `TESTING`/`CONTRIBUTING`/`README`) don't carry a copyright banner, and neither does ROCgdb's own `README-ROCM.md`. Markdown docs in this ecosystem conventionally omit it (unlike source/`.exp` files). So I've left it out to match that convention - happy to add an SPDX/AMD header instead if a maintainer prefers.

TESTING.md
spatrang: Done — replaced the bare folder list with categories that convey meaning (core `gdb.base`, architecture `gdb.arch`, language support `gdb.cp`/`gdb.python`, machine interface `gdb.mi`, and backend-specific such as the GPU suite `gdb.rocm`).

TESTING.md
spatrang: Done — added a parallel example: `make -j$(nproc) check-gdb TESTS=...`.

TESTING.md
spatrang: Done — replaced all `->`-style arrows with plain ASCII (`->`).

TESTING.md
spatrang: Done — same, replaced with ASCII `->`.

TESTING.md
spatrang: Done — replaced the em dashes with plain ASCII hyphens; the file is now pure ASCII (verified no extended-ASCII characters remain).

spatrang:
### Update: addressed @czidev-amd's review + light alignment with the ROCm testing-doc standard

Pushed an update:
- **ASCII only** — replaced all extended-ASCII characters (em dashes and arrows) with plain ASCII; verified none remain.
- **Test directories described by meaning** rather than a bare list (core `gdb.base`, architecture `gdb.arch`, language support `gdb.cp`/`gdb.python`, machine interface `gdb.mi`, backend-specific such as the GPU suite `gdb.rocm`).
- **Parallel runs** — added a `make -j$(nproc) check-gdb TESTS=...` example.
- **Testing principles** — added a short section (layered validation, local-first, scale to available resources) that mirrors the emerging ROCm testing strategy in ROCm/TheRock#6711, which @lumachad pointed to.

On the copyright-banner question: TheRock's testing/CONTRIBUTING/README Markdown docs and ROCgdb's own `README-ROCM.md` don't carry a banner, so I left it out to match convention — happy to add an SPDX/AMD header if a maintainer prefers (left that thread open).

All other review threads have been addressed and resolved. PTAL.

czidev-amd (approved):
LGTM,

TESTING.md
```diff
@@ -0,0 +1,200 @@
...
+`check-readmore` I/O-stress variants, parallel and per-board variants, etc.) are
+most directly invoked from the testsuite build directory, e.g.:
+
+```shell
+make -C build/gdb/testsuite check-read1 TESTS="gdb.base/break.exp"
+```
+
+Invoking `runtest` directly is possible but not recommended for routine use - it
+needs the right environment (e.g. `-data-directory`) that the make targets set up
+for you. Prefer `make check`/`check-gdb`.
```
palves: runtest would also need to be run from the gdb/testsuite/ directory so that site.exp is picked up.  Better not even suggest this is possible, there is no need to run it directly.

TESTING.md
```diff
@@ -0,0 +1,200 @@
...
+# Whole suite
+make check-gdb
+
+# A subset or a single test
+make check-gdb TESTS="gdb.rocm/*.exp"
+make check-gdb TESTS="gdb.rocm/alu-exceptions.exp"
+
+# Run in parallel to speed things up
+make -j$(nproc) check-gdb TESTS="gdb.rocm/*.exp"
+```
```
lumachad: This doesn't work. We have serial locks that don't let things run in parallel. See https://github.com/ROCm/ROCgdb/pull/149

lumachad (dismissed):

TESTING.md
palves: It still speeds thing up though, as compilation is not under the locks, so you get parallel compilation.  I.e., the comment isn't wrong.

TESTING.md
spatrang: Done - removed the paragraph entirely; the doc no longer mentions running `runtest` directly and only documents the `make` targets.

spatrang: Done

spatrang:
### Update: addressed @palves's comment

- Removed the paragraph about invoking `runtest` directly; the doc now documents only the `make` targets (which set up `site.exp` / `-data-directory` for you).
- Kept the parallel `-j$(nproc)` example: as @palves noted, compilation still parallelizes even though GPU test execution is serialized by the GPU lock.

All review threads are resolved. PTAL.

spatrang:
### Update: restructured to follow the ROCm testing-doc outline

Reworked `TESTING.md` to follow the outline of the emerging ROCm-wide testing strategy in ROCm/TheRock#6711 (adapted, not verbatim):

- **Philosophy intro** + the existing "upstream is the source of truth" callout.
- **Testing principles** - make tests accessible during development, use layered validation, scale coverage to available resources, use static analysis for mechanical checks, add reliable tests to required CI.
- **Testing changes to ROCgdb** - by change type (core/target-independent, AMDGPU/ROCm-specific with the illustrative `gdb.rocm/` skeleton, performance-sensitive).
- **Testing ROCgdb against ROCm** - build with amd-dbgapi, reproducible environments, run the same tests locally and in CI.
- **Validating ROCgdb on hardware and in CI** - supported GPU hardware, when tests run, reading results / triage.

All earlier review fixes are preserved (ASCII-only, defers to upstream, no build-dep duplication, no direct `runtest`, GPU-lock/parallel nuance, accurate GPU skip semantics). PTAL.

TESTING.md
```diff
@@ -0,0 +1,167 @@
...
+that adds heterogeneous debugging support for AMD GPUs.** It inherits GDB's
+testing infrastructure and adds a GPU-specific test suite (`gdb.rocm/`) and CI on
+top of it. The goal is to keep ROCgdb correct for both CPU and GPU debugging as
+it tracks upstream GDB and successive ROCm releases.
+
+**Testing spans a broad surface.** ROCgdb must work across many AMD GPU
+architectures, host operating systems, and toolchains, on top of everything
+upstream GDB already supports. Testing every combination on every change is not
+practical, so tests are layered by cost: fast, high-signal checks run on every
+change, while broader and GPU-hardware-heavy runs happen nightly or on demand.
```
lumachad: We only run nightly and on-demand GPU tests for things the CI doesn't support or for architectures that don't have the capacity for per-PR tests.

TESTING.md
```diff
@@ -0,0 +1,167 @@
...
+> **Upstream is the source of truth.** GDB's testing conventions are defined
+> upstream, and where this document diverges, upstream wins. Read these first:
+> - `gdb/testsuite/README` - the reference for the GDB testsuite.
+> - <https://sourceware.org/gdb/wiki/TestingGDB> - the GDB testing wiki.
+>
+> This page summarizes and extends that for ROCgdb; it is a practical starting
+> point, not a complete manual. For building and installing ROCgdb, see
+> `README-ROCM.md` and `gdb/README` (build/runtime prerequisites are not
+> duplicated here). The outline mirrors the broader ROCm testing strategy in
+> [ROCm/TheRock#6711](https://github.com/ROCm/TheRock/issues/6711).
```
lumachad: Let's not mention things that aren't closely-related to testing and validation, like building and runtime dependencies.

Though we are using TheRock's as a base, we do not have to mention that.

TESTING.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Testing principles
+
+### Make tests accessible during development
+
+Prefer tests a contributor can run quickly on a normal workstation. Selftests
+need no GPU and run in-process; the functional testsuite runs against a locally
+built debugger with `make check`. Local runs give the fastest feedback before CI.
+
```
lumachad, line 38: To be honest, I don't quite understand what this is trying to say. We should strive for tests that validates/verifies the feature/bug being worked on.

TESTING.md
```diff
@@ -0,0 +1,167 @@
...
+### Use layered validation
+
+ROCgdb is validated in three layers, cheapest first. Prefer the cheapest layer
+that can catch a given problem:
+
+- **Selftests (unit tests)** - in-process C++ checks compiled into GDB
+  (`gdb/unittests/`), run with `maintenance selftest`.
+- **Functional tests** - end-to-end DejaGnu tests under `gdb/testsuite/gdb.*/`,
+  including the GPU suite `gdb.rocm/`.
+- **Performance tests** - the perftest harness under `gdb/testsuite/gdb.perf/`.
```
lumachad: Though one can run selftests by hand, those are also exercised by gdb's testsuite.

We should mention we have 3 types of tests, unit tests (selftests), functional tests and performance tests. I have to say the performance tests don't get exercised a lot.

TESTING.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Scale coverage to available resources
+
+Fast, high-signal tests run on every change. GPU hardware is limited and GPU
+test execution is serialized by a GPU lock, so broad GPU runs are reserved for
+nightly and on-demand CI. Passing `-j$(nproc)` still helps, since compilation is
+parallelized even though device execution is serialized.
```
lumachad: This is another one that doesn't quite say much to me. We don't serialize because GPU hardware it limited, but rather because some GPU's don't support multi-inferior debugging.

What exactly are we trying to say with this paragraph?

For every PR we currently run gdb.rocm (skipping corefile tests) in rocgdb-gpu, gdb.dwarf2 in rocgdb-cpu and soon we will run GPU corefile tests in rocgdb-corefile.

TESTING.md
```diff
@@ -0,0 +1,167 @@
...
+### Core, target-independent changes
+
+Cover self-contained logic (utilities, containers, parsers) with a **selftest**
+next to the code, and observable debugger behavior with a **DejaGnu** test in the
+matching `gdb.*/` directory, chosen by area: core debugger features (`gdb.base`),
+architecture (`gdb.arch`), language support such as C++ (`gdb.cp`) and Python
+(`gdb.python`), the machine interface (`gdb.mi`), and so on. Run a subset with:
+
+```shell
+make check-gdb TESTS="gdb.base/mytest.exp"
```
lumachad: Above it says we use test_rocgdb.py, but here we invoke it directly. Simply invoking it like this won't work with library paths that aren't in /

TESTING.md
```diff
@@ -0,0 +1,167 @@
...
+```
+
+### AMDGPU and ROCm-specific changes
+
+Add a test under `gdb.rocm/`. A GPU test loads `rocm.exp`, gates on a capability
+check (`require allow_hip_tests`), compiles its program as HIP device code, and
+serializes hardware access with `with_rocm_gpu_lock`. Illustrative skeleton (copy
+an existing `gdb.rocm/` test as your real template):
```
lumachad: It feels we're going into too much detail here, something that might be better suited to rocgdb's gdb/testsuite/README file.

TESTING.md
```diff
@@ -0,0 +1,167 @@
...
+`--help` for the current options.
+
+## Validating ROCgdb on hardware and in CI
+
+### Test on supported GPU hardware
+
+GPU results are only meaningful on supported devices. On unsupported or mixed
+hardware, expect real failures rather than clean skips: ensure every visible
+device is supported, or isolate the run to a supported device using the usual
+ROCm device-visibility controls.
```
lumachad: We might want to mention ROCR_VISIBLE_DEVICES here.

lumachad (commented):
A few comments on the current versions. I think this is mostly good. Just need to polish it a bit more and improve the signal to noise ratio.

TESTING.md
spatrang: Done - reworded: per-PR runs the high-signal suites, and configurations CI cannot cover per PR (e.g. GPU architectures without dedicated per-PR runner capacity) run nightly or on demand.

TESTING.md
spatrang: Done - dropped the build/runtime-dependency pointer and the TheRock reference; the page now sticks to testing/validation.

TESTING.md
spatrang: Reworded to your point: 'Every fix or feature should come with a test that verifies the specific behavior being changed, and that a contributor can run quickly while developing.'

TESTING.md
spatrang: Done - framed as the three types (unit/selftests, functional, performance); noted selftests are also exercised by the functional testsuite, that functional is where most testing happens, and that perf tests are currently exercised only rarely.

TESTING.md
spatrang: Fixed the reasoning and the intent: serialization is because some GPUs cannot debug multiple processes concurrently (not limited hardware). I also added the concrete per-PR cadence you described (gdb.rocm minus corefile on GPU runners, gdb.dwarf2 on CPU, corefile runner coming) to the 'When tests run' section.

TESTING.md
spatrang: Done - removed the direct 'make check-gdb ...' invocation; the doc now points to the 'Run the same tests locally and in CI' section (test_rocgdb.py), which handles library paths.

TESTING.md
spatrang: Done - removed the skeleton; the section now defers to gdb/testsuite/README and to copying an existing gdb.rocm/ test as a template.

TESTING.md
spatrang: Done - now mentions restricting the run to a supported device with ROCR_VISIBLE_DEVICES.

spatrang:
### Update: addressed @lumachad's review

Pushed changes addressing all comments from the latest review:

- **Nightly/on-demand cadence** reworded: high-signal suites run per PR; configurations CI cannot cover per PR (e.g. GPU architectures without dedicated per-PR runner capacity) run nightly or on demand.
- **Removed** the build/runtime-dependency pointer and the TheRock reference from the intro; the page now stays focused on testing and validation.
- **"Make tests accessible"** reworded to your point: every fix or feature should come with a test that verifies the specific behavior being changed.
- **Three test types** framed explicitly (unit/selftests, functional, performance); noted selftests are also exercised by the functional testsuite, that functional is where most testing happens, and that perf tests are currently exercised only rarely.
- **Serialization reason corrected**: GPU tests serialize device access because some GPUs cannot debug multiple processes concurrently (not because hardware is limited). Added the concrete per-PR cadence (gdb.rocm minus corefile on GPU runners, gdb.dwarf2 on CPU, GPU corefile runner coming) under "When tests run".
- **Dropped** the direct `make check-gdb ...` invocation (won't work with non-standard library paths); the doc points to running via `test_rocgdb.py`.
- **Removed** the illustrative `.exp` skeleton; the doc now defers to `gdb/testsuite/README` and to copying an existing `gdb.rocm/` test as a template.
- **Added** `ROCR_VISIBLE_DEVICES` as the way to restrict a run to a supported device.

lumachad (approved):
I'm happy with this one as a first draft. We can iterate on it going forward.

PR #229, opened 2026-07-23
merge origin/master 20260723

aktemur:
There is one important conflict that needed to be resolved. I added the following code in `gdb/testsuite/lib/gdb.exp`

```tcl
        # Also add -O0 to early flags for the HIP language.  If
        # "optimize" is also requested, another -O flag (e.g. -O2)
        # will be added to the flags, overriding this -O0.
        set hip_early_flags "-O0 $hip_early_flags"
```


lancesix:
The downstream version has dropped the explicit -O0.

The bit

```
+       # Also add -O0 to early flags for the HIP language.  If
+       # "optimize" is also requested, another -O flag (e.g. -O2)
+       # will be added to the flags, overriding this -O0.
+       set hip_early_flags "-O0 $hip_early_flags"
```

looks like it does not belong to this merge.  If we want to reinstate the early -O0, that would be a patch on its own.

aktemur:
> The downstream version has dropped the explicit -O0.
> 
> The bit
> 
> ```
> +       # Also add -O0 to early flags for the HIP language.  If
> +       # "optimize" is also requested, another -O flag (e.g. -O2)
> +       # will be added to the flags, overriding this -O0.
> +       set hip_early_flags "-O0 $hip_early_flags"
> ```
> 
> looks like it does not belong to this merge. If we want to reinstate the early -O0, that would be a patch on its own.

I removed this part.

spatrang (approved):
Reviewed **lib/gdb.exp** for the hipcc→amdclang++ part. The amdclang++ HIP handling is fully preserved (spill-cfi early_flags, `--hip-link`, `--rocm-path`, `--offload-arch`, and the `-x hip`/`-x` none tagging). No conflict markers, Tcl parses clean, and it correctly kept our amdclang++ block over upstream's hipcc version. Only diff inside the HIP block vs amd-staging is one cosmetic blank line. LGTM.

palves (approved):
Thanks, looks like the more troublesome conflicts where both from my patches upstream.  Judging from "git show --remerge-diff", the resolutions look good.

PR #231, opened 2026-07-23
gdb/testsuite: restore -O0 for amdclang++ HIP test compiles

lfmeadow:
Commit `ab3a83eb` migrated the gdb.rocm testsuite from `hipcc` to `amdclang++` and dropped `-O0` from `hip_early_flags`, assuming `amdclang++` defaults to `-O0` for `-x hip` compiles. With WW28-era compiler builds that is not always true.

Without `-O0`, device helper functions can be emitted with `DW_AT_calling_convention (DW_CC_nocall)`. rocgdb then warns that functions "do not follow the target calling convention" and tests such as `finish.exp` and `device-barrier.exp` fail (LCOMPILER-2452: 32 unexpected failures on MI300X with the WW28 tarball).

Problem:
- WW28 build + WW28 amdclang++ harness (no `-O0`) → 32 failures on MI300X
- Same build + WW25 hipcc harness (with `-O0`) → pass
- Same build + this patch → pass (122 pass, 0 unexpected fail on MI300X)

Manual compile on device objects from `finish0.o`:
- `amdclang++ -x hip` without `-O0` → `DW_CC_nocall` on helpers
- `amdclang++ -x hip -O0` → no `DW_CC_nocall`

Solution:
Restore explicit `-O0` in `hip_early_flags` in `gdb/testsuite/lib/gdb.exp`, matching the old hipcc path.

Tested locally on MI300X (`gfx942`, `ROCR_VISIBLE_DEVICES=3`) with the WW28 TheRock tests tarball.

### DIFF
--- a/gdb/testsuite/lib/gdb.exp
+++ b/gdb/testsuite/lib/gdb.exp
@@ -6832,7 +6832,9 @@ proc gdb_compile {source dest type options} {
 	# before .cpp/.c sources and "-x none" before .o/.a files.
 	# --hip-link is still required when linking to bring in the HIP
 	# runtime.
-	set hip_early_flags "-mllvm=-amdgpu-spill-cfi-saved-regs\
+	# amdclang++ does not always default to -O0 for -x hip device compiles;
+	# omitting -O0 can emit DW_CC_nocall and break gdb.rocm finish tests.
+	set hip_early_flags "-O0 -mllvm=-amdgpu-spill-cfi-saved-regs\
 			     -Wno-unused-command-line-argument"
 	if {$type eq "executable"} {
 	    set hip_early_flags "--hip-link $hip_early_flags"


lumachad:
@lancesix @palves Just a FYI, as we've discussed this recently. It isn't clear if making this -O0 come back is what we really want.

lancesix:
This was removed because the we assumed that naked amdclang would default to -O0 (contrary to hipcc which defaulted to -02).  It is odd that the default is not -O0, but if it is not, then we need to spell it explicitly.

I would be really interested in getting more information regarding "amdclang++ does not always default to -O0 for -x hip device compiles".  Does that imply that for other languages it does?  Why is hip any different?  Note that on one target, we saw other consequences of not having -O0 where the compiler would inline some functions (and not alloc scratch) which we did not expect.

Also, all ROCgdb contributions should be ready to be sent upstream.  At the moment, it is unclear what is the policy regarding AI generated content.  I expect this to be small enough (and we have other similar patches in the pipeline around that upstream) so I do not think this particular patch will eventually be different from upstream.

palves:
Agreed with @lancesix .  We need to understand if this compiler behavior is really intended.  Having the compiler not apply optimizations by default is the decades-long established practice for clang (likely inherited from gcc, which itself probably inherited it from earlier proprietary unix compilers), and it's strange that amdclang++ would behave any different.  Off hand, this looks to me like a compiler regression.

palves:
> Without -O0, device helper functions can be emitted with DW_AT_calling_convention (DW_CC_nocall).

I'd guess this is the result of some optimization pass running (even though optimization wasn't requested), which then really affected the function's calling convention, rather than just a decision to emit the attribute.

ronlieb:
@jhuber6  could you drill into the -O0 vs omitted for amdclang ? shouldnt -O0 and omit be the same ?
amdclang++ -x hip without -O0 → DW_CC_nocall on helpers
amdclang++ -x hip -O0 → no DW_CC_nocall

jhuber6:
The current default path is that you compile with `-O0` which puts `optnone` and `noinline` on everything. LTO then handles the linking + backend stage, which runs at `lto<O2>` by default, but due to the `optnone` attribute most passes are skipped. I believe this is intended behavior.

> optnone
>
> This function attribute indicates that most optimization passes will skip this function, with the exception of interprocedural optimization passes. Code generation defaults to the “fast” instruction selector. This attribute cannot be used together with the alwaysinline attribute; this attribute is also incompatible with the minsize, optsize, and optdebug attributes.

Now, the issue is caused by the fact that we leave both `clang` and `ld.lld` as the default. If this is not the desired behavior the solution is most likely to have `clang` forward its optimization level even when it is not passed.

lfmeadow:
I'm closing this PR. 
I missed the "small" point that -g was being used. I absolutely agree that -g in the absence of any -O flags has historically implied -O0.
Using amdclang++ for pure device code compilation (no -x hip) honors this behavior.
My opinion is that -x hip is broken. It sounds like hipcc was always broken. I don't understand why this was never brought up.

What I found fascinating about this entire discussion is that nobody ever mentioned "-g implies -O0 without other optimization flags". It was just "amdclang++ compiles at -O0 by default" which is certainly not true.

Anyway, we know what the problem is.

PR #235, opened 2026-07-27
gdb: Implement stop-on-solib-events for GPU code objects

amd-bfilipov:
# stop-on-solib-events support for GPU code objects

GDB's `set stop-on-solib-events 1` setting allows users to stop execution when shared libraries are loaded or unloaded, enabling inspection and breakpoint placement before library code executes. This feature works for CPU shared libraries but was not working for GPU code objects loaded by the AMD ROCm runtime.

This PR implements `stop-on-solib-events` support for GPU code objects, making GPU code object load/unload events behave consistently with CPU shared library events.

**Related:** AIROCGDB-589

---

## Root Cause

The `amd_dbgapi_target_breakpoint::check_status()` function unconditionally set `bs->stop = 0` and `bs->print_it = print_it_noop`, regardless of the `stop_on_solib_events` setting. This is in contrast to `internal_breakpoint::check_status()` for CPU shared libraries, which respects the setting.

---

### Commit 1: Implement stop-on-solib-events for GPU code objects

**Implementation:**

- Added `code_object_list_updated` flag to `amd_dbgapi_inferior_info` to track when `AMD_DBGAPI_EVENT_KIND_CODE_OBJECT_LIST_UPDATED` events occur
- Modified `check_status()` to check this flag after processing events and update `bs->stop`, `bs->print`, and `bs->print_it` based on `stop_on_solib_events`
- Set the flag in `process_one_event()` when `CODE_OBJECT_LIST_UPDATED` events are seen

**Custom message printing:**

- Implemented `print_it()` method in `amd_dbgapi_target_breakpoint` to print a custom message distinguishing GPU code object events from CPU shared library events
- Message format follows the same pattern as CPU events but uses "GPU code object event" instead of "shared library event"
- Lists loaded/unloaded code objects with `file://` or `memory://` URIs
- Added MI interface support with `object-kind` field set to `"gpu-code-object"` to allow IDE frontends to distinguish GPU from CPU solib events

**Testing:**

- Added `gdb.rocm/gpu-solib-event.exp` test to verify GPU code object events trigger stops
- Test uses `hipModuleLoad()` for `file://` events and `hipModuleLoadData()` for `memory://` events
- Tests both load and unload events for both URI types

### Commit 2: Add GPU/CPU control to stop-on-solib-events setting

**Implementation:**

Modified `stop-on-solib-events` to accept integer values that control stopping on CPU shared libraries vs GPU code objects independently:
- **0** = Do not stop on shared library events
- **1** = Stop on CPU shared library events only (backward compatible - previous behavior)
- **2** = Stop on GPU code object events only
- **3** = Stop on both CPU and GPU library events

This follows GDB's precedent for integer-encoded mode settings (like `annotation_level`) and preserves backward compatibility - users who previously set `stop-on-solib-events` to 1 continue to get CPU-only behavior.

**User experience improvements:**

- The `show` command now displays a human-readable explanation of the current value:
  ```
  (gdb) show stop-on-solib-events
  Stopping for shared library events is 1 (CPU libraries only).
  ```
- Updated help text documents all four values clearly

**Test changes:**

- Updated `gpu-solib-event.exp` test to use value 2 (GPU events only)
- This eliminates the need to skip CPU shared library events during test execution
- CPU solib event testing is already covered by existing GDB testsuite tests

---

## Example Output

With the fix applied:

```
(gdb) set stop-on-solib-events 2
(gdb) run
Starting program: /path/to/simple
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Stopped due to GPU code object event:
  Inferior loaded file:///path/to/simple.co#offset=4096&size=63200
(gdb) c
Continuing.
Stopped due to GPU code object event:
  Inferior loaded memory://2785254#offset=0x457e30&size=43616
(gdb) c
Continuing.
result is 3
Stopped due to GPU code object event:
  Inferior unloaded file:///path/to/simple.co#offset=4096&size=63200
(gdb) c
Continuing.
Stopped due to GPU code object event:
  Inferior unloaded memory://2785254#offset=0x457e30&size=43616
(gdb) c
Continuing.
[Inferior 1 (process 2785254) exited normally]
```

**Note:** With value 2, CPU shared library events are skipped entirely, showing only GPU code object events.

---

## Backward Compatibility

- Setting `stop-on-solib-events` to 1 (or any non-zero value in the old behavior) continues to work for CPU libraries
- GPU code object events now require value 2 or 3 to stop
- Default value remains 0 (disabled)

## Benefits

1. **Consistent debugging experience** - GPU code objects behave like CPU shared libraries
2. **Independent control** - Users can choose to stop on CPU events only, GPU events only, or both
3. **IDE integration** - MI interface `object-kind` field allows programmatic distinction
4. **Improved test isolation** - Tests can focus on GPU events without CPU noise


lumachad:
Does this need a review? If so, we need to flip it to review.

amd-bfilipov:
> Does this need a review? If so, we need to flip it to review.

Wanted to check some things first, it's ready for review now.

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,92 @@
...
+if {[build_executable "failed to prepare" $testfile $srcfile {debug hip}]} {
+    return
+}
+
+proc do_test {} {
+    clean_restart
+    gdb_load $::binfile
+
+    with_rocm_gpu_lock {
+	if {![runto_main]} {
```
czidev-amd, line 31: It seems the spaces are not ok.

gdb/amd-dbgapi-target.c
```diff
@@ -910,6 +921,69 @@ amd_dbgapi_target_breakpoint::check_status (struct bpstat *bs)
...
+
+  /* If a CODE_OBJECT_LIST_UPDATED event was seen during event processing,
+     check if the user requested to stop on solib events.  This implements
+     stop-on-solib-events for GPU code objects.  */
+  if (info.code_object_list_updated)
+    {
+      bs->stop = stop_on_solib_events != 0;
+      bs->print = stop_on_solib_events != 0;
+      /* Set print_it to normal so that print_it() method is called.  */
+      bs->print_it = print_it_normal;
```
czidev-amd, line 934: I think gdb calls print_it when bs->print is true, what is it happening when bs->print is false? maybe guard bs->print_it assign with stop_on_solib_events?

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,92 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test that "set stop-on-solib-events 1" works for GPU code objects.
+
+load_lib rocm.exp
+
+require allow_hip_tests
+
+standard_testfile simple.cpp
```
czidev-amd: This is using a shared resource, not a blocker, but it may be wiser to have a dedicated source.

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,92 @@
...
+		-re "Stopped due to shared library event.*\r\n$::gdb_prompt $" {
+		    # CPU shared library event, keep going.
+		    pass $test
+		}
+		-re "Inferior .* exited normally.*\r\n$::gdb_prompt $" {
+		    # Program ended.
+		    pass $test
+		    break
+		}
+		-re ".*\r\n$::gdb_prompt $" {
```
czidev-amd: I think you can use `-re -warp ""`

czidev-amd (changes requested):
IMHO, you need to check `bs->print_it` assign. Any other comment is non blocking.

lumachad (commented):
Thanks for the patch! The overall approach is solid and mirrors how internal_breakpoint::check_status handles CPU solib events. A few things to address before merging.

gdb/amd-dbgapi-target.c
```diff
@@ -882,6 +890,9 @@ amd_dbgapi_target_breakpoint::check_status (struct bpstat *bs)
 
   require_forward_progress (info, false);
 
+  /* Reset the code_object_list_updated flag before processing events.  */
+  info.code_object_list_updated = false;
```
lumachad, line 895: The comment describes the what but not the why. Something like "Clear the flag so event processing below can detect a new CODE_OBJECT_LIST_UPDATED event." would be clearer about the intent.

gdb/amd-dbgapi-target.c
lumachad, line 934: "Set print_it to normal so that print_it() method is called." explains the mechanism rather than the intent. A clearer wording would be "Allow print_it() to print the GPU code object event message."

gdb/amd-dbgapi-target.c
```diff
@@ -910,6 +921,69 @@ amd_dbgapi_target_breakpoint::check_status (struct bpstat *bs)
...
+{
+  /* Check if this is a GPU code object event by looking at the solib lists.
+     We only reach here if check_status() set bs->print_it to print_it_normal,
+     which only happens for GPU code object events when stop_on_solib_events
+     is enabled.  */
+  bool any_deleted = !current_program_space->deleted_solibs.empty ();
+  bool any_added = !current_program_space->added_solibs.empty ();
+
+  if (any_added || any_deleted)
+    current_uiout->text (_("Stopped due to GPU code object event:\n"));
```
lumachad, line 953: The CPU counterpart `print_solib_event()` emits an MI async-reason field when `is_mi_like_p()` is true (see `breakpoint.c:5264`). This function omits that, so IDE clients using `--interpreter=mi` will not receive a notification for GPU code object events. Please add the same block:

```cpp
if (current_uiout->is_mi_like_p ())
  current_uiout->field_string ("reason",
    async_reason_lookup (EXEC_ASYNC_SOLIB_EVENT));
```

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,92 @@
...
+	# We should see at least one "Stopped due to GPU code object event"
+	# message.
+	set seen_gpu_solib_event 0
+	set seen_code_object_loaded 0
+
+	# Continue until we see GPU code object events or the program ends.
+	for {set i 0} {$i < 20} {incr i} {
+	    set test "continue and check for GPU code object event"
+	    gdb_test_multiple "continue" $test {
+		-re "Stopped due to GPU code object event.*Inferior loaded.*\r\n$::gdb_prompt $" {
```
lumachad: The test name is a fixed string reused for all 20 iterations of the loop. DejaGnu records each pass and fail by name, so duplicates make the log ambiguous and can hide regressions. Appending `$i` to the name would fix this, e.g. `"continue and check for GPU code object event (iter $i)"`.

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,92 @@
...
+		    pass $test
+		}
+		-re "Inferior .* exited normally.*\r\n$::gdb_prompt $" {
+		    # Program ended.
+		    pass $test
+		    break
+		}
+		-re ".*\r\n$::gdb_prompt $" {
+		    # Some other stop reason, keep going.
+		    pass $test
```
lumachad: The catch-all pattern calls `pass` for any unrecognised output, including unexpected errors or unknown stop reasons. That means a broken stop reason would be silently recorded as a pass. Consider using `fail` here instead, or at least adding a `timeout` arm.

lumachad (changes requested):

lumachad:
@amd-bfilipov When force-pushing, please make a comment clarifying what it is that you changed. Or go through the review comments and clarify for each of those what you've done.

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: created new file

gdb/amd-dbgapi-target.c
amd-bfilipov: Fixed. Changed comment to: "Allow print_it () to print the GPU code object event message."

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: Fixed. Changed test names to be unique per iteration AND per result type using $gdb_test_name

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: Fixed. Removed the catch-all pattern entirely and added explicit timeout arm that fails

amd-bfilipov:
> @amd-bfilipov When force-pushing, please make a comment clarifying what it is that you changed. Or go through the review comments and clarify for each of those what you've done.

@lumachad Understood. Replied to comments.

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,91 @@
...
+
+standard_testfile .cpp
+
+if {[build_executable "failed to prepare" $testfile $srcfile {debug hip}]} {
+    return
+}
+
+proc do_test {} {
+    clean_restart
+    gdb_load $::binfile
```
aktemur: Nit: `clean_restart $::testfile`. Actually, we don't even need this if we use above `prepare_for_testing` instead of `build_executable`.

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,91 @@
...
+		}
+	    }
+	}
+
+	# Verify we saw at least one GPU code object event.
+	if {$seen_gpu_solib_event == 0} {
+	    fail "did not see any GPU code object events"
+	} else {
+	    pass "saw GPU code object events"
+	}
```
aktemur: We can do `gdb_assert {$seen_gpu_solib_event}` instead of the whole if-else.

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,91 @@
...
+
+	# Verify we saw at least one code object load event.
+	if {$seen_code_object_loaded == 0} {
+	    fail "did not see GPU code object load event"
+	} else {
+	    pass "saw GPU code object load event"
+	}
```
aktemur: `gdb_assert` here, too.

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,91 @@
...
+    clean_restart
+    gdb_load $::binfile
+
+    with_rocm_gpu_lock {
+	if {![runto_main]} {
+	    return
+	}
+
+	# Enable stop-on-solib-events.
+	gdb_test_no_output "set stop-on-solib-events 1"
```
aktemur: There will be lots of CPU-side shared lib events until we reach the GPU event. GPU code object is loaded later. I think we can put a breakpoint just before the kernel invocation and then set `stop-on-solib-events`. Like this:
```
(gdb) b 37
Breakpoint 2 at 0x211930: file ...
(gdb) r
...
Thread 1 "solib-event" hit Breakpoint 2, main () at ...
37        do_an_addition<<<dim3(1), dim3(1), 0, 0>>> (1, 2, result_ptr);
(gdb) set stop-on-solib-events 1
(gdb) c
Continuing.
Stopped due to GPU code object event:
  Inferior loaded file:///...event#offset=8192&size=58744
(gdb)
```
Let's check with @lancesix that this is a reliable approach. We don't need then the loop below.

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,91 @@
...
+
+	# Continue and expect to stop when GPU code objects are loaded.
+	# We should see at least one "Stopped due to GPU code object event"
+	# message.
+	set seen_gpu_solib_event 0
+	set seen_code_object_loaded 0
+
+	# Continue until we see GPU code object events or the program ends.
+	for {set i 0} {$i < 20} {incr i} {
```
aktemur: I don't think we need this loop. See the comment above.

gdb/testsuite/gdb.rocm/solib-event.cpp
```diff
@@ -0,0 +1,48 @@
...
+}
+
+int
+main ()
+{
+  int *result_ptr, result;
+
+  /* Allocate memory for the device to write the result to.  */
+  hipError_t error = hipMalloc (&result_ptr, sizeof (int));
+  assert (error == hipSuccess);
```
aktemur: Do we even need these? Something as simple as 
```
kernel<<<1, 1>>> ();
```
where kernel is just
```
__global void
kernel ()
{
  NOP (1);
}
```

should suffice.

gdb/amd-dbgapi-target.c
```diff
@@ -910,6 +922,74 @@ amd_dbgapi_target_breakpoint::check_status (struct bpstat *bs)
...
+     check if the user requested to stop on solib events.  This implements
+     stop-on-solib-events for GPU code objects.  */
+  if (info.code_object_list_updated)
+    {
+      bs->stop = stop_on_solib_events != 0;
+      bs->print = stop_on_solib_events != 0;
+      /* Allow print_it () to print the GPU code object event message.  */
+      if (stop_on_solib_events != 0)
+	bs->print_it = print_it_normal;
+    }
```
aktemur, line 935: IMHO, this would read cleaner:
```
  if (info.code_object_list_updated && stop_on_solib_events != 0)
    {
      bs->stop = true;
      bs->print = true;
      /* Allow print_it () to print the GPU code object event message.  */
      bs->print_it = print_it_normal;
    }
```

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: Used the loop-based approach instead of breakpoint-before-kernel because the suggested approach had issues with CPU solib events still appearing. Even when I put breakpoint on kernel call line, I still get  cpu load event inbetween.

gdb/testsuite/gdb.rocm/solib-event.cpp
amd-bfilipov: ok, simplified it

gdb/amd-dbgapi-target.c
amd-bfilipov: done

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: Replaced if-else block with gdb_assert {$seen_file_prefix}.

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: Replaced if-else block with gdb_assert {$seen_gpu_solib_event}

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: Replaced build_executable with prepare_for_testing which handles both build and load in one step

gdb/testsuite/gdb.rocm/solib-event.cpp
```diff
@@ -0,0 +1,48 @@
...
+    if (error != hipSuccess)                                                 \
+      {                                                                      \
+	fprintf (stderr, "error: '%s'(%d) at %s:%d\n",                       \
+		 hipGetErrorString (error), error, __FILE__, __LINE__);      \
+	exit (EXIT_FAILURE);                                                 \
+      }                                                                      \
+  }
+
+/* Prevent inlining to ensure kernel stays visible for debugging.  */
+#define NOP(x) asm ("s_nop " #x)
```
aktemur: Let's please include `rocm-test-utils.h`. We then don't need to define `CHECK` and `NOP`.

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,72 @@
...
+
+require allow_hip_tests
+
+standard_testfile .cpp
+
+if {[prepare_for_testing "failed to prepare" $testfile $srcfile {debug hip}]} {
+    return
+}
+
+proc do_test {} {
```
aktemur: IMHO we don't need this proc, but this style is used in some other tests, so I wouldn't require that you remove it.

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,72 @@
...
+
+	# Continue through solib events until we hit the GPU code object event.
+	# There may be CPU shared library events before the GPU event.
+	set seen_gpu_solib_event 0
+	set seen_file_prefix 0
+	set max_continues 20
+
+	for {set i 0} {$i < $max_continues} {incr i} {
+	    gdb_test_multiple "continue" "continue looking for GPU event" {
+		-re "Stopped due to GPU code object event.*Inferior loaded (file://\[^\r\n\]+).*$::gdb_prompt $" {
```
aktemur: Please use `-wrap` and remove the prompt anchor at the end. Also applies to the cases below.

gdb/amd-dbgapi-target.c
```diff
@@ -910,6 +922,73 @@ amd_dbgapi_target_breakpoint::check_status (struct bpstat *bs)
...
+
+  if (current_uiout->is_mi_like_p ())
+    current_uiout->field_string ("reason",
+				 async_reason_lookup (EXEC_ASYNC_SOLIB_EVENT));
+
+  if (any_added || any_deleted)
+    current_uiout->text (_("Stopped due to GPU code object event:\n"));
+  else
+    current_uiout->text (_("Stopped due to GPU code object event (no "
+			   "code objects added or removed)\n"));
```
lumachad, line 962: Do you know if/when this happens? Why would we have a code object event where code objects haven't been added or removed?

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,72 @@
...
+		    set seen_gpu_solib_event 1
+		    set seen_file_prefix 1
+		    break
+		}
+		-re "Stopped due to shared library event.*$::gdb_prompt $" {
+		    # CPU solib event, continue looping.
+		}
+		-re "Inferior.*exited.*$::gdb_prompt $" {
+		    fail "program exited before GPU event"
```
aktemur: We don't need this fail. There is no "pass" correspondence. The asserts below will emit FAILs anyway.

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,72 @@
...
+		    break
+		}
+	    }
+
+	    if {$seen_gpu_solib_event} {
+		break
+	    }
```
aktemur: This seems unneeded. We already have a `break` at the single place where we set the boolean to true.

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	# Verify we saw the GPU code object event.
+	gdb_assert {$seen_gpu_solib_event} "saw GPU code object event"
+
+	# Verify the loaded code object has the file:// prefix.
+	gdb_assert {$seen_file_prefix} "code object has file:// prefix"
```
aktemur: These two flags are either both false or both true. There is only one case that sets both. We should either separate the scenarios (is there a scenario that emits a GPU event but no file prefix?) or only one flag would suffice.

lumachad (commented):
This seems reasonable. But I wonder if whoever is chasing GPU bugs and wants to see only GPU solib/code object events would be best server by a switch that only stops for GPU code object events rather than all the non-interesting CPU solib events.

aktemur:
> This seems reasonable. But I wonder if whoever is chasing GPU bugs and wants to see only GPU solib/code object events would be best server by a switch that only stops for GPU code object events rather than all the non-interesting CPU solib events.

I'd think that the user gets frustrated by the CPU solib events in the first try, and then they quickly learn that they should rather turn the setting on at a later point after CPU libs are loaded.

I wouldn't mind a separate "stop-on-gpu-solib-events" or a similar setting. It sounds useful.

lumachad (commented):
I have a few more automated comments, will post shortly.

gdb/amd-dbgapi-target.c
```diff
@@ -284,6 +286,11 @@ struct amd_dbgapi_inferior_info
   /* List of pending events the amd-dbgapi target retrieved from the dbgapi.  */
   std::list<std::pair<ptid_t, target_waitstatus>> wave_events;
 
+  /* Flag to track if a CODE_OBJECT_LIST_UPDATED event was seen during
+     process_event_queue.  Used to implement stop-on-solib-events for GPU
+     code objects.  */
+  bool code_object_list_updated = false;
+
```
lumachad, line 293: Nit: the comment does not mention that this flag is reset at the top of `check_status` before each use. A brief note would help future readers understand the intended lifetime.

gdb/amd-dbgapi-target.c
```diff
@@ -910,6 +922,73 @@ amd_dbgapi_target_breakpoint::check_status (struct bpstat *bs)
...
+      /* Allow print_it () to print the GPU code object event message.  */
+      bs->print_it = print_it_normal;
+    }
+}
+
+enum print_stop_action
+amd_dbgapi_target_breakpoint::print_it (const bpstat *bs) const
+{
+  /* Check if this is a GPU code object event by looking at the solib lists.
+     We only reach here if check_status () set bs->print_it to print_it_normal,
```
lumachad: The comment contradicts itself: the first sentence says "Check if this is a GPU code object event" but the rest establishes we already know it is. Suggested rewording:
```c
/* We only reach here when check_status set bs->print_it to print_it_normal,
   which happens only for GPU code object events when stop_on_solib_events
   is enabled.  */
```

gdb/amd-dbgapi-target.c
```diff
@@ -910,6 +922,73 @@ amd_dbgapi_target_breakpoint::check_status (struct bpstat *bs)
...
+     which only happens for GPU code object events when stop_on_solib_events
+     is enabled.  */
+  bool any_deleted = !current_program_space->deleted_solibs.empty ();
+  bool any_added = !current_program_space->added_solibs.empty ();
+
+  if (current_uiout->is_mi_like_p ())
+    current_uiout->field_string ("reason",
+				 async_reason_lookup (EXEC_ASYNC_SOLIB_EVENT));
+
```
lumachad, line 957: Both GPU code-object events and CPU shared-library events emit `reason="solib-event"` over MI. An MI front-end has no way to distinguish them without parsing free-form text. Consider a new async reason or an extra field such as `"object-kind", "gpu-code-object"` to allow programmatic differentiation.

gdb/amd-dbgapi-target.c
```diff
@@ -910,6 +922,73 @@ amd_dbgapi_target_breakpoint::check_status (struct bpstat *bs)
...
+  else
+    current_uiout->text (_("Stopped due to GPU code object event (no "
+			   "code objects added or removed)\n"));
+
+  if (any_deleted)
+    {
+      current_uiout->text (_("  Inferior unloaded "));
+      ui_out_emit_list list_emitter (current_uiout, "removed");
+      for (int ix = 0;
+	   ix < current_program_space->deleted_solibs.size ();
```
lumachad: The `deleted_solibs` block uses an index-based `for` loop while the `added_solibs` block below uses a range-based loop. Please pick one style and apply it to both.

gdb/testsuite/gdb.rocm/solib-event.cpp
```diff
@@ -0,0 +1,48 @@
...
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.  */
+
+#include <hip/hip_runtime.h>
+#include <cstdlib>
+#include <stdio.h>
+
```
lumachad: Nit: mixed C and C++ headers — `<cstdlib>` (C++) alongside `<stdio.h>` (C). Please use `<cstdio>` instead of `<stdio.h>` for consistency, or drop `<cstdlib>` if `EXIT_FAILURE` is already provided by the HIP header.

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	gdb_test_no_output "set stop-on-solib-events 1"
+
+	# Continue through solib events until we hit the GPU code object event.
+	# There may be CPU shared library events before the GPU event.
+	set seen_gpu_solib_event 0
+	set seen_file_prefix 0
+	set max_continues 20
+
```
lumachad: The magic number `20` has no explanation. Please add a comment saying why 20 iterations is a sufficient upper bound to cover all CPU solib events before the first GPU event.

gdb/testsuite/gdb.rocm/solib-event.exp
lumachad: `gdb_test_multiple` is missing a catch-all prompt arm. If GDB stops for an unexpected reason the call returns without setting any flag and the loop silently continues. Consider adding:
```tcl
-re "$::gdb_prompt $" {
    fail "unexpected stop before GPU code object event"
    break
}
```

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,72 @@
...
+	for {set i 0} {$i < $max_continues} {incr i} {
+	    gdb_test_multiple "continue" "continue looking for GPU event" {
+		-re "Stopped due to GPU code object event.*Inferior loaded (file://\[^\r\n\]+).*$::gdb_prompt $" {
+		    set seen_gpu_solib_event 1
```
lumachad: The test only verifies `file://` code object events. The PR description shows both `file://` and `memory://` events occur. Consider also checking that `memory://` events are printed correctly.

aktemur:
This is a user-visible feature. I think it's worth adding a CHANGELOG entry so that it is mentioned in the release notes.

gdb/testsuite/gdb.rocm/solib-event.cpp
```diff
@@ -0,0 +1,48 @@
+/* This testcase is part of GDB, the GNU debugger.
+
+   Copyright 2022-2026 Free Software Foundation, Inc.
```
lancesix: why do we start from 2022?

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,72 @@
...
+proc do_test {} {
+    with_rocm_gpu_lock {
+	if {![runto_main]} {
+	    return
+	}
+
+	gdb_test_no_output "set stop-on-solib-events 1"
+
+	# Continue through solib events until we hit the GPU code object event.
+	# There may be CPU shared library events before the GPU event.
```
lancesix: I am not sure the approach is the best one.  For example, if someone runs in an environment where the environment variable `HIP_ENABLE_DEFERRED_LOADING=0` all GPU code objects will be loaded before you reach main.

A better way would be to use the `hipModule*` series of functions to explicitly load a GPU code object at runtime.  The testcase could run until the load, step over (or continue over) the load and check that we receive the GPU code object loading.  This way, we control everything, which should ensure we have just one event coming.

You can check testcases such as gdb.rocm/code-object-load-while-breakpoint-hip or gdb.rocm/snapshot-objfile-on-load for examples.

aktemur: > A better way would be to use the `hipModule*` series of functions to explicitly load a GPU code object at runtime. The testcase could run until the load, step over (or continue over) the load and check that we receive the GPU code object loading. This way, we control everything, which should ensure we have just one event coming.

This sounds very good to me.

gdb/amd-dbgapi-target.c
amd-bfilipov: I don't know if it happens for GPU events but while testing I've seen "no code objects added or removed" events on CPU.

gdb/testsuite/gdb.rocm/solib-event.cpp
amd-bfilipov: Replaced custom CHECK and NOP macros with #include "rocm-test-utils.h" which provides the same definitions

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: Added -wrap to all gdb_test_multiple patterns and removed explicit $::gdb_prompt $ anchors

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: kept it as it is used in other tests

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: Removed

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: Removed

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: Merged seen_gpu_solib_event and seen_file_prefix into single flag

gdb/amd-dbgapi-target.c
amd-bfilipov: done

gdb/amd-dbgapi-target.c
amd-bfilipov: Added object-kind='gpu-code-object' field to MI output at amd-dbgapi-target.c:953, allowing MI frontends to distinguish GPU from CPU solib events programmatically

gdb/amd-dbgapi-target.c
amd-bfilipov: Converted deleted_solibs to range-based loop

gdb/testsuite/gdb.rocm/solib-event.cpp
amd-bfilipov: Removed all manual headers since rocm-test-utils.h includes both stdio.h and stdlib.h

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: Added comment explaining that the limit of 20 iterations should be sufficient to cover all CPU shared library events that load before the first GPU code object

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: Added catch-all pattern with -wrap and fixed CPU solib pattern to include .* at end to match optional text like '(no libraries added or removed)'

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: Completely rewrote test using hipModule* API to test both file:// (hipModuleLoad) and memory:// (hipModuleLoadData) events, as well as unload events for both

gdb/testsuite/gdb.rocm/solib-event.cpp
amd-bfilipov: fixed

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: Completely rewrote test using hipModule* API to test both file:// (hipModuleLoad) and memory:// (hipModuleLoadData) events, as well as unload events for both

amd-bfilipov:
Added independent CPU/GPU control (Commit 2)

  Modified stop-on-solib-events to accept integer values for independent control:
  - 0 = disabled (no stops)
  - 1 = CPU libraries only (backward compatible - preserves previous behavior)
  - 2 = GPU code objects only
  - 3 = both CPU and GPU

  This follows GDB's precedent for integer-encoded settings (like annotation_level). The show command now displays
  human-readable explanations:

  (gdb) show stop-on-solib-events
  Stopping for shared library events is 2 (GPU code objects only).

  Updated the test to use value 2, eliminating the need to filter out CPU shared library events during execution.
  CPU solib testing is already covered by existing GDB testsuite tests.

lumachad:

## Review summary
The fix correctly wires GPU code-object load/unload events into the existing `stop-on-solib-events` machinery, mirroring the pattern already used by `internal_breakpoint::check_status()` for CPU shared libraries, and is backed by a reasonable test. The main issue is that `amd_dbgapi_target_breakpoint::print_it()` duplicates `print_solib_event()` instead of extending/reusing it, and the change alters user-visible behavior/output without any doc or NEWS update.

## Bugs
- `gdb/testsuite/gdb.rocm/solib-event.exp:348-350` — the catch-all `-re -wrap ".*" { fail "unexpected stop before $test_name"; break }` and the `Inferior.*exited` branch both `break` out of the retry loop while leaving `seen_gpu_event` at 0, so `gdb_assert {$seen_gpu_event} $test_name` right after fires again, producing two FAIL entries for a single failure. Consider `return`ing from the proc (or setting a distinct message and skipping the trailing `gdb_assert`) in those two branches so each real failure is reported once.

## GNU coding standards
- `gdb/amd-dbgapi-target.c:940-993` (`amd_dbgapi_target_breakpoint::print_it`) — duplicates the logic of `print_solib_event()` in `gdb/breakpoint.c` (same text/list-emission structure, only the message string and field name `"code-object"` vs `"library"` differ, plus the extra `object-kind` MI field). Consider generalizing `print_solib_event()` to take the event text and field name as parameters, rather than reimplementing the added/deleted-solib printing loop a second time. This also means any future fix to the CPU path won't automatically apply to the GPU path.

## Cosmetic / style
- `gdb/testsuite/gdb.rocm/solib-event.exp:58,61,68,104,107,112,115` — several new lines exceed the 80-column limit. Suggested wraps:
  - Line 58:
    ```tcl
    if {[string equal $event_type "load"]
        && [string equal $expect_out(1,string) "loaded"]} {
    ```
  - Line 61:
    ```tcl
    } elseif {[string equal $event_type "unload"]
              && [string equal $expect_out(1,string) "unloaded"]} {
    ```
  - Line 68 (comment):
    ```tcl
    # GPU event with different URI type (e.g., memory:// when
    # looking for file://), continue.
    ```
  - Lines 104/107/112/115:
    ```tcl
    find_gpu_event "load" "file://" \
        "saw GPU code object load event with file:// prefix"

    find_gpu_event "unload" "file://" \
        "saw GPU code object unload event with file:// prefix"

    find_gpu_event "load" "memory://" \
        "saw GPU code object load event with memory:// prefix"

    find_gpu_event "unload" "memory://" \
        "saw GPU code object unload event with memory:// prefix"
    ```
- `gdb/amd-dbgapi-target.c:949-954` vs `gdb/breakpoint.c` (`print_solib_event`) — field emission order differs from the reference implementation (`reason`/`object-kind` emitted before the text vs. text-then-`reason`). Not a bug, but worth aligning for consistency since this is meant to parallel the CPU path.

## Tests
- No test exercises `stop-on-solib-events 0` (default/off) still suppressing GPU code-object stops, i.e. no regression coverage confirming the old "never stop" behavior is preserved when the setting is disabled.
- `gdb/testsuite/gdb.rocm/solib-event.exp:371` sets `stop-on-solib-events 1`, so CPU shared-library events also trigger stops during the test; the `find_gpu_event` helper tolerates/skips these via a hardcoded `max_continues 20` heuristic, which is somewhat fragile.

## Nits / optional
- `gdb/NEWS` and `gdb/doc/gdb.texinfo` (`@item set stop-on-solib-events`) are not updated even though this changes user-visible behavior (GPU events now also honor the setting, and a new "Stopped due to GPU code object event" message/`object-kind` MI field is introduced). Worth a NEWS entry and a doc note.
- `gdb/amd-dbgapi-target.c:960` — the "(no code objects added or removed)" fallback message path is untested.

lumachad:
@amd-bfilipov Please also update the PR description to discard stale information.

gdb/testsuite/gdb.rocm/solib-event-kernel.cpp
aktemur, line 1: Nitpicking in the commit message: 
Using present tense is more usual. That is, instead of "...setting was not working...", we can write "...setting is not working..." The general state of mind is "there is currently a problem; we fix the problem by applying this patch". 

Also, dot-space-space in the commit message too. E.g. "... AMD ROCm runtime. While CPU ..."
There are a few more instances.

gdb/testsuite/gdb.rocm/solib-event.cpp
```diff
@@ -0,0 +1,92 @@
...
+  std::ifstream mod (module_path, std::ios::binary | std::ios::ate);
+  if (!mod.is_open ())
+    {
+      fprintf (stderr, "Failed to open module file\n");
+      exit (EXIT_FAILURE);
+    }
+
+  size_t module_size = mod.tellg ();
+  mod.seekg (0, std::ios::beg);
+  std::vector<char> module_buffer (module_size);
```
aktemur, line 54: Nit: `unsigned char`.

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,119 @@
...
+	}
+    }
+
+    gdb_assert {$seen_gpu_event} $test_name
+}
+
+proc do_test {} {
+    with_rocm_gpu_lock {
+	clean_restart
+	gdb_load $::binfile
```
aktemur, line 93: `clean_restart $::testfile`

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,119 @@
...
+    set max_continues 20
+
+    # The limit of 20 iterations should be sufficient to cover all CPU
+    # shared library events that may occur during hipModule* operations.
+    # HIP runtime may also load internal memory:// code objects during
+    # hipModuleLoad, so we skip GPU events that don't match our criteria.
+    for {set i 0} {$i < $max_continues} {incr i} {
+	gdb_test_multiple "continue" "continue looking for $test_name" {
+	    -re -wrap "Stopped due to GPU code object event.*Inferior (loaded|unloaded) ${uri_pattern}\[^\r\n\]+" {
+		if {[string equal $event_type "load"] && [string equal $expect_out(1,string) "loaded"]} {
```
aktemur, line 58: Can we not use `$event_type eq "load"`?

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,119 @@
...
+	    }
+	    -re -wrap ".*" {
+		fail "unexpected stop before $test_name"
+		break
+	    }
+	}
+
+	if {$seen_gpu_event} {
+	    break
+	}
```
aktemur, line 84: Do we need this? There are two places where `seen_gpu_event` can become true and both cases are followed immediately with a `break`.

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,119 @@
...
+		# GPU event with different URI type (e.g., memory:// when looking for file://), continue.
+	    }
+	    -re -wrap "Stopped due to shared library event.*" {
+		# CPU solib event, continue looping.
+	    }
+	    -re -wrap "Inferior.*exited" {
+		break
+	    }
+	    -re -wrap ".*" {
+		fail "unexpected stop before $test_name"
```
aktemur, line 77: Do we need this `fail`? We are going to fail anyway because we haven't seen the GPU event.

gdb/testsuite/gdb.rocm/solib-event.exp
```diff
@@ -0,0 +1,119 @@
...
+#   uri_pattern: regex pattern to match (e.g., "file://" or "memory://")
+#   test_name: descriptive name for the test
+proc find_gpu_event {event_type uri_pattern test_name} {
+    set seen_gpu_event 0
+    set max_continues 20
+
+    # The limit of 20 iterations should be sufficient to cover all CPU
+    # shared library events that may occur during hipModule* operations.
+    # HIP runtime may also load internal memory:// code objects during
+    # hipModuleLoad, so we skip GPU events that don't match our criteria.
```
lancesix, line 54: wouldn't it easier to just launch a dummy dispatch first on the device, then start our test doing load / unload?

lancesix:
Here is a possible simplification apphoach which uses a warm up kernel to force the runtime to load any internal shader before we start our test.  This removes the need of "find_gpu_event", and the arbitrary limit of event number to ignore.

Also, this would probably qualify for a CHANGELOG entry.

```diff
diff --git a/gdb/testsuite/gdb.rocm/solib-event.cpp b/gdb/testsuite/gdb.rocm/solib-event.cpp
index 8a0922d3c01..671b1186f7a 100644
--- a/gdb/testsuite/gdb.rocm/solib-event.cpp
+++ b/gdb/testsuite/gdb.rocm/solib-event.cpp
@@ -74,6 +74,11 @@ test_memory_load (const char *module_path)
   CHECK (hipModuleUnload (module));
 }
 
+__global__ void
+warm_up ()
+{
+}
+
 int
 main (int argc, char **argv)
 {
@@ -85,6 +90,13 @@ main (int argc, char **argv)
 
   const char *module_path = argv[1];
 
+  /* Submit an empty kernel to force the runtime to do any necessary
+     setup which might include loading internal code objects.  */
+  warm_up<<<1, 1>>>();
+  CHECK (hipDeviceSynchronize ());
+
+  /* Enable SOLIB events here.  */
+
   test_file_load (module_path);
   test_memory_load (module_path);
 
diff --git a/gdb/testsuite/gdb.rocm/solib-event.exp b/gdb/testsuite/gdb.rocm/solib-event.exp
index 5618ac3292b..f04c50b877f 100644
--- a/gdb/testsuite/gdb.rocm/solib-event.exp
+++ b/gdb/testsuite/gdb.rocm/solib-event.exp
@@ -39,54 +39,6 @@ if {[build_executable "failed to prepare" $testfile $srcfile {debug hip}] == -1}
     return
 }
 
-# Helper proc to skip CPU solib events and find GPU events.
-# Arguments:
-#   event_type: "load" or "unload"
-#   uri_pattern: regex pattern to match (e.g., "file://" or "memory://")
-#   test_name: descriptive name for the test
-proc find_gpu_event {event_type uri_pattern test_name} {
-    set seen_gpu_event 0
-    set max_continues 20
-
-    # The limit of 20 iterations should be sufficient to cover all CPU
-    # shared library events that may occur during hipModule* operations.
-    # HIP runtime may also load internal memory:// code objects during
-    # hipModuleLoad, so we skip GPU events that don't match our criteria.
-    for {set i 0} {$i < $max_continues} {incr i} {
-       gdb_test_multiple "continue" "continue looking for $test_name" {
-           -re -wrap "Stopped due to GPU code object event.*Inferior (loaded|unloaded) ${uri_pattern}\[^\r\n\]+" {
-               if {[string equal $event_type "load"] && [string equal $expect_out(1,string) "loaded"]} {
-                   set seen_gpu_event 1
-                   break
-               } elseif {[string equal $event_type "unload"] && [string equal $expect_out(1,string) "unloaded"]} {
-                   set seen_gpu_event 1
-                   break
-               }
-               # GPU event doesn't match our criteria (wrong operation type), continue.
-           }
-           -re -wrap "Stopped due to GPU code object event.*" {
-               # GPU event with different URI type (e.g., memory:// when looking for file://), continue.
-           }
-           -re -wrap "Stopped due to shared library event.*" {
-               # CPU solib event, continue looping.
-           }
-           -re -wrap "Inferior.*exited" {
-               break
-           }
-           -re -wrap ".*" {
-               fail "unexpected stop before $test_name"
-               break
-           }
-       }
-
-       if {$seen_gpu_event} {
-           break
-       }
-    }
-
-    gdb_assert {$seen_gpu_event} $test_name
-}
-
 proc do_test {} {
     with_rocm_gpu_lock {
        clean_restart
@@ -96,23 +48,40 @@ proc do_test {} {
            return
        }
 
+       gdb_breakpoint [gdb_get_line_number "Enable SOLIB events here"] -temporary
+       gdb_continue_to_breakpoint "at enable solib-event"
+
        gdb_test_no_output "set stop-on-solib-events 1"
 
-       # Test 1: file:// load event (hipModuleLoad).
-       gdb_breakpoint "test_file_load" -temporary
-       gdb_continue_to_breakpoint "at test_file_load"
-       find_gpu_event "load" "file://" "saw GPU code object load event with file:// prefix"
+       with_test_prefix "file://" {
+           gdb_breakpoint "test_file_load" -temporary
+           gdb_continue_to_breakpoint "at test_file_load"
+
+           # Test 1: file:// load event (hipModuleLoad).
+           gdb_test "continue" \
+               "Stopped due to GPU code object event.*Inferior loaded file://\[^\r\n\]+" \
+               "load"
 
-       # Test 2: file:// unload event (hipModuleUnload).
-       find_gpu_event "unload" "file://" "saw GPU code object unload event with file:// prefix"
+           # Test 2: file:// unload event (hipModuleUnload).
+           gdb_test "continue" \
+               "Stopped due to GPU code object event.*Inferior unloaded file://\[^\r\n\]+" \
+               "unload"
+       }
+
+       with_test_prefix "memory://" {
+           gdb_breakpoint "test_memory_load" -temporary
+           gdb_continue_to_breakpoint "at test_memory_load"
 
-       # Test 3: memory:// load event (hipModuleLoadData).
-       gdb_breakpoint "test_memory_load" -temporary
-       gdb_continue_to_breakpoint "at test_memory_load"
-       find_gpu_event "load" "memory://" "saw GPU code object load event with memory:// prefix"
+           # Test 3: memory:// load event (hipModuleLoadData).
+           gdb_test "continue" \
+               "Stopped due to GPU code object event.*Inferior loaded memory://\[^\r\n\]+" \
+               "load"
 
-       # Test 4: memory:// unload event (hipModuleUnload).
-       find_gpu_event "unload" "memory://" "saw GPU code object unload event with memory:// prefix"
+           # Test 4: memory:// unload event (hipModuleUnload).
+           gdb_test "continue" \
+               "Stopped due to GPU code object event.*Inferior unloaded memory://\[^\r\n\]+" \
+               "unload"
+       }
     }
 }
 
```

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: ye, seems like a good idea. Implemented it.

lancesix (commented):
Ovarall looks reasonable to me.  There are still comments from Baris which are still unresolved.

Could this patch apply upstream?  Feels it could be merged upstream first.

amd-bfilipov:
> Ovarall looks reasonable to me. There are still comments from Baris which are still unresolved.
> 
> Could this patch apply upstream? Feels it could be merged upstream first.

I'm fixing Luis' big automated comment first, and then what Baris added. I'll ping you when it's all complete.
Then we can try upstreaming it.

gdb/testsuite/gdb.rocm/solib-event-kernel.cpp
amd-bfilipov: fixed

gdb/testsuite/gdb.rocm/solib-event.cpp
amd-bfilipov: fixed

gdb/testsuite/gdb.rocm/solib-event.exp
amd-bfilipov: fixed

amd-bfilipov:
> Ovarall looks reasonable to me. There are still comments from Baris which are still unresolved.
> 
> Could this patch apply upstream? Feels it could be merged upstream first.

All Baris' and Luis' comments are fixed. There are two commits now, but if everything looks good I will squash them into a single commit and send it upstream.

lumachad:
gdb.rocm/solib-event.exp seems to be failing in https://github.com/ROCm/ROCgdb/actions/runs/35749024758/job/106823605336?pr=235.

lumachad:

## Review summary

The core fix (using `stop_on_solib_events` to control GPU code-object breakpoint stops, and generalizing `print_solib_event`) is sound and well-tested, but the PR description is significantly out of sync with the actual code, and there's a real user-visible message regression in the shared (CPU) path.

## Bugs

- `gdb/breakpoint.c:5261-5264` — Pluralization bug in the generalized "no libraries added" message. The new code does `string_printf (_("... no %ss added or removed)\n"), event_description, event_description)`, which for the default/CPU case (`event_description == "shared library"`) produces "no shared librarys added or removed" instead of the original, correct "no libraries added or removed". This is a regression in existing CPU-path user-visible output. Since irregular plurals can't be built by appending "s", the caller needs to pass the plural form explicitly (e.g. add a `plural_description` parameter, or pass a full pre-pluralized phrase) rather than deriving it mechanically.

- PR description vs. actual code mismatch: the PR body describes a "Commit 2: Add GPU/CPU control to stop-on-solib-events setting" that turns `stop-on-solib-events` into an integer-encoded 0-3 setting (0=none, 1=CPU only, 2=GPU only, 3=both), with an updated `show` message and help text. None of this exists in the actual diff. The real code (`gdb/amd-dbgapi-target.c:931`) simply does `if (info.code_object_list_updated && stop_on_solib_events != 0)`, reusing the existing boolean-ish semantics from `internal_breakpoint::check_status`. Consequences:
  - Setting `stop-on-solib-events 1` (previously CPU-only) will now also stop on GPU code-object events — the "backward compatible" claim in the PR description is false for the code actually submitted.
  - The test (`solib-event.exp:59,86`) uses `set stop-on-solib-events 1`, not `2` as the description claims.
  - `gdb/infrun.c:11035-11043` (`add_setshow_zinteger_cmd` for `stop-on-solib-events`) is untouched — the help/show text still only mentions "shared library events" with no reference to GPU code objects, so this behavior change is undocumented to users.

  This needs to be resolved before merge: either the PR description is stale (most likely, e.g. from a squashed/rewritten commit) and should be corrected, or the intended value-based CPU/GPU split was dropped unintentionally and should be re-added along with updated help text and a NEWS entry.

## Documentation

- `gdb/doc/gdb.texinfo` — not touched. The `stop-on-solib-events` section of the user manual still only documents the boolean/CPU-only behavior; it does not mention that GPU code-object events are now covered, nor any 0-3 value scheme (which, per the finding above, doesn't exist in the code anyway).
- `gdb/infrun.c:11035-11043` (`add_setshow_zinteger_cmd` help/show strings for `stop-on-solib-events`) — not touched. The in-tool help text still describes only the CPU dynamic-linker case, with no reference to GPU code objects, even though the setting's behavior now changes for GPU-loaded code.
- `gdb/NEWS` — not touched. A NEWS entry is the convention for documenting user-visible behavior changes to existing settings, and its absence should be flagged and fixed.

## Nits / optional

- `gdb/amd-dbgapi-target.c:892-895` — comment on the flag reset is slightly redundant with itself; consider trimming.
- `gdb/testsuite/gdb.rocm/solib-event.exp:39` and `solib-event.cpp:56` exceed 80 columns — minor GNU style nit, not blocking.

## Tests

- Test coverage for the actual implemented behavior (boolean stop on GPU events, load/unload, `file://`/`memory://` URIs) is solid.
- Given the description's claimed 0-3 value scheme isn't implemented, there's naturally no test for GPU-only (`2`) vs. combined (`3`) modes either — worth calling out so it isn't assumed to be covered.

---

Bottom line: the mechanical GPU-stop implementation and the `print_solib_event` refactor are reasonable and DRY, but (1) there's a real string bug that garbles the existing CPU message, and (2) neither the PR description's claimed CPU/GPU value scheme nor any of the documentation it implies (help text, manual, NEWS) actually landed in this diff — this should be resolved before merging so reviewers and future readers aren't misled about what `stop-on-solib-events` actually does after this change.

lumachad (commented):
Made some comments.

amd-bfilipov:
> 
> ## Review summary
> The core fix (using `stop_on_solib_events` to control GPU code-object breakpoint stops, and generalizing `print_solib_event`) is sound and well-tested, but the PR description is significantly out of sync with the actual code, and there's a real user-visible message regression in the shared (CPU) path.
> 
> ## Bugs
> * `gdb/breakpoint.c:5261-5264` — Pluralization bug in the generalized "no libraries added" message. The new code does `string_printf (_("... no %ss added or removed)\n"), event_description, event_description)`, which for the default/CPU case (`event_description == "shared library"`) produces "no shared librarys added or removed" instead of the original, correct "no libraries added or removed". This is a regression in existing CPU-path user-visible output. Since irregular plurals can't be built by appending "s", the caller needs to pass the plural form explicitly (e.g. add a `plural_description` parameter, or pass a full pre-pluralized phrase) rather than deriving it mechanically.
> * PR description vs. actual code mismatch: the PR body describes a "Commit 2: Add GPU/CPU control to stop-on-solib-events setting" that turns `stop-on-solib-events` into an integer-encoded 0-3 setting (0=none, 1=CPU only, 2=GPU only, 3=both), with an updated `show` message and help text. None of this exists in the actual diff. The real code (`gdb/amd-dbgapi-target.c:931`) simply does `if (info.code_object_list_updated && stop_on_solib_events != 0)`, reusing the existing boolean-ish semantics from `internal_breakpoint::check_status`. Consequences:
>   
>   * Setting `stop-on-solib-events 1` (previously CPU-only) will now also stop on GPU code-object events — the "backward compatible" claim in the PR description is false for the code actually submitted.
>   * The test (`solib-event.exp:59,86`) uses `set stop-on-solib-events 1`, not `2` as the description claims.
>   * `gdb/infrun.c:11035-11043` (`add_setshow_zinteger_cmd` for `stop-on-solib-events`) is untouched — the help/show text still only mentions "shared library events" with no reference to GPU code objects, so this behavior change is undocumented to users.
>   
>   This needs to be resolved before merge: either the PR description is stale (most likely, e.g. from a squashed/rewritten commit) and should be corrected, or the intended value-based CPU/GPU split was dropped unintentionally and should be re-added along with updated help text and a NEWS entry.
> 
> ## Documentation
> * `gdb/doc/gdb.texinfo` — not touched. The `stop-on-solib-events` section of the user manual still only documents the boolean/CPU-only behavior; it does not mention that GPU code-object events are now covered, nor any 0-3 value scheme (which, per the finding above, doesn't exist in the code anyway).
> * `gdb/infrun.c:11035-11043` (`add_setshow_zinteger_cmd` help/show strings for `stop-on-solib-events`) — not touched. The in-tool help text still describes only the CPU dynamic-linker case, with no reference to GPU code objects, even though the setting's behavior now changes for GPU-loaded code.
> * `gdb/NEWS` — not touched. A NEWS entry is the convention for documenting user-visible behavior changes to existing settings, and its absence should be flagged and fixed.
> 
> ## Nits / optional
> * `gdb/amd-dbgapi-target.c:892-895` — comment on the flag reset is slightly redundant with itself; consider trimming.
> * `gdb/testsuite/gdb.rocm/solib-event.exp:39` and `solib-event.cpp:56` exceed 80 columns — minor GNU style nit, not blocking.
> 
> ## Tests
> * Test coverage for the actual implemented behavior (boolean stop on GPU events, load/unload, `file://`/`memory://` URIs) is solid.
> * Given the description's claimed 0-3 value scheme isn't implemented, there's naturally no test for GPU-only (`2`) vs. combined (`3`) modes either — worth calling out so it isn't assumed to be covered.
> 
> Bottom line: the mechanical GPU-stop implementation and the `print_solib_event` refactor are reasonable and DRY, but (1) there's a real string bug that garbles the existing CPU message, and (2) neither the PR description's claimed CPU/GPU value scheme nor any of the documentation it implies (help text, manual, NEWS) actually landed in this diff — this should be resolved before merging so reviewers and future readers aren't misled about what `stop-on-solib-events` actually does after this change.

  Comment 0 - Summary:

  ▎ Review summary: The core fix (using stop_on_solib_events to control GPU code-object breakpoint stops, and
  ▎ generalizing print_solib_event) is sound and well-tested, but the PR description is significantly out of sync
  ▎ with the actual code, and there's a real user-visible message regression in the shared (CPU) path.

  All issues addressed in this round. PR description updated separately.

  ---
  Comment 1 - Pluralization Bug:

  ▎ Pluralization bug in the generalized "no libraries added" message. The new code does string_printf (_(\"... no
  ▎ %ss added or removed)\\n\"), event_description, event_description), which for the default/CPU case
  ▎ (event_description == "shared library") produces "no shared librarys added or removed" instead of the original,
  ▎ correct "no libraries added or removed". This is a regression in existing CPU-path user-visible output. Since
  ▎ irregular plurals can't be built by appending "s", the caller needs to pass the plural form explicitly (e.g. add
  ▎ a plural_description parameter, or pass a full pre-pluralized phrase) rather than deriving it mechanically.

  Fixed in commit 9dfd7532b5b. Added plural_description parameter to print_solib_event() with default value
  "libraries". GPU caller passes "GPU code objects". Also fixed 80-column violation on line 5272.

  ---
  Comment 2 - PR Description Mismatch:

  ▎ PR description vs. actual code mismatch: the PR body describes a "Commit 2: Add GPU/CPU control to
  ▎ stop-on-solib-events setting" that turns stop-on-solib-events into an integer-encoded 0-3 setting (0=none, 1=CPU
  ▎ only, 2=GPU only, 3=both), with an updated show message and help text. None of this exists in the actual diff.
  ▎ The real code (gdb/amd-dbgapi-target.c:931) simply does if (info.code_object_list_updated &&
  ▎ stop_on_solib_events != 0), reusing the existing boolean-ish semantics from internal_breakpoint::check_status.

  PR description updated to reflect actual boolean implementation where any non-zero value enables both CPU and GPU
  stops.

  ---
  Comment 3 - gdb.texinfo Missing:

  ▎ gdb/doc/gdb.texinfo — not touched. The stop-on-solib-events section of the user manual still only documents the
  ▎ boolean/CPU-only behavior; it does not mention that GPU code-object events are now covered.

  Fixed in commit fd733016ca8. Updated both set stop-on-solib-events and show stop-on-solib-events documentation to
  mention GPU code objects.

  ---
  Comment 4 - infrun.c Help Text:

  ▎ gdb/infrun.c:11035-11043 (add_setshow_zinteger_cmd help/show strings for stop-on-solib-events) — not touched.
  ▎ The in-tool help text still describes only the CPU dynamic-linker case, with no reference to GPU code objects.

  Fixed in commit fd733016ca8. Updated help text to include: "or when GPU code objects are loaded or unloaded (AMD
  ROCm targets)".

  ---
  Comment 5 - NEWS Entry:

  ▎ gdb/NEWS — not touched. A NEWS entry is the convention for documenting user-visible behavior changes to existing
  ▎ settings, and its absence should be flagged and fixed.

  Fixed in commit fd733016ca8. Added NEWS entry at top of "Changes since GDB 17" section describing GPU code object
  support for stop-on-solib-events.

  ---
  Comment 6 - Redundant Comment:

  ▎ gdb/amd-dbgapi-target.c:892-895 — comment on the flag reset is slightly redundant with itself; consider
  ▎ trimming.

  Fixed in commit fd733016ca8. Removed redundant sentence, keeping only: "Clear the flag so event processing below
  can detect a new CODE_OBJECT_LIST_UPDATED event."

  ---
  Comment 7 - 80-Column Violations:

  ▎ gdb/testsuite/gdb.rocm/solib-event.exp:39 and solib-event.cpp:56 exceed 80 columns — minor GNU style nit, not
  ▎ blocking.

  Fixed in commit fd733016ca8. All three violations corrected:
  - solib-event.exp line 39 (62 chars)
  - solib-event.exp line 100 (73 chars)
  - solib-event.cpp line 56 (66 chars)

PR #238, opened 2026-07-30
gdb/testsuite: resume-exception.exp - handle failure to hit breakpoint

akondrat-amd:
If GDB fails to hit the breakpoint in raise_fpe, gpu_thread is never
set and referencing it later would cause a Tcl error.  Initialize
gpu_thread to "invalid" before running and return early if it was not
updated by the breakpoint handler.

Jira: AIROCGDB-640

### DIFF
--- a/gdb/testsuite/gdb.rocm/resume-exception.exp
+++ b/gdb/testsuite/gdb.rocm/resume-exception.exp
@@ -123,11 +123,15 @@ proc interrupt_running_threads { } {
 with_rocm_gpu_lock {
     # Run to GPU thread.
     gdb_breakpoint "raise_fpe" -allow-pending
+    set gpu_thread "invalid"
     gdb_test_multiple "run" "" {
 	-re -wrap ".*Thread ($::decimal) \[^\r\n]* hit Breakpoint .*" {
 	    set gpu_thread $expect_out(1,string)
 	    pass $gdb_test_name
 	}
+    }
+    if {$gpu_thread eq "invalid"} {
+	return
     }
     gdb_test "thread $gpu_thread" "raise_fpe.*" "select GPU thread"
 


lumachad:
Hi @akondrat-amd. Do you know why such a thing would happen? Is it a timeout issue, a specific architecture that behaves differently? A bug?

gdb/testsuite/gdb.rocm/resume-exception.exp
```diff
@@ -127,7 +127,11 @@ with_rocm_gpu_lock {
...
 	    set gpu_thread $expect_out(1,string)
 	    pass $gdb_test_name
-	}
+		}
+	-re -wrap ".*" {
+	    fail $gdb_test_name
+        return
+    	}
     }
     gdb_test "thread $gpu_thread" "raise_fpe.*" "select GPU thread"
```
aktemur, line 136: AFAIU, the main problem for running the testsuite is `gpu_thread` is undefined in case the breakpoint above is not hit, which leads to a Tcl error. Apart from the fact that we need to understand why the BP is not hit, as commented by @lumachad, the usual way to address to the undefined variable issue is to define it before `gdb_test_multiple`, like so:
```
set gpu_thread "invalid"
gdb_test_multiple "run" ...
```

This is necessary in any case to avoid using a leaked global variable from another test. With that change, I think the `".*"` case can be removed. We may want to change nothing else and let the subsequent tests fail (e.g. "thread invalid" command will fail and the failure will propagate to other tests). This is better than a Tcl error.

If the test is causing other problems or taking too much time because of timeouts when the BP is not hit, we can also add
```
if {$gpu_thread eq "invalid} {
    fail ....
    return
}
```
after `gdb_test_multiple`.

I'm not sure `fail` above would even be needed, because `gdb_test_multiple` that wants to capture `gpu_thread` would have already failed. So, we would already have at least one failure.

palves: > This is necessary in any case to avoid using a leaked global variable from another test.

Note globals don't leak between tests anymore nowadays.  See gdb_cleanup_globals in lib/gdb.exp.

palves: Though TBC, I agree that initializing the variable upfront is better, also because the `-re -wrap ".*"` case will NOT catch all failures.  If GDB crashes with an internal error for example, it won't, that is caught with an regexp internal to gdb_test_multiple.  This is why it is typically better to do the 

```
if {$gpu_thread eq "invalid"} {
    return
}
```

... style of early-return, after gdb_test_multiple.

aktemur: > Note globals don't leak between tests anymore nowadays. See gdb_cleanup_globals in lib/gdb.exp.

Good to know, thanks!

akondrat-amd:
> Hi @akondrat-amd. Do you know why such a thing would happen? Is it a timeout issue, a specific architecture that behaves differently? A bug?

It's an issue with rocjitsu that will be eventually fixed.

gdb/testsuite/gdb.rocm/resume-exception.exp
akondrat-amd: Thanks, Pedro, I've updated the commit.

lumachad:
> > Hi @akondrat-amd. Do you know why such a thing would happen? Is it a timeout issue, a specific architecture that behaves differently? A bug?
> 
> It's an issue with rocjitsu that will be eventually fixed.

Why would we need to adjust ROCgdb if it's a rocjitsu issue?

akondrat-amd:
> > > Hi @akondrat-amd. Do you know why such a thing would happen? Is it a timeout issue, a specific architecture that behaves differently? A bug?
> > 
> > 
> > It's an issue with rocjitsu that will be eventually fixed.
> 
> Why would we need to adjust ROCgdb if it's a rocjitsu issue?

It's our bug, we do not initialize the variable, rocjitsu just exposed it by failing in the right place. It's more cosmetic than functional. If the subtest fails, we get a TCL error with a call stack instead of test failure.

lumachad:
> > > > Hi @akondrat-amd. Do you know why such a thing would happen? Is it a timeout issue, a specific architecture that behaves differently? A bug?
> > > 
> > > 
> > > It's an issue with rocjitsu that will be eventually fixed.
> > 
> > 
> > Why would we need to adjust ROCgdb if it's a rocjitsu issue?
> 
> It's our bug, we do not initialize the variable, rocjitsu just exposed it by failing in the right place. It's more cosmetic than functional. If the subtest fails, we get a TCL error with a call stack instead of test failure.

Oops. Sorry, that wasn't clear to me.

lumachad (approved):
This looks OK to me. Thanks.

PR #239, opened 2026-07-30
feat(test_rocgdb): add --timing, --toolchain, and --sanity-check modes

lumachad:
## Summary

Three new `test_rocgdb.py` options:

- `--toolchain NAME [NAME ...]` — run against selected compiler(s) only (default: all, in order). Only the selected toolchains' compilers must be on PATH.
- `--timing` — record per-test wall-clock time in one-by-one mode and write `rocgdb_timing.log` (grouped by compiler/directory, longest first; N/A for tests that never passed). `--timing-log-file` overrides the path.
- `--sanity-check` (needs `--one-by-one`) — build a known-good HIP program once from `gdb.rocm/simple.cpp` and run it before each test as a GPU health probe. On timeout or non-zero exit, log the output, capture `dmesg`, and abort instead of emitting spurious failures. `--sanity-check-timeout` (default 10s) tunes the probe.

Notes: the probe is rebuilt every run (never cached, since `--offload-arch=native` bakes in the GPU); `dmesg` falls back to `sudo -n dmesg` and warns if unavailable; `amdclang++` is required only when `gdb.rocm/` tests are in scope.

## Test plan

- [x] `--help` lists `--toolchain`, `--timing`, `--sanity-check`
- [x] `--toolchain gnu` / `llvm` run one compiler and don't require the other's tools
- [x] `--timing --gpu-tests` writes `rocgdb_timing.log`; `--timing --parallel` is rejected
- [x] `--sanity-check` is rejected without `--one-by-one`, with `--parallel`, or with no `gdb.rocm` tests in scope
- [x] GPU box: `--one-by-one --sanity-check --gpu-tests` builds the probe once and runs healthy; forced failure writes `dmesg.log` and exits 1

### DIFF
--- a/.github/scripts/test_rocgdb.py
+++ b/.github/scripts/test_rocgdb.py
@@ -55,6 +55,15 @@ ONE_BY_ONE_DEFAULT_WALL_CLOCK_TIMEOUT = 3600
 # ignored at runtime, so listing extras here is harmless.
 SANITIZERS = ("asan", "tsan", "ubsan", "msan", "lsan", "hwasan")
 
+# Supported toolchains, keyed by the identifier the toolchain option accepts.
+# Each one maps to the compiler label (a load bearing key used across results
+# and ignore lists) and the C, C++, and Fortran executables we pass to the
+# testsuite. Insertion order defines the default run order.
+TOOLCHAINS = {
+    "gnu": {"label": "GCC", "cc": "gcc", "cxx": "g++", "fc": "gfortran"},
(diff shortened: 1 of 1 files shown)


lumachad:
@akondrat-amd You might want to give this a try/review.

akondrat-amd (approved):
Tried with those options, works great:
python3 test_rocgdb.py \
  --try-rocm-path \
  --tests gdb.rocm \
  --runtestflags "-v -a --log_dialog" \
  --max-failed-retries 0 \
  --one-by-one
  --one-by-one-log-dir $LOGDIR \
  --toolchain llvm \
  --sanity-check

PR #242, opened 2026-08-03
gdb/testsuite: strip ANSI escapes before pruning omp-rocm ld.lld warning

spatrang:
## Summary

The OpenMP-offload ROCm tests (`gdb.rocm/omp-target-*.exp`) compile device code with `amdclang`. The device linker `ld.lld` emits a benign warning:

```
ld.lld: warning: <unknown>:0:0: in function __keep_alive void (): local memory global used by non-kernel function
```

`gdb_compile` treats leftover compiler/linker output as a build failure, so an existing `regsub` already prunes this warning. That prune is anchored on the literal `ld.lld: warning:`.

With newer toolchains (LLVM 23), `ld.lld` colorizes its own diagnostics: `--color-diagnostics` defaults to `auto` and turns on whenever stderr is a tty, which is the case when the compiler is spawned under a pty (as DejaGnu does), regardless of clang's `-fdiagnostics-color=never`. The warning then carries ANSI SGR escapes (e.g. `ESC[0;35m`) between `ld.lld:` and `warning:`, defeating the anchored regexp. The warning survives, the build is treated as failed, and the testcase is skipped (UNSUPPORTED/UNTESTED).

This strips ANSI SGR escapes from the captured compiler output before the existing `omp-rocm` prunes run. Only escape bytes are removed (never diagnostic text), so genuine `error:` diagnostics are still detected and real build failures still fail.

## JIRA

AIROCGDB-642

### DIFF
--- a/gdb/testsuite/lib/gdb.exp
+++ b/gdb/testsuite/lib/gdb.exp
@@ -7006,6 +7006,19 @@ proc gdb_compile {source dest type options} {
 	set omp_early_flags "-fopenmp -fopenmp-targets=amdgcn-amd-amdhsa\
 			     -mllvm=-amdgpu-spill-cfi-saved-regs"
 
+	# The benign __keep_alive warning pruned below is emitted by the
+	# device ld.lld at link time.  Newer lld (e.g. LLVM 23) defaults
+	# --color-diagnostics to "auto" and colorizes when spawned under a
+	# pty (as DejaGnu does), inserting ANSI escapes between "ld.lld:"
+	# and "warning:" that would stop the anchored regexp from matching.
+	# Tell the offload (device) linker not to colorize; only on the
+	# link step, so the compile step does not warn about an unused
+	# argument.  -Wl, would only reach the host linker, not the device
+	# ld.lld that emits this warning.
+	if {$type eq "executable"} {
+	    append omp_early_flags " -Xoffload-linker --no-color-diagnostics"
+	}
+
 	# amdflang does not understand some clang-only diagnostic
 	# flags, so only pass them on the C/C++ (non-f90) paths.
 	if {[lsearch -exact $options f90] == -1} {


gdb/testsuite/lib/gdb.exp
```diff
@@ -7307,6 +7307,13 @@ proc gdb_compile {source dest type options} {
     # Scoped to omp-rocm builds, whose flags we control, and to the
     # unused-argument warning specifically (errors are never matched).
     if {[lsearch -exact $options "omp-rocm"] != -1} {
+	# Newer lld/clang (e.g. LLVM 23) colorize their diagnostics with
+	# ANSI escape sequences even with -fdiagnostics-color=never (lld
+	# has its own --color-diagnostics, which defaults to "auto" and
```
lancesix: Can't we use `-Wl,--color-diagnostics=never` instead?

spatrang: Good idea, but `-Wl,--color-diagnostics=never` doesn't work here: this warning is emitted by the **device** `ld.lld` (the amdgcn offload-image link run by the offload wrapper), whereas `-Wl,` only reaches the **host** linker. I tested it under a color pty (LLVM 23) and the warning stays colorized; it also adds a `-Wunused-command-line-argument` at the `-c` step.

The working equivalent is `-Xoffload-linker --no-color-diagnostics`, which is forwarded to the device linker and disables coloring at the source (warning then comes out plain-text, so the existing anchored `ld.lld: warning: ... __keep_alive ...` prune matches without needing the escape strip). Verified accepted on LLVM 22 and 23.

One impact to note if we switch to it: since it lives in `omp_early_flags` it's also passed on the compile (`-c`) step. On C/C++ that's silenced by `-Wno-unused-command-line-argument`, but the Fortran path (amdflang) doesn't take that flag, so `-c` emits:

```
flang-23: warning: argument unused during compilation: '-Xoffload-linker --no-color-diagnostics' [-Wunused-command-line-argument]
```

That's benign and already removed by the existing unused-argument prune a few lines below, so no test outcome changes and no skips — but it does add one more pruned warning on the Fortran path. The current ANSI-strip introduces no new warnings and handles C/C++/Fortran uniformly.

Both are functionally correct. Do you want me to switch to `-Xoffload-linker --no-color-diagnostics` (source suppression, with the extra pruned Fortran warning), or keep the ANSI-strip? Happy to change it either way.

lancesix: I think I'd prefer the flag to not have color rather than filetr the ASCII colouring.

spatrang:
Updated per review: switched from stripping the ANSI escapes to suppressing device-linker coloring at the source.

The omp-rocm build now passes `-Xoffload-linker --no-color-diagnostics` (added **only on the link step**, `$type eq "executable"`), so the device `ld.lld` never colorizes the benign `__keep_alive` warning and the existing anchored prune matches. As discussed, `-Wl,` only reaches the host linker, not the device `ld.lld` that emits this warning.

Adding it link-only avoids an unused-argument warning on the Fortran `-c` step (amdflang doesn't take `-Wno-unused-command-line-argument`), so no new warnings are introduced on any language path.

Verified with LLVM 22/23: C/C++/Fortran compile+link succeed and the warning is no longer colorized.

lancesix (approved):
LGTM, thanks.  Maybe, please wait for https://github.com/ROCm/ROCgdb/pull/244 to land before merging to difficulties merging the merge from upstream.

spatrang:
> LGTM, thanks. Maybe, please wait for #244 to land before merging to difficulties merging the merge from upstream.

Sure, @lancesix

lumachad:
Should we cherry-pick this to amd-staging-rocgdb-16?

lancesix:
> Should we cherry-pick this to amd-staging-rocgdb-16?

Most probably, yes.

lumachad:
> > Should we cherry-pick this to amd-staging-rocgdb-16?
> 
> Most probably, yes.

Done in https://github.com/ROCm/ROCgdb/pull/252.

PR #243, opened 2026-08-04
github: add daily TheRock dependency update workflow

lumachad:
Adds a workflow intended to run twice daily (03:17 and 15:17 UTC) to bump `THEROCK_COMMIT_REF` and the `therock_build_manylinux_x86_64` container digest in `therock-ci-linux.yml`. If an open PR labelled `therock-deps` already exists, the run skips cleanly.

The schedule is currently commented out for testing; the workflow can be triggered manually via workflow_dispatch in the meantime.

The build container digest is fetched via the OCI Distribution API using Python's stdlib `urllib`, avoiding any external tool dependency. The TheRock commit is resolved against `refs/heads/main` explicitly rather than relying on HEAD.

### DIFF
--- a/.github/scripts/update_therock_deps.py
+++ b/.github/scripts/update_therock_deps.py
@@ -1,0 +1,368 @@
+#!/usr/bin/env python3
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+"""Update TheRock references in ROCgdb's CI workflow.
+
+Fetches the latest TheRock commit and the latest therock_build_manylinux_x86_64
+container digest, updates .github/workflows/therock-ci-linux.yml, then pushes a
+branch and opens a ROCgdb PR labelled 'therock-deps'. If an update PR already
+carries that label, the run skips without touching git.
+
+The test container is not updated here: it is pinned transitively through
+THEROCK_COMMIT_REF (TheRock's fetch_test_configurations.py supplies the test
+image), so bumping the commit ref updates it automatically.
+"""
+
+import argparse
+import json
+import re
+import shlex
+import subprocess
+import sys
+import time
+import urllib.error
+import urllib.request
+from datetime import date
+from pathlib import Path
+
+# ---------------------------------------------------------------------------
+# Configuration
+# ---------------------------------------------------------------------------
+
+ROCGDB_REPO = "ROCm/ROCgdb"
+BASE_BRANCH = "amd-staging"
+CI_LINUX = Path(".github/workflows/therock-ci-linux.yml")
+
+THEROCK_REPO_URL = "https://github.com/ROCm/TheRock.git"
+BUILD_IMAGE = "ghcr.io/rocm/therock_build_manylinux_x86_64"
+
+UPDATE_LABEL = "therock-deps"
+BRANCH_PREFIX = "users/github/update-therock-refs"
+
+SHA_RE = re.compile(r"^[0-9a-f]{40}$")
+DIGEST_RE = re.compile(r"^sha256:[0-9a-f]{64}$")
+
+MAX_RETRIES = 3  # number of retries for network requests
+
+# Never pass secrets as CLI args — use env vars (e.g. GH_TOKEN) instead.
+
+
+# ---------------------------------------------------------------------------
+# Network helper
+# ---------------------------------------------------------------------------
+
+
+def urlopen_with_retry(req, timeout: int = 30):
+    """Open a URL with retries on transient errors."""
+    for attempt in range(1, MAX_RETRIES + 1):
+        try:
+            return urllib.request.urlopen(req, timeout=timeout)
+        except (urllib.error.URLError, TimeoutError) as exc:
+            if attempt == MAX_RETRIES:
+                raise
+            print(f"[RETRY] attempt {attempt}/{MAX_RETRIES} failed: {exc}")
+            time.sleep(2 ** attempt)
+
+
+# ---------------------------------------------------------------------------
+# Subprocess helper
+# ---------------------------------------------------------------------------
+
+
+def run(cmd: list[str], check: bool = True) -> subprocess.CompletedProcess:
+    """Run a subprocess, printing the command that is executed."""
+    print(f"[RUN] {' '.join(shlex.quote(str(a)) for a in cmd)}")
+    result = subprocess.run(cmd, capture_output=True, text=True)
+    if result.stdout:
+        print(result.stdout.rstrip())
+    if result.returncode != 0 and result.stderr:
+        print(result.stderr.rstrip(), file=sys.stderr)
+    if check and result.returncode != 0:
+        raise subprocess.CalledProcessError(
+            result.returncode, cmd, result.stdout, result.stderr
+        )
+    return result
+
+
+# ---------------------------------------------------------------------------
+# GitHub helpers (gh CLI)
+# ---------------------------------------------------------------------------
+
+
+def ensure_label(label: str) -> None:
+    """Verify that the label exists in the repo; raise if it does not."""
+    result = run(
+        [
+            "gh",
+            "label",
+            "list",
+            "--repo",
+            ROCGDB_REPO,
+            "--search",
+            label,
+            "--json",
+            "name",
+        ],
+    )
+    existing = [entry["name"] for entry in json.loads(result.stdout or "[]")]
+    if label not in existing:
+        raise RuntimeError(
+            f"Label '{label}' not found in {ROCGDB_REPO}. "
+            "Please create it in the repo settings before running."
+        )
+
+
+def find_open_update_pr() -> str | None:
+    """Return the URL of an open, labelled update PR if one exists, else None."""
+    result = run(
+        [
+            "gh",
+            "pr",
+            "list",
+            "--repo",
+            ROCGDB_REPO,
+            "--base",
+            BASE_BRANCH,
+            "--state",
+            "open",
+            "--label",
+            UPDATE_LABEL,
+            "--json",
+            "url",
+            "--jq",
+            ".[0].url // empty",
+        ],
+    )
+    url = (result.stdout or "").strip()
+    return url if url else None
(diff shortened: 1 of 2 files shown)


.github/scripts/update_therock_deps.py
```diff
@@ -0,0 +1,320 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Fetch latest references
+# ---------------------------------------------------------------------------
+
+
+def get_therock_commit() -> tuple[str, str]:
+    """Return (commit_sha, today) for TheRock's HEAD."""
+    result = run(["git", "ls-remote", THEROCK_REPO_URL, "HEAD"])
```
lancesix: probably `s#HEAD#refs/heads/main#` to actually name the branch we track and not rely on HEAD

.github/scripts/update_therock_deps.py
```diff
@@ -0,0 +1,320 @@
...
+    parts = result.stdout.split()
+    commit = parts[0] if parts else ""
+    if not SHA_RE.match(commit):
+        raise ValueError(f"Invalid TheRock SHA received: {commit!r}")
+    return commit, date.today().isoformat()
+
+
+def get_build_digest() -> tuple[str, str]:
+    """Return (digest, created_date) for the latest build container image."""
+    result = run(["skopeo", "inspect", f"docker://{BUILD_IMAGE}:latest"])
```
lancesix: Is that tool guaranteed to be present?  Feels there should be an endpoint we could query to get this information without introducing a dependency.

.github/workflows/therock-deps-update.yml
```diff
@@ -0,0 +1,41 @@
...
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Install skopeo
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y skopeo
```
lancesix: I guess that answers one of the previous question.

lancesix (commented):
quick pass, with some minor remarks.

.github/scripts/update_therock_deps.py
lumachad: Done, changed to `refs/heads/main`.

.github/workflows/therock-deps-update.yml
lumachad: Right, but see above...

.github/scripts/update_therock_deps.py
lumachad: There is an endpoint. This sort of evolved from a helper script, and it was using skopeo then. I've switched to the OCI registry.

.github/workflows/therock-deps-update.yml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Configure git identity and credentials
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+          git config --global credential.helper "!gh auth git-credential"
+
+      - name: Update TheRock references
+        run: |
+          python3 .github/scripts/update_therock_deps.py ${{ inputs.dry_run != 'false' && '--dry-run' || '' }}
```
spatrang: **CRITICAL:** The dry-run gate always passes `--dry-run`, so the scheduled job can never open a PR.

- On `schedule`, `inputs` is empty → `inputs.dry_run` is null; GitHub compares numerically (`null`→0, `'false'`→NaN) so `0 != NaN` is `true` → `--dry-run` gets added.
- On `workflow_dispatch`, `dry_run` is a boolean; both `true != 'false'` and `false != 'false'` coerce to `true`, so even selecting "dry run = false" still passes `--dry-run`.

Suggest relying on the boolean directly (null-on-schedule is falsy, giving a real run):

```yaml
python3 .github/scripts/update_therock_deps.py ${{ inputs.dry_run && '--dry-run' || '' }}
```

.github/workflows/therock-deps-update.yml
```diff
@@ -0,0 +1,41 @@
...
+concurrency:
+  group: therock-deps-update
+  cancel-in-progress: false
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
```
spatrang, line 26: **IMPORTANT:** `ensure_label()` in the script runs `gh label create`, and labels are part of the Issues API, so the token needs `issues: write`. This job only grants `contents`/`pull-requests`, so the first real run (before the `therock-deps` label exists) will likely 403 and abort. Once the label exists `ensure_label()` skips creation, which makes this an easy-to-miss first-run failure.

Either add `issues: write` here, or pre-create the `therock-deps` label once in repo settings.

.github/workflows/therock-deps-update.yml
```diff
@@ -0,0 +1,41 @@
...
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
spatrang, line 28: **IMPORTANT:** PRs created with the default `GITHUB_TOKEN` do not trigger further workflow runs (GitHub's recursion guard). Since the whole point of this automation is to get CI to validate the bumped TheRock deps, the auto-opened PR will have an empty checks section — yet the PR body the script writes says "Please validate CI, then merge."

Options: use a PAT or GitHub App token when creating the PR (actions by those do trigger CI), or update the PR body to tell the reviewer to manually kick CI (e.g. close/reopen or push an empty commit) before merging.

.github/scripts/update_therock_deps.py
```diff
@@ -0,0 +1,343 @@
...
+    return url if url else None
+
+
+# ---------------------------------------------------------------------------
+# Fetch latest references
+# ---------------------------------------------------------------------------
+
+
+def get_therock_commit() -> tuple[str, str]:
+    """Return (commit_sha, today) for TheRock's HEAD."""
```
spatrang: Minor: stale docstring — says "for TheRock's HEAD" but the function now fetches `refs/heads/main`. Update "HEAD" → "main".

.github/scripts/update_therock_deps.py
```diff
@@ -0,0 +1,343 @@
...
+        headers={
+            "Authorization": f"Bearer {token}",
+            "Accept": (
+                "application/vnd.docker.distribution.manifest.v2+json,"
+                "application/vnd.oci.image.manifest.v1+json,"
+                "application/vnd.oci.image.index.v1+json"
+            ),
+        },
+    )
+    with urllib.request.urlopen(req) as resp:
```
spatrang: Minor: no `timeout=` on either `urllib.request.urlopen` call (here and the manifest request below). A hung GHCR endpoint would stall the job until the runner's global timeout. Suggest passing e.g. `timeout=30`.

spatrang (changes requested):

.github/workflows/therock-deps-update.yml
lumachad: Fixed. Changed the expression to `${{ inputs.dry_run && '--dry-run' || '' }}`. On `schedule` `inputs.dry_run` is null (falsy), giving a real run; on `workflow_dispatch` the boolean `false` is also falsy, so unchecking the box now correctly omits `--dry-run`.

Also commented out the `schedule:` trigger for now so we can test the workflow manually via `workflow_dispatch` before enabling the schedule.

.github/workflows/therock-deps-update.yml
lumachad: Noted. We will add `issues: write` once we validate the workflow runs correctly end-to-end. In the meantime, the `therock-deps` label can be pre-created in repo settings to avoid the 403 on the first real run.

.github/workflows/therock-deps-update.yml
lumachad: Acknowledged. Updated the PR body to tell reviewers to manually kick CI (close/reopen or push an empty commit) before merging. A PAT or App token can be wired in later if we want fully automated validation.

.github/scripts/update_therock_deps.py
lumachad: Fixed.

.github/scripts/update_therock_deps.py
lumachad: Fixed. Both `urlopen` calls now go through a `urlopen_with_retry` helper with a 30 s timeout and up to `MAX_RETRIES = 3` attempts with exponential backoff.

.github/workflows/therock-deps-update.yml
lumachad: I've now changed this. I've create the label manually and we don't need to bother with adding extra permissions.

spatrang: Sounds good — pre-creating the label covers the first-run 403, so no objection to deferring the permission. Worth a quick follow-up issue/TODO so `issues: write` is not forgotten when the schedule is enabled.

.github/workflows/therock-deps-update.yml
```diff
@@ -0,0 +1,41 @@
+name: Update TheRock Dependencies
+
+on:
+  # schedule: # runs twice daily (03:17 and 15:17 UTC)
+  #   - cron: '17 3 * * *'
+  #   - cron: '17 15 * * *'
```
spatrang, line 6: Non-blocking: the `schedule:` triggers are commented out, so the workflow is manual-only (`workflow_dispatch`) for now - makes sense for validating via manual runs first. Consider a brief inline note (or a tracking issue) so re-enabling the cron isn't forgotten.

spatrang (approved):
LGTM. One non-blocking question below.

.github/workflows/therock-deps-update.yml
lumachad: Not a concern. This is default dry-run for experimentation before we activate it.

PR #245, opened 2026-08-04
[workflows] Run extended ROCgdb tests in parallel on PRs

lumachad:
## Summary

Add five extended test configurations that run in parallel with the
standard tests on every pull request. All jobs reuse the single build
from `therock-build-linux` (no extra build cost) and fan out together
once the build completes.

The extended configurations are:

| Job | Options | Architectures |
|-----|---------|---------------|
| `Test GPU (-O3)` | `--optimization=-O3` | all (expands with `ci:run-all-archs`) |
| `Test GPU (-O3 -flto)` | `--optimization='-O3 -flto'` | all (expands with `ci:run-all-archs`) |
| `Test check-read1` | `--check-type check-read1` | gfx94X only |
| `Test check-readmore` | `--check-type check-readmore` | gfx94X only |
| `Test GPU (hip board)` | `--runtestflags=--target_board=hip --tests gdb.base/...` | gfx94X only (GPU only) |

**GPU runner budget:** without `ci:run-all-archs`, all five extended jobs
run on gfx94X only (5 extra GPU jobs). With `ci:run-all-archs`, the two
optimization jobs each fan out to all six architecture families (12 extra
GPU jobs from optimization alone, plus 3 for check-type/hip-board). The
multi-arch coverage for optimization jobs is intentional: the goal is to
catch compiler-specific failures across all supported architectures.

To support per-config scoping, `therock-test-packages.yml` gains one
new input and updates another:
- `test_options` — appended to each component's `test_script` (passed
  via a `TEST_OPTIONS` env var in the jq patch step to avoid injection).
  When `test_options` contains `--tests`, any pre-existing `--tests` in
  the upstream `test_script` is stripped first to make the override
  explicit rather than relying on argparse last-wins.
- `projects_to_test` — restricts the component matrix (already existed;
  gains a description). The two optimization jobs use
  `rocgdb-gpu,rocgdb-corefile` since the options are compiler-specific;
  the hip board job uses `rocgdb-gpu` only.

## Commits

1. `[workflows] Drop now-redundant test_rocgdb.py path rewrite` — TheRock
   now emits the installed script path directly; the gsub rewrite was
   already a no-op at the pinned commit. Removes `OUTPUT_ARTIFACTS_DIR`.
2. `[workflows] Run extended ROCgdb tests in parallel on PRs` — the
   extended jobs and updated `therock-test-packages.yml` inputs.

### DIFF
--- a/.github/workflows/therock-ci.yml
+++ b/.github/workflows/therock-ci.yml
@@ -138,6 +138,100 @@ jobs:
       artifact_run_id: ${{ github.run_id }}
       projects_to_test: rocgdb-corefile
 
+  # Extended test configurations. These reuse the single build from
+  # therock-build-linux (no rebuild) and fan out in parallel with each other
+  # and with the standard tests, all gated on a successful build.
+  #
+  # The two optimization jobs (test-gpu-o3, test-gpu-lto) follow the same
+  # test_families matrix as test-gpu, so with ci:run-all-archs each adds up
+  # to six GPU jobs (one per architecture family). The check-type and hip
+  # board jobs run on gfx94X only.
+  test-gpu-o3:
+    name: Test GPU (-O3)
+    needs: [setup, therock-build-linux]
+    if: ${{ needs.setup.outputs.enable_therock_ci == 'true' }}
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        amdgpu_family: ${{ fromJSON(needs.setup.outputs.test_families) }}
+    uses: ./.github/workflows/therock-test-packages.yml
+    with:
+      amdgpu_families: ${{ matrix.amdgpu_family }}
+      artifact_group: ${{ needs.setup.outputs.artifact_group }}
+      therock_ref: ${{ needs.therock-build-linux.outputs.therock_ref }}
+      artifact_run_id: ${{ github.run_id }}
+      projects_to_test: rocgdb-gpu,rocgdb-corefile
+      test_options: "--optimization=-O3"
+
+  test-gpu-lto:
+    name: Test GPU (-O3 -flto)
+    needs: [setup, therock-build-linux]
+    if: ${{ needs.setup.outputs.enable_therock_ci == 'true' }}
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        amdgpu_family: ${{ fromJSON(needs.setup.outputs.test_families) }}
+    uses: ./.github/workflows/therock-test-packages.yml
+    with:
+      amdgpu_families: ${{ matrix.amdgpu_family }}
+      artifact_group: ${{ needs.setup.outputs.artifact_group }}
+      therock_ref: ${{ needs.therock-build-linux.outputs.therock_ref }}
+      artifact_run_id: ${{ github.run_id }}
+      projects_to_test: rocgdb-gpu,rocgdb-corefile
+      test_options: "--optimization='-O3 -flto'"
+
+  test-gpu-check-read1:
+    name: Test GPU (check-read1)
+    needs: [setup, therock-build-linux]
+    if: ${{ needs.setup.outputs.enable_therock_ci == 'true' }}
+    permissions:
+      contents: read
+    uses: ./.github/workflows/therock-test-packages.yml
+    with:
+      amdgpu_families: gfx94X
+      artifact_group: ${{ needs.setup.outputs.artifact_group }}
+      therock_ref: ${{ needs.therock-build-linux.outputs.therock_ref }}
+      artifact_run_id: ${{ github.run_id }}
+      test_options: "--check-type check-read1"
+
+  test-gpu-check-readmore:
+    name: Test GPU (check-readmore)
+    needs: [setup, therock-build-linux]
+    if: ${{ needs.setup.outputs.enable_therock_ci == 'true' }}
+    permissions:
+      contents: read
+    uses: ./.github/workflows/therock-test-packages.yml
+    with:
+      amdgpu_families: gfx94X
+      artifact_group: ${{ needs.setup.outputs.artifact_group }}
+      therock_ref: ${{ needs.therock-build-linux.outputs.therock_ref }}
+      artifact_run_id: ${{ github.run_id }}
+      test_options: "--check-type check-readmore"
+
+  test-gpu-hip-board:
+    name: Test GPU (hip board)
+    needs: [setup, therock-build-linux]
+    if: ${{ needs.setup.outputs.enable_therock_ci == 'true' }}
+    permissions:
+      contents: read
+    uses: ./.github/workflows/therock-test-packages.yml
+    with:
+      amdgpu_families: gfx94X
+      artifact_group: ${{ needs.setup.outputs.artifact_group }}
+      therock_ref: ${{ needs.therock-build-linux.outputs.therock_ref }}
+      artifact_run_id: ${{ github.run_id }}
+      projects_to_test: rocgdb-gpu
+      test_options: >-
+        --runtestflags=--target_board=hip
+        --tests gdb.base/attach.exp gdb.base/start.exp gdb.base/callfuncs.exp
+        gdb.base/consecutive-step-over.exp gdb.base/ending-run.exp
+        gdb.base/finish.exp gdb.base/hbreak.exp gdb.base/signals.exp
+        gdb.base/step-test.exp gdb.base/watchpoint.exp
+
   therock_ci_summary:
     name: TheRock CI Summary
     if: always()
@@ -147,6 +241,11 @@ jobs:
       - test-gpu
       - test-cpu
       - test-corefile
+      - test-gpu-o3
+      - test-gpu-lto
+      - test-gpu-check-read1
+      - test-gpu-check-readmore
+      - test-gpu-hip-board
     runs-on: ubuntu-24.04
     steps:
       - name: Output failed jobs
--- a/.github/workflows/therock-ci.yml
+++ b/.github/workflows/therock-ci.yml
@@ -185,7 +185,7 @@ jobs:
       test_options: "--optimization='-O3 -flto'"
 
   test-gpu-check-read1:
-    name: Test GPU (check-read1)
+    name: Test check-read1
     needs: [setup, therock-build-linux]
     if: ${{ needs.setup.outputs.enable_therock_ci == 'true' }}
     permissions:
@@ -199,7 +199,7 @@ jobs:
       test_options: "--check-type check-read1"
 
   test-gpu-check-readmore:
-    name: Test GPU (check-readmore)
+    name: Test check-readmore
     needs: [setup, therock-build-linux]
     if: ${{ needs.setup.outputs.enable_therock_ci == 'true' }}
     permissions:
--- a/.github/workflows/therock-test-packages.yml
+++ b/.github/workflows/therock-test-packages.yml
@@ -13,9 +13,14 @@ on:
         required: true
       artifact_run_id:
         type: string
+      test_options:
+        type: string
+        description: 'Job-specific options appended to each component test_script.'
+        default: ''
       projects_to_test:
         type: string
-        default: "rocgdb-cpu,rocgdb-gpu,rocgdb-corefile"
+        description: 'Components to build the test matrix for.'
+        default: 'rocgdb-cpu,rocgdb-gpu,rocgdb-corefile'
   workflow_dispatch:
     inputs:
       amdgpu_families:
@@ -28,9 +33,14 @@ on:
         required: true
       artifact_run_id:
         type: string
+      test_options:
+        type: string
+        description: 'Job-specific options appended to each component test_script.'
+        default: ''
       projects_to_test:
         type: string
-        default: "rocgdb-cpu,rocgdb-gpu,rocgdb-corefile"
+        description: 'Components to build the test matrix for.'
+        default: 'rocgdb-cpu,rocgdb-gpu,rocgdb-corefile'
 
 permissions:
   contents: read
@@ -84,13 +94,32 @@ jobs:
         id: configure
         run: python ./build_tools/github_actions/fetch_test_configurations.py
 
-      - name: Set rocgdb test_script
-        # TheRock now emits the installed script path directly; no rewrite needed.
+      - name: Patch rocgdb test_script options
+        # Append caller-supplied options to each component's test_script.
+        # When test_options contains --tests, first strip any pre-existing
+        # --tests from the upstream test_script so the caller's value wins
+        # explicitly rather than relying on argparse last-wins behaviour.
         id: patch-scripts
         env:
           COMPONENTS: ${{ steps.configure.outputs.components }}
+          TEST_OPTIONS: ${{ inputs.test_options }}
         run: |
-          echo "components=$COMPONENTS" >> $GITHUB_OUTPUT
+          strip_tests=false
+          if [[ " $TEST_OPTIONS " == *" --tests "* ]]; then
+            strip_tests=true
+          fi
+          updated=$(echo "$COMPONENTS" | jq -c \
+            --arg opts "$TEST_OPTIONS" \
+            --argjson strip "$strip_tests" '
+            map(
+              .test_script |= (
+                if $strip then gsub("--tests(\\s+[^-]\\S*)*"; "") else . end
+                | if $opts != "" then . + " " + $opts else . end
+                | gsub("  +"; " ") | sub("\\s+$"; "")
+              )
+            )
+          ')
+          echo "components=$updated" >> $GITHUB_OUTPUT
 
   test_components:
     name: "Test ${{ matrix.components.job_name }}"
--- a/.github/workflows/therock-test-packages.yml
+++ b/.github/workflows/therock-test-packages.yml
@@ -37,7 +37,6 @@ permissions:
 
 env:
   VENV_DIR: ${{ github.workspace }}/.venv
-  OUTPUT_ARTIFACTS_DIR: "./build"
 
 jobs:
   configure_test_matrix:
@@ -85,23 +84,13 @@ jobs:
         id: configure
         run: python ./build_tools/github_actions/fetch_test_configurations.py
 
-      - name: Override rocgdb test_script to use artifact-installed script
-        # TheRock's fetch_test_configurations.py points test_rocgdb.py at its own
-        # build_tools copy. Rewrite to the script installed in the artifacts at
-        # OUTPUT_ARTIFACTS_DIR/tests/rocgdb/test_rocgdb.py.
+      - name: Set rocgdb test_script
+        # TheRock now emits the installed script path directly; no rewrite needed.
         id: patch-scripts
+        env:
+          COMPONENTS: ${{ steps.configure.outputs.components }}
         run: |
-          components='${{ steps.configure.outputs.components }}'
-          script_path="${{ env.OUTPUT_ARTIFACTS_DIR }}/tests/rocgdb/test_rocgdb.py"
-          updated=$(echo "$components" | jq -c --arg s "$script_path" '
-            map(
-              .test_script |= gsub(
-                "build_tools/github_actions/test_executable_scripts/test_rocgdb\\.py";
-                $s
-              )
-            )
-          ')
-          echo "components=$updated" >> $GITHUB_OUTPUT
+          echo "components=$COMPONENTS" >> $GITHUB_OUTPUT
 
   test_components:
     name: "Test ${{ matrix.components.job_name }}"


lumachad:
Blocked on failures fixed by https://github.com/ROCm/ROCgdb/pull/247.

lumachad:
Blocked on https://github.com/ROCm/ROCgdb/pull/261.

lumachad:
I think this is a good first cut. Once we merge https://github.com/ROCm/ROCgdb/pull/296, we should be able to run this one clean.

I'm still unsure if we should always run these or if we should put the extra tests behind a selectable label. Given we want to run these optimization tests, I think we should do it for every ROCgdb PR.

lumachad:
@spatrang Could you please eyeball this one?

lumachad:
> @spatrang Could you please eyeball this one?

@spatrang ping?

.github/workflows/therock-ci.yml
```diff
@@ -138,6 +138,104 @@ jobs:
...
+        --tests gdb.base/attach.exp
+        --tests gdb.base/break.exp
+        --tests gdb.base/callfuncs.exp
+        --tests gdb.base/consecutive-step-over.exp
+        --tests gdb.base/ending-run.exp
+        --tests gdb.base/finish.exp
+        --tests gdb.base/hbreak.exp
+        --tests gdb.base/signals.exp
+        --tests gdb.base/step-test.exp
+        --tests gdb.base/watchpoint.exp
```
spatrang: **Blocker: this runs 1 of the 10 tests, and that one is skipped.**

`--tests` is declared `nargs="+"` with no `action="append"` (`.github/scripts/test_rocgdb.py:419`), so repeated occurrences overwrite each other and only the last survives. Because `--tests` sits in a mutually exclusive group, repeating the *same* option raises no argparse error, so nothing warns you.

From the `hip board / Test rocgdb-gpu` log in run 32959987250:

```
Number of tests to run: 1
Running .../gdb.base/watchpoint.exp ...
# of unsupported tests		1
```

One test selected, and it reports `unsupported` under `--target_board=hip`. So the job is green while exercising zero coverage, and it will stay green regardless of regressions.

Since `nargs="+"` takes a list, collapse into a single flag:

```yaml
      test_options: >-
        --runtestflags=--target_board=hip
        --tests gdb.base/attach.exp gdb.base/break.exp gdb.base/callfuncs.exp
        gdb.base/consecutive-step-over.exp gdb.base/ending-run.exp
        gdb.base/finish.exp gdb.base/hbreak.exp gdb.base/signals.exp
        gdb.base/step-test.exp gdb.base/watchpoint.exp
```

One caveat even after that fix: correctness then relies on argparse last-wins to defeat the `--tests gdb.rocm` that TheRock already puts in the `rocgdb-gpu` `test_script`. That is implicit and fragile, so please make it explicit (strip the pre-existing `--tests` segment in the jq step). Also worth confirming how many of these ten are actually supported on the hip board before spending runner time on them.

.github/workflows/therock-ci.yml
```diff
@@ -138,6 +138,104 @@ jobs:
       artifact_run_id: ${{ github.run_id }}
       projects_to_test: rocgdb-corefile
 
+  # Extended test configurations. These reuse the single build from
+  # therock-build-linux (no rebuild) and fan out in parallel with each other
+  # and with the standard tests, all gated on a successful build. PR-only.
+  extended-test-optimization:
+    name: "--optimization=-O3"
+    if: ${{ github.event_name == 'pull_request' && needs.setup.outputs.enable_therock_ci == 'true' }}
```
spatrang: **The `github.event_name == 'pull_request'` guard is dead code** (repeated on lines 166, 180, 194, 214).

This workflow has no `push:` trigger on `amd-staging`, only `pull_request` and `workflow_dispatch` (lines 3-9). So the "PR-only ... do not run on branch pushes" claim in the comment just above (lines 141-143) describes a case that cannot occur, which misleads the next reader.

It is not merely inert, though: a `workflow_dispatch` run skips all five extended jobs, and since `therock_ci_summary` treats `skipped` as success (line 260), the manual run reports fully green having tested nothing extra. Either drop the condition and let the five inputs stand on their own, or make the exclusion deliberate and documented.

.github/workflows/therock-ci.yml
```diff
@@ -138,6 +138,104 @@ jobs:
       artifact_run_id: ${{ github.run_id }}
       projects_to_test: rocgdb-corefile
 
+  # Extended test configurations. These reuse the single build from
+  # therock-build-linux (no rebuild) and fan out in parallel with each other
+  # and with the standard tests, all gated on a successful build. PR-only.
+  extended-test-optimization:
+    name: "--optimization=-O3"
```
spatrang: Nit: job `name:` values double as required-status-check identifiers, so tuning a flag renames the check and silently breaks any branch-protection rule referencing it. Stable names like `Test GPU (-O3)`, with the exact flags left to the workflow inputs, would be safer.

The `extended-test-*` job-id prefix is also inconsistent with the existing `test-gpu` / `test-cpu` / `test-corefile` naming.

.github/workflows/therock-ci.yml
```diff
@@ -138,6 +138,104 @@ jobs:
...
+    if: ${{ github.event_name == 'pull_request' && needs.setup.outputs.enable_therock_ci == 'true' }}
+    needs: [setup, therock-build-linux]
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        amdgpu_family: ${{ fromJSON(needs.setup.outputs.test_families) }}
```
spatrang, line 158: With the `ci:run-all-archs` label, `test_families` expands to six entries, so this job plus `extended-test-optimization-lto` become 12 GPU jobs on top of the existing `test-gpu` fan-out.

This is the trade-off you raised in the PR discussion. Worth stating the intended per-PR GPU-runner budget in the description, and whether the optimization configs should participate in all-archs runs at all.

.github/workflows/therock-ci.yml
```diff
@@ -138,6 +138,104 @@ jobs:
...
+    uses: ./.github/workflows/therock-test-packages.yml
+    with:
+      amdgpu_families: ${{ matrix.amdgpu_family }}
+      artifact_group: ${{ needs.setup.outputs.artifact_group }}
+      therock_ref: ${{ needs.therock-build-linux.outputs.therock_ref }}
+      artifact_run_id: ${{ github.run_id }}
+      projects_to_test: rocgdb-gpu,rocgdb-corefile
+      base_gpu_test_options: "--toolchain llvm"
```
spatrang: `--toolchain llvm` is already present in TheRock's `test_script` for both `rocgdb-gpu` and `rocgdb-corefile` at the pinned commit `f438028` (see `fetch_test_configurations.py`), so this appends a duplicate. From the actual log:

```
... --toolchain llvm --tests gdb.rocm --parallel -f 0.25 --toolchain llvm --runtestflags=...
```

As far as I can tell this input has no effect anywhere today. Same applies to lines 209 and 225. Consider dropping `base_gpu_test_options` entirely.

.github/workflows/therock-ci.yml
```diff
@@ -138,6 +138,104 @@ jobs:
...
+    needs: [setup, therock-build-linux]
+    permissions:
+      contents: read
+    uses: ./.github/workflows/therock-test-packages.yml
+    with:
+      amdgpu_families: gfx94X
+      artifact_group: ${{ needs.setup.outputs.artifact_group }}
+      therock_ref: ${{ needs.therock-build-linux.outputs.therock_ref }}
+      artifact_run_id: ${{ github.run_id }}
+      test_options: "--check-type check-read1"
```
spatrang: **This job is currently failing.** In run 32959987250, both this configuration and `check-readmore` fail on `rocgdb-gpu` after all four retry iterations:

```
FAIL: gdb.rocm/addr-bp-gpu-no-deb-info.exp: continue until exit
fatal error: Per-queue memory reserved for the debugger is missing
amd_dbgapi_displaced_stepping_start failed (A fatal error has occurred)
```

That it reproduces identically under both `check-read1` and `check-readmore` suggests a real bug or a runner/dbgapi environment problem rather than read-size-dependent flakiness. Merging as-is would turn `amd-staging` CI red for everyone, so this needs a root cause (or an xfail entry) first.

.github/workflows/therock-test-packages.yml
```diff
@@ -13,9 +13,22 @@ on:
         required: true
       artifact_run_id:
         type: string
+      base_test_options:
+        type: string
+        description: 'Options prepended to every component test_script before test_options.'
+        default: '--parallel -f 0.25'
```
spatrang: **These flags are already set upstream.** At the pinned TheRock commit `f438028`, `fetch_test_configurations.py` bakes `--parallel -f 0.25` into every rocgdb component:

```python
"rocgdb-gpu": {
    "test_script": "python ./build/tests/rocgdb/test_rocgdb.py --parallel -f 0.25 --toolchain llvm --tests gdb.rocm",
```

So the real command line becomes (verbatim from the `check-read1` job log):

```
python ./build/tests/rocgdb/test_rocgdb.py --parallel -f 0.25 --toolchain llvm --tests gdb.rocm --parallel -f 0.25 --check-type check-read1
```

Functionally equivalent today since both resolve to the same value, but it is confusing in logs and it silently overrides any future upstream change to the fraction. Either default to `''` and rely on TheRock's value, or get the flags dropped upstream so this workflow clearly owns parallelism.

If this and `base_gpu_test_options` both go, only `test_options` and `projects_to_test` are doing real work, which would be a noticeably smaller and clearer change.

.github/workflows/therock-test-packages.yml
```diff
@@ -28,9 +41,22 @@ on:
         required: true
       artifact_run_id:
         type: string
+      base_test_options:
+        type: string
+        description: 'Options prepended to every component test_script before test_options.'
```
spatrang: Nit: "prepended ... before `test_options`" contradicts the implementation, which appends (line 126). I assume "ordered before" is meant, but read literally it says the opposite.

Suggest: `Options appended to every component test_script, before base_gpu_test_options and test_options.` Same wording on line 18.

.github/workflows/therock-test-packages.yml
```diff
@@ -85,19 +111,21 @@ jobs:
         id: configure
         run: python ./build_tools/github_actions/fetch_test_configurations.py
 
-      - name: Override rocgdb test_script to use artifact-installed script
-        # TheRock's fetch_test_configurations.py points test_rocgdb.py at its own
-        # build_tools copy. Rewrite to the script installed in the artifacts at
-        # OUTPUT_ARTIFACTS_DIR/tests/rocgdb/test_rocgdb.py.
+      - name: Patch rocgdb test_script options
+        # Append caller-supplied options to each component's test_script.
         id: patch-scripts
```
spatrang, line 102: Deleting the `gsub` path rewrite is correct at the current pin. TheRock now emits `python ./build/tests/rocgdb/test_rocgdb.py ...` directly, so the rewrite was already a no-op.

That said, it is a distinct behavioral change riding inside a commit titled "Run extended ROCgdb tests in parallel on PRs", and it removes a safety net that kept this workflow independent of TheRock's path choice. Please split it into its own commit, or at least note in the message why it is now safe.

Follow-on: `OUTPUT_ARTIFACTS_DIR` (line 66) was consumed only by the deleted step and is now unreferenced. Worth removing in the same change.

.github/workflows/therock-test-packages.yml
```diff
@@ -85,19 +111,21 @@ jobs:
...
+        env:
+          BASE_TEST_OPTIONS: ${{ inputs.base_test_options }}
+          BASE_GPU_TEST_OPTIONS: ${{ inputs.base_gpu_test_options }}
+          TEST_OPTIONS: ${{ inputs.test_options }}
         run: |
           components='${{ steps.configure.outputs.components }}'
```
spatrang: Nit: the three new option strings correctly go through `env:` to avoid injection, which is the right call, so it is a little inconsistent to leave `components` interpolated directly into a single-quoted shell string here.

It is pre-existing and the value is machine-generated JSON, so the risk is low, but moving it to `env:` too would make the step uniformly safe and match the pattern this PR establishes.

.github/workflows/therock-test-packages.yml
```diff
@@ -85,19 +111,21 @@ jobs:
...
-          script_path="${{ env.OUTPUT_ARTIFACTS_DIR }}/tests/rocgdb/test_rocgdb.py"
-          updated=$(echo "$components" | jq -c --arg s "$script_path" '
+          updated=$(echo "$components" | jq -c --arg base "$BASE_TEST_OPTIONS" --arg gpu "$BASE_GPU_TEST_OPTIONS" --arg opts "$TEST_OPTIONS" '
             map(
-              .test_script |= gsub(
-                "build_tools/github_actions/test_executable_scripts/test_rocgdb\\.py";
-                $s
+              .test_script |= (
+                  if $base != "" then . + " " + $base else . end
+                | if $gpu != "" then . + " " + $gpu else . end
```
spatrang: `base_gpu_test_options` is documented as applying "for GPU-targeted jobs" (lines 22 and 50), but there is no GPU predicate here: it is appended to every component in the matrix unconditionally.

It only looks GPU-scoped because the callers that set it happen to pass `projects_to_test: rocgdb-gpu,...`. Set it on a job that includes `rocgdb-cpu` and it lands there too. Either rename it to something honest like `extra_test_options`, or actually gate it on `matrix.components.linux_cpu_runner != true`.

spatrang (commented):
Reviewed at head `a06e86a`. Good direction and the mechanism is sound, but there is one blocker plus some redundancy against options TheRock already sets. Details inline.

Highlights:
- **Blocker:** the `hip board` job runs 1 of its 10 tests, and that one reports `unsupported`, so it passes while testing nothing.
- **CI is red:** `check-read1` and `check-readmore` both fail on `gdb.rocm/addr-bp-gpu-no-deb-info.exp`.
- `base_test_options` and `base_gpu_test_options` duplicate `--parallel -f 0.25 --toolchain llvm`, which TheRock already puts in every rocgdb `test_script`.

Not line-anchorable: the PR description is stale. It lists six configurations (there are five, since `--parallel` became the shared default) and says the hip board job runs only `gdb.base/break.exp` when the code lists ten.

.github/workflows/therock-ci.yml
lumachad: Very obviously broken. My bad, let me fix this.

lumachad: Fixed. Collapsed to a single `--tests` flag. Also made the override explicit: the `patch-scripts` step now detects when `test_options` contains `--tests` and strips any pre-existing `--tests` from the upstream `test_script` before appending, so the behaviour no longer relies on argparse last-wins.

.github/workflows/therock-ci.yml
lumachad: Fixed. Dropped the `github.event_name == 'pull_request'` guard from all five extended jobs. The `enable_therock_ci == 'true'` condition is sufficient.

.github/workflows/therock-ci.yml
lumachad: Fixed. Renamed to stable display names (`Test GPU (-O3)`, `Test GPU (-O3 -flto)`, `Test GPU (check-read1)`, `Test GPU (check-readmore)`, `Test GPU (hip board)`) and aligned job IDs to the existing naming scheme (`test-gpu-o3`, `test-gpu-lto`, `test-gpu-check-read1`, `test-gpu-check-readmore`, `test-gpu-hip-board`).

.github/workflows/therock-ci.yml
lumachad: Acknowledged and documented. The two optimization jobs intentionally participate in the all-archs fan-out -- the point is to catch compiler-specific failures across all supported architectures. Added a comment to the workflow noting that each adds up to six GPU jobs with `ci:run-all-archs`, and updated the PR description with the per-PR budget.

.github/workflows/therock-ci.yml
lumachad: Fixed. Dropped `base_gpu_test_options` entirely from both files -- it only ever carried `--toolchain llvm` which TheRock already sets.

.github/workflows/therock-ci.yml
lumachad: These are currently expected to fail due to a runner issues.

.github/workflows/therock-test-packages.yml
lumachad: Fixed. Dropped `base_test_options` entirely (same rationale as `base_gpu_test_options`): the default `--parallel -f 0.25` was already baked into TheRock's `test_script`. The only remaining appending input is `test_options`.

.github/workflows/therock-test-packages.yml
lumachad: Moot now that `base_test_options` is dropped entirely.

.github/workflows/therock-test-packages.yml
lumachad: Split into its own commit (`[workflows] Drop now-redundant test_rocgdb.py path rewrite`) as suggested. `OUTPUT_ARTIFACTS_DIR` removed in the same commit.

.github/workflows/therock-test-packages.yml
lumachad: Fixed. `components` is now passed via `COMPONENTS` in `env:`, making the step uniformly injection-safe.

.github/workflows/therock-test-packages.yml
lumachad: Fixed by dropping `base_gpu_test_options` entirely -- the naming/scoping ambiguity goes away with it.

lumachad:
Updated with the fixes suggested. I'll let CI run again.

.github/workflows/therock-test-packages.yml
```diff
@@ -85,19 +94,28 @@ jobs:
...
+          fi
+          updated=$(echo "$COMPONENTS" | jq -c \
+            --arg opts "$TEST_OPTIONS" \
+            --argjson strip "$strip_tests" '
             map(
-              .test_script |= gsub(
-                "build_tools/github_actions/test_executable_scripts/test_rocgdb\\.py";
-                $s
+              .test_script |= (
+                if $strip then gsub("--tests \\S+"; "") else . end
```
spatrang: **This strips only the first test name, not the whole list.**

`gsub("--tests \\S+"; "")` consumes `--tests` plus exactly one token. That is fine for a component whose upstream `--tests` carries a single value, but `rocgdb-corefile` upstream carries five. Replaying this jq against the real component JSON:

```
python ./build/tests/rocgdb/test_rocgdb.py --parallel -f 0.25 --toolchain llvm gdb.rocm/core-no-read-special-files.exp gdb.rocm/gcore-after-attach.exp --runtestflags=--target_board=hip --tests gdb.base/attach.exp ...
```

The leftover test names become bare positional arguments. `test_rocgdb.py` calls strict `parser.parse_args()` (line 650) and defines no positionals, so this exits 2 with "unrecognized arguments".

It does not trigger today, purely because the only caller passing `--tests` is the hip board job and it is scoped to `projects_to_test: rocgdb-gpu`, whose upstream `--tests gdb.rocm` happens to be one token. Add `rocgdb-corefile` to that job, or let TheRock expand the `rocgdb-gpu` list, and it breaks — and it breaks in the matrix-configure output rather than anywhere obvious.

Making the regex consume the full list fixes both cases:

```
gsub("--tests(\\s+[^-]\\S*)*"; "")
```

I verified this against both the single-value (`rocgdb-gpu`) and multi-value (`rocgdb-corefile`) scripts.

Two small things while you are in here:
- The guard on line 108, `[[ "$TEST_OPTIONS" == *"--tests"* ]]`, also fires on any token merely containing the string (`--tests-foo`). A word-boundary form such as `[[ " $TEST_OPTIONS " == *" --tests "* ]]` would be tighter.
- `rtrimstr(" ")` on line 118 removes at most one trailing space. It is correct only because the space-collapsing `gsub` runs immediately before it, which is subtle enough that `sub("\\s+$"; "")` would be self-evidently right.

.github/workflows/therock-ci.yml
```diff
@@ -147,6 +241,11 @@ jobs:
       - test-gpu
       - test-cpu
       - test-corefile
+      - test-gpu-o3
+      - test-gpu-lto
+      - test-gpu-check-read1
+      - test-gpu-check-readmore
```
spatrang, line 247: **Known-failing jobs listed here will leave a permanently red required check.**

Re your reply that `check-read1` / `check-readmore` are expected to fail from a pre-existing dbgapi bug and are "left in to keep the coverage visible": that intent does not survive contact with this job. `therock_ci_summary` fails on any dependency whose result is neither `success` nor `skipped` (line 256), so adding these two to `needs` means the summary fails on **every** ROCgdb PR from the moment this merges, regardless of what the PR changes.

That is the mirror image of the original hip board problem. There, green meant nothing; here, red would mean nothing. Once the summary is red by default, people stop reading it, and the genuine failures it is supposed to surface get lost.

Keeping the coverage visible is a good goal, but it needs a mechanism that distinguishes "known-broken" from "newly broken". Any of these would do:

- Add `gdb.rocm/addr-bp-gpu-no-deb-info.exp: continue until exit` to the ignore list, so the jobs pass and a *new* failure still fails them. This is my preference, since it keeps the jobs load-bearing.
- Set `continue-on-error: true` on the two jobs, so they report without gating the summary.
- Leave them out of the summary `needs` until the dbgapi root cause is fixed.

Whichever you pick, please reference the dbgapi issue in a comment so the exemption can be removed once it is resolved. As it stands I do not think this should merge — the first PR to land on top would be blamed for a red summary it did not cause.

spatrang (commented):
Second round, reviewed at head `94425ff`. Thanks for the thorough turnaround — the blocker is genuinely fixed, and dropping `base_test_options` / `base_gpu_test_options` outright was better than what I asked for, since it also reverts the unintended change to the three pre-existing test jobs.

I verified the new `patch-scripts` step against the real component JSON and the hip board case now produces a single `--tests` with the full list. I also confirmed that removing `OUTPUT_ARTIFACTS_DIR` is safe: `therock-test-component.yml` defines its own copy (line 42) and passes it to the test step (line 71), and workflow-level `env:` never propagated into the called workflow anyway.

Two remaining items inline, both of which I think should block merge.

.github/workflows/therock-ci.yml
lumachad: That's fine. We currently deal with flaky runners and we just need to re-trigger these failing jobs until they pass.

lumachad: It isn't actually a dbgapi issue, but it comes up as a dbgapi error. But it is caused by runner driver/GPU corruption. Still undiagnosed.

.github/workflows/therock-test-packages.yml
lumachad: Fixed all three:
- Pattern changed to `gsub("--tests(\\s+[^-]\\S*)*"; "")` to consume the full argument list.
- Guard tightened to `[[ " \$TEST_OPTIONS " == *" --tests "* ]]` to avoid matching substrings like `--tests-foo`.
- `rtrimstr(" ")` replaced with `sub("\\s+\$"; "")` to make the trailing-space trim self-evidently correct.

lumachad: Still, this patching mechanism is a bit fragile. But we don't have a better interface into TheRock's.

.github/workflows/therock-ci.yml
```diff
@@ -138,6 +138,100 @@ jobs:
...
+      artifact_group: ${{ needs.setup.outputs.artifact_group }}
+      therock_ref: ${{ needs.therock-build-linux.outputs.therock_ref }}
+      artifact_run_id: ${{ github.run_id }}
+      projects_to_test: rocgdb-gpu
+      test_options: >-
+        --runtestflags=--target_board=hip
+        --tests gdb.base/attach.exp gdb.base/start.exp gdb.base/callfuncs.exp
+        gdb.base/consecutive-step-over.exp gdb.base/ending-run.exp
+        gdb.base/finish.exp gdb.base/hbreak.exp gdb.base/signals.exp
+        gdb.base/step-test.exp gdb.base/watchpoint.exp
```
spatrang, line 233: Non-blocking, but worth recording somewhere: `gdb.base/break.exp` was swapped out for `gdb.base/start.exp` in the last push without a note in the commit message or the discussion.

The reason is visible in the previous run's log — `break.exp` accounted for **all 26** failures in this job, was the only failing test of the ten, and failed persistently across all four retry iterations:

```
130 FAIL: gdb.base/break.exp
  test_break: run until function breakpoint
  test_break: break lineno
  test_break_user_call: hit breakpoint on called function
  ... 23 more
```

So this job is green partly because the failing test was removed rather than because it now passes. That may well be the correct call: `break.exp` leans on plain C program flow and much of it may simply not apply under `--target_board=hip`. But `break.exp` is the most substantive generic breakpoint test in this set and the one originally named in the PR description, so quietly dropping it makes the job look like broader coverage than it is.

Could you record which it is? Either:
- these 26 are genuine ROCgdb bugs under the hip board and deserve an issue to link here, or
- they are expected-under-hip, in which case an ignore-list entry would let `break.exp` stay in the list and still catch *new* regressions, which seems strictly better than omitting it.

A one-line comment next to the test list explaining why `break.exp` is absent would also be enough to stop the next reader re-adding it and rediscovering this.

spatrang (approved):
LGTM — approving at `1282d323`. Both blocking items are resolved, and I verified them rather than relying on the replies.

**Verified the patching logic.** I replayed the new `patch-scripts` step against all three real rocgdb components across five `test_options` cases (empty, `--check-type`, `--optimization=-O3`, `--optimization='-O3 -flto'`, and the hip board case), then fed every resulting command line through a mirror of `test_rocgdb.py`'s parser. All parse cleanly:

- Empty `test_options` is a proper no-op, so the three standard jobs are untouched.
- `rocgdb-corefile` keeps its full five-file `--tests` list when the caller does not pass `--tests`.
- The hip board case strips the upstream `--tests` completely, including the multi-value corefile form that broke the previous regex, and lands a single `--tests` with all ten files.

**Verified on hardware.** Run 33069089574 is green across all 26 jobs including `TheRock CI Summary`. The hip board job now reports:

```
Number of tests to run: 10
# of expected passes		38
# of unsupported tests		3
[✓] OVERALL STATUS: PASS
```

Up from the single skipped test that started this review. `check-read1` and `check-readmore` also passed across all three components in this run, which supports the flaky-runner explanation over a systematic failure.

Thanks for the fast and thorough turnaround, particularly for dropping `base_test_options` / `base_gpu_test_options` outright instead of just documenting them — that removed the unintended behaviour change to the three pre-existing jobs and made the whole diff smaller.

One non-blocking question inline about the `break.exp` swap, plus two cosmetic notes below. None of these need to hold the merge.

**Cosmetic, for a follow-up whenever convenient:**
- `Test GPU (check-read1)` and `Test GPU (check-readmore)` do not set `projects_to_test`, so they also run `rocgdb-cpu` and `rocgdb-corefile` — visible in the run as `Test GPU (check-read1) / Test rocgdb-cpu`. The names claim GPU. Either scope them to `rocgdb-gpu` or rename to `Test check-read1`. This one is partly my fault for suggesting the `Test GPU (...)` pattern.
- The description says the hip board job uses `rocgdb-gpu,rocgdb-corefile` (the code uses `rocgdb-gpu` only) and that the workflow "gains two new inputs", though `projects_to_test` already existed and only gained a description.

.github/workflows/therock-ci.yml
lumachad: hip-board is not meant to pass all the tests, so we only include a handful of the most basic tests to make sure we don't regress things.

lumachad:
> LGTM — approving at `1282d323`. Both blocking items are resolved, and I verified them rather than relying on the replies.
> 
> **Verified the patching logic.** I replayed the new `patch-scripts` step against all three real rocgdb components across five `test_options` cases (empty, `--check-type`, `--optimization=-O3`, `--optimization='-O3 -flto'`, and the hip board case), then fed every resulting command line through a mirror of `test_rocgdb.py`'s parser. All parse cleanly:
> 
> * Empty `test_options` is a proper no-op, so the three standard jobs are untouched.
> * `rocgdb-corefile` keeps its full five-file `--tests` list when the caller does not pass `--tests`.
> * The hip board case strips the upstream `--tests` completely, including the multi-value corefile form that broke the previous regex, and lands a single `--tests` with all ten files.
> 
> **Verified on hardware.** Run 33069089574 is green across all 26 jobs including `TheRock CI Summary`. The hip board job now reports:
> 
> ```
> Number of tests to run: 10
> # of expected passes		38
> # of unsupported tests		3
> [✓] OVERALL STATUS: PASS
> ```
> 
> Up from the single skipped test that started this review. `check-read1` and `check-readmore` also passed across all three components in this run, which supports the flaky-runner explanation over a systematic failure.
> 
> Thanks for the fast and thorough turnaround, particularly for dropping `base_test_options` / `base_gpu_test_options` outright instead of just documenting them — that removed the unintended behaviour change to the three pre-existing jobs and made the whole diff smaller.
> 
> One non-blocking question inline about the `break.exp` swap, plus two cosmetic notes below. None of these need to hold the merge.
> 
> **Cosmetic, for a follow-up whenever convenient:**
> 
> * `Test GPU (check-read1)` and `Test GPU (check-readmore)` do not set `projects_to_test`, so they also run `rocgdb-cpu` and `rocgdb-corefile` — visible in the run as `Test GPU (check-read1) / Test rocgdb-cpu`. The names claim GPU. Either scope them to `rocgdb-gpu` or rename to `Test check-read1`. This one is partly my fault for suggesting the `Test GPU (...)` pattern.
> * The description says the hip board job uses `rocgdb-gpu,rocgdb-corefile` (the code uses `rocgdb-gpu` only) and that the workflow "gains two new inputs", though `projects_to_test` already existed and only gained a description.

Addressed these cosmetic issues as well in a new push.

PR #246, opened 2026-08-04
[ROCgdb] Update ignore list: remove unused gdb.rocm entries

lumachad:
Remove 4 gdb.rocm test entries from the Generic ignore list that are
no longer failing and were reported as unused by the CI results from
2026-08-04.


lancesix (approved):
LGTM, thanks for the cleanup.

PR #247, opened 2026-08-04
[gdb/testsuite] Fix gdb.rocm tests for optimized builds

lumachad:
## Summary

- `alu-exceptions.exp`: The `precise=off` variant single-steps through
  device code to find the exception location, which is not reliable on
  optimized device code. Add `require no_optimized_code` to skip for
  optimized builds, consistent with other gdb.rocm tests.

- `hip-catch-errors.cpp`: The break marker "Break after reference
  initialization" was a comment with no associated code, causing the
  breakpoint to miss under -O3. Add `asm volatile ("" ::: "memory")`
  to anchor it. Also make `hiperr_params_ref::no` volatile to prevent
  the optimizer from keeping the value in a register, which caused GDB
  to read 0.

- `hip-catch-errors.exp`: `__device_stub__kernel` is inlined into its
  caller at -O3 and disappears from the backtrace. Drop it from the
  required backtrace pattern; `hipLaunchKernel` being present is
  sufficient.

## Test plan

- [x] Run `gdb.rocm/alu-exceptions.exp` with an optimized build and confirm it is skipped
- [x] Run `gdb.rocm/hip-catch-errors.exp` with an optimized build and confirm all tests pass

### DIFF
--- a/gdb/testsuite/gdb.rocm/alu-exceptions.exp
+++ b/gdb/testsuite/gdb.rocm/alu-exceptions.exp
@@ -23,6 +23,11 @@ load_lib rocm.exp
 
 require allow_hip_tests
 
+# With optimized device code, the compiler may promote integer division to
+# floating-point, so enabling the integer divide-by-zero exception mode does
+# not trigger a SIGFPE.  Skip for optimized builds.
+require no_optimized_code
+
 standard_testfile .cpp
 
 if {[build_executable "failed to prepare $testfile" $testfile $srcfile \
--- a/gdb/testsuite/gdb.rocm/hip-catch-errors.cpp
+++ b/gdb/testsuite/gdb.rocm/hip-catch-errors.cpp
@@ -26,7 +26,7 @@ static bool gen_ref = false;
 
 struct hiperr_params_ref
 {
-  int no;
+  volatile int no;  /* Prevent the optimizer from keeping this in a register.  */
   const char *name;
   const char *str;
 
@@ -130,7 +130,7 @@ main (int argc, const char **argv)
       hip_set_device ();
       hip_get_device ();
       hip_launch_kernel ();
-      /* Break after reference initialization.  */
+      asm volatile ("" ::: "memory"); /* Break after reference initialization.  */
     }
   else if (test == "one")
     hip_set_device ();
--- a/gdb/testsuite/gdb.rocm/hip-catch-errors.exp
+++ b/gdb/testsuite/gdb.rocm/hip-catch-errors.exp
@@ -252,8 +252,10 @@ proc do_test {} {
 	    gdb_test_no_output "set args launch"
 	    catch_hiperr
 	    check_hip_launch_kernel_err "run"
+	    # __device_stub__kernel is inlined into hip_launch_kernel at
+	    # higher optimization levels and may not appear in the backtrace.
 	    gdb_test "backtrace" \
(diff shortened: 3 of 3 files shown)


lumachad:
@lancesix @amd-shahab ping?

gdb/testsuite/gdb.rocm/alu-exceptions.exp
```diff
@@ -23,6 +23,10 @@ load_lib rocm.exp
 
 require allow_hip_tests
 
+# Single-stepping through device code to find the exception location is not
+# reliable on optimized device code.  Skip for optimized builds.
+require no_optimized_code
```
lancesix, line 29: Can you elaborate on what is not reliable?

I am not really opposed to the change, as what this test does does not rely on debug info.

lumachad: From what I recall, we keep instruction-stepping but never reach the expected exception within the timeout allocated, or at all.

lancesix: did we get the exception if running without single stepping?

It fells odd to have something different depending on stepping or not.

lumachad: > did we get the exception if running without single stepping?
> 
> It fells odd to have something different depending on stepping or not.

No. It's actually just int_div0. For some reason the thread simply exits cleanly without hitting the exception, regardless of "continue" or "si".

```
FAIL: gdb.rocm/alu-exceptions.exp: int_div0, precise=off: step until exception
FAIL: gdb.rocm/alu-exceptions.exp: int_div0, precise=off: stopped after fault
FAIL: gdb.rocm/alu-exceptions.exp: int_div0, precise=off: capture exception_location
FAIL: gdb.rocm/alu-exceptions.exp: int_div0, precise=off: pass exception to runtime

                === gdb Summary ===

# of expected passes            89
# of unexpected failures        4
# of unsupported tests          7
```

lancesix: I had a quick look on gfx1031, and on that tests, it seems the compiler promotes the `int` into `float`.

The instruction which should raise is:

```
v_rcp_f32_e32 v3, v3
```

but because this test enabled `int_div0` it does not catch it.  Enabling the float div by 0 raises the SIGFPE.

Can you update the comment to reflect that the issue is not single stepping, but this float to int promotion with optimisation?  With this, things look good to me.

lumachad: Do we want to take measures to make this work instead of skipping this particular test when optimization is used?

lancesix: I thought about this for a tiny bit, but I am not sure how we can prevent the compiler from making such optimisation.

But as said before, this particular test is exercising HW feature more than anything based on the debug info.  Said differently, I do not think we would expect to catch any regression running with compiler optimisations, so I am fine saying this is a -O0 test only.

lumachad: Fair enough. Thanks for clarifying. The comment now reads:

```
# With optimized device code, the compiler may promote integer division to
# floating-point, so enabling the integer divide-by-zero exception mode does
# not trigger a SIGFPE.  Skip for optimized builds.
```

lancesix (approved):
Seems reasonable to me.  Thanks.

PR #249, opened 2026-08-05
gdb.rocm/runtime-core.exp: Fix unreliable wave identification

lumachad:
## Summary
- The "identify waves" block used two separate patterns anchored with ^
  to match GPU wave threads from "info thread" output, which fails when
  the entire output arrives as a single chunk (^ only anchors to the
  buffer start, not individual lines).
- Even without the anchor, two competing patterns cause the first-listed
  one to greedily consume past the line intended for the second.
- Fix by using a single pattern matching any AMDGPU Wave line, then
  classifying in Tcl using the leading "*" (selected/faulty wave) and
  the quoted thread name (aux_kernel). Robust regardless of output
  chunking.
- Also removes the now-unused fault_loc variable, simplifies the gpusig
  conditional, and fixes a typo in the do_test comment.

## Test plan
- [x] Verify passes on mi300x (gfx942)
- [x] Verify passes on mi300a (gfx942)
- [x] Verify passes on gfx1201

### DIFF
--- a/gdb/testsuite/gdb.rocm/runtime-core.exp
+++ b/gdb/testsuite/gdb.rocm/runtime-core.exp
@@ -50,7 +50,7 @@ set success_count 0
 #   Run the test exercising generation of a core dump on fault FAULT
 #   using a file or a pipe.
 #
-# FAULT is one of "page", "abort",  or "assert", and selects which
+# FAULT is one of "pagefault", "abort", or "assert", and selects which
 #   fault the application should generate.
 #
 # CORE_TYPE is one of "full" or "lightweight".
@@ -79,13 +79,8 @@ proc do_test { fault core_type output_type } {
 
     if {$fault == "pagefault"} {
 	set gpusig "SIGSEGV"
-	set fault_loc "pagefault_kernel"
-    } elseif {$fault == "abort"} {
+    } else {
 	set gpusig "SIGABRT"
-	set fault_loc "abort"
-    } elseif {$fault == "assert"} {
-	set gpusig "SIGABRT"
-	set fault_loc "__assert_fail"
     }
 
     clean_restart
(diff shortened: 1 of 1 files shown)


gdb/testsuite/gdb.rocm/runtime-core.exp
```diff
@@ -100,16 +95,24 @@ proc do_test { fault core_type output_type } {
...
+	-re "(\\\*?) *($::decimal) *AMDGPU Wave \[^\r\n\]* \"(\[^\"]+)\" \[^\r\n\]*\r\n" {
+	    if {$expect_out(1,string) eq "*"} {
+		set faulty_wave $expect_out(2,string)
+	    } elseif {$expect_out(3,string) eq "aux_kernel"} {
+		set aux_wave $expect_out(2,string)
+	    }
 	    exp_continue
 	}
-	-re "^\r\n$::gdb_prompt $" {
+	-re "$::gdb_prompt $" {
```
czidev-amd, line 115: I think you can just use \"\"

gdb/testsuite/gdb.rocm/runtime-core.exp
```diff
@@ -100,16 +95,24 @@ proc do_test { fault core_type output_type } {
...
+    # Use a single pattern to match any AMDGPU Wave line regardless of which
+    # kernel it belongs to.  Using two separate patterns (one per kernel name)
+    # is unreliable because the entire "info thread" output can arrive as one
+    # chunk, causing a greedy pattern listed first to consume past the line
+    # intended for the second pattern.  A single pattern sidesteps that by
+    # iterating over all AMDGPU Wave lines with exp_continue and classifying
+    # each in Tcl.  The leading "*" identifies the selected (faulty) wave, and
+    # the quoted thread name identifies the auxiliary wave.
+    gdb_test_multiple "info thread" "identify waves" {
+	-re "(\\\*?) *($::decimal) *AMDGPU Wave \[^\r\n\]* \"(\[^\"]+)\" \[^\r\n\]*\r\n" {
```
lancesix, line 107: shouldn't the pattern for the function name between quotes (`\"(\[^\"]+)\"`) also exclude \r\n?

i.e. `\"(\[^\r\n\"]+)\"`

otherwise it seems it would be valid to match from the start of a function until the end of the next one.

lumachad: I'm not sure I see that. Do you have a concrete example of the expression you're trying to avoid?

gdb/testsuite/gdb.rocm/runtime-core.exp
lumachad: Do you mean -wrap -re ""?

gdb/testsuite/gdb.rocm/runtime-core.exp
lancesix: actually, we would stop on the fst ", and no symbol should have \r\n anyway, so should be good.  Sorry, my bad.

gdb/testsuite/gdb.rocm/runtime-core.exp
lumachad: FTR, I don't think -wrap will work here, as it forces a newline pattern at the beginning, and that's what we're removing.

czidev-amd: Understood. lgtm

lumachad:
@lancesix @czidev-amd If this looks OK, could I get an approval so we can move on with https://github.com/ROCm/TheRock/pull/6378?

czidev-amd (approved):
LGTM

PR #250, opened 2026-08-05
gdb, testsuite: add gdb.rocm/aspace-nullptr.exp to test nullptr access

aktemur:
Add a new test to check accessing nullptr in various address spaces.

### DIFF
--- a/gdb/testsuite/gdb.rocm/aspace-nullptr.cpp
+++ b/gdb/testsuite/gdb.rocm/aspace-nullptr.cpp
@@ -1,0 +1,48 @@
+/* This testcase is part of GDB, the GNU debugger.
+
+   Copyright (C) 2026 Free Software Foundation, Inc.
+   Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.  */
+
+#include <hip/hip_runtime.h>
+#include "gdb_watchdog.h"
+#include "rocm-test-utils.h"
+
+__global__ void
+kern ()
+{
+  /* Define an LDS (i.e. local address space) variable to make sure
+     the address space is allocated.  */
+  __shared__ int local_var;
+  if (threadIdx.x == 0)
+    local_var = 42;
+
+  NOP (1); /* Break here.  */
+}
+
+int
+main()
+{
+  /* Make sure that if anything goes wrong, the program eventually
+     gets killed.  */
+  gdb_watchdog (30);
+
+  /* Make dimensions large enough to create workgroups with multiple
+     waves.  */
+  kern<<<2, 320>>> ();
+  CHECK (hipDeviceSynchronize ());
+
+  return 0;
+}
--- a/gdb/testsuite/gdb.rocm/aspace-nullptr.exp
+++ b/gdb/testsuite/gdb.rocm/aspace-nullptr.exp
@@ -1,0 +1,72 @@
+# Copyright (C) 2026 Free Software Foundation, Inc.
+# Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+# This file is part of GDB.
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test accessing memory through nullptr for various address spaces.
+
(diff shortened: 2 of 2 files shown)


aktemur:
dbgapi companion PR: https://github.com/ROCm/rocm-systems/pull/9707

gdb/testsuite/gdb.rocm/aspace-nullptr.exp
```diff
@@ -0,0 +1,48 @@
...
+    if {![runto_main]} {
+	return
+    }
+    set bp_line [gdb_get_line_number "Break here"]
+    gdb_breakpoint $bp_line -allow-pending -temporary
+    gdb_continue_to_breakpoint "kernel bp"
+
+    # Access nullptr in several address spaces.
+    foreach_with_prefix addr {0x0 generic#0x0 private_lane#0xffffffff \
+			      private_wave#0xffffffff local#0xffffffff} {
```
lancesix, line 44: I am not sure we can assume this in GDB.  Ideally, this should wait until #218 lands in, and use `maint print address-spaces` to enumerate the address spaces and associated null ptr addresses.

gdb/testsuite/gdb.rocm/aspace-nullptr.cpp
```diff
@@ -0,0 +1,39 @@
...
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.  */
+
+#include <hip/hip_runtime.h>
+#include "gdb_watchdog.h"
+#include "rocm-test-utils.h"
+
+__global__ void
+kern ()
+{
+  NOP (1); /* Break here.  */
```
lancesix, line 26: might be worth forcing some LDS variable to be sure some LDS is allocated.  This way we really test the null address, not the fact that dbgapi detects this address space as no data at all.

aktemur:
Update above is only a rebase.

gdb/testsuite/gdb.rocm/aspace-nullptr.cpp
aktemur: > might be worth forcing some LDS variable to be sure some LDS is allocated. This way we really test the null address, not the fact that dbgapi detects this address space as no data at all.

Done.

gdb/testsuite/gdb.rocm/aspace-nullptr.exp
aktemur: > I am not sure we can assume this in GDB. Ideally, this should wait until #218 lands in, and use `maint print address-spaces` to enumerate the address spaces and associated null ptr addresses.

Done.

gdb/testsuite/gdb.rocm/aspace-nullptr.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+  NOP (1); /* Break here.  */
+}
+
+int
+main()
+{
+  /* Make sure that if anything goes wrong, the program eventually
+     gets killed.  */
+  gdb_watchdog (30);
```
lancesix, line 40: Is this needed?  I do not really see the program having an infinite loop or anything that would prevent forward progress.

This seems unnecessary (and often annoying when building manually testcases, as pretty quickly in the debug session, the program just terminates unless we explicitly ask GDB to ignore SIGALRM).

gdb/testsuite/gdb.rocm/aspace-nullptr.cpp
lancesix: usually, I would prefer having any inline assembly, unless there is on other way around it.

It is unlikely this macro will be incorrect for future architecture any time soon, but having something purely in HIP avoids this question entirely.  Something like `volatile int foo = 42` saves the same purpose, but I kind of expect none of this is really necessary, as we could set the breakpoint on kernel entry or even on the `if`.

gdb/testsuite/gdb.rocm/aspace-nullptr.exp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+clean_restart $::testfile
+
+with_rocm_gpu_lock {
+    if {![runto_main]} {
+	return
+    }
+    set bp_line [gdb_get_line_number "Break here"]
+    gdb_breakpoint $bp_line -allow-pending -temporary
```
lancesix, line 39: As far as I can see, could just break at `kern` and have the same effect.  No need for a breakpoint by line.

gdb/testsuite/gdb.rocm/aspace-nullptr.exp
```diff
@@ -0,0 +1,72 @@
...
+
+    set trailer "\[^\r\n\]*(?=\r\n)"
+    set tok "\[^ \t\r\n\]+"
+    set sp "\[ \t\]+"
+
+    # Collect address spaces and their nullptrs into a dictionary.
+    set aspaces [dict create]
+
+    gdb_test_multiple "maint print address-spaces" "" {
+	-re "Name$trailer" {
```
lancesix, line 50: could as well test that `maint print address-space` gives us the colums we are looking for, so we can detect early if GDB changes in a way which breaks this test

```
 -re "Name${sp}Fiald2${sp}..." {
    exp_continue
  }
  -re "Name${trailer}" {
    # indicates the first pattern failed, the header changed, so we can fail the test early
    verbose -log "columns changed"
    fail $gdb_test_name
  }
```

czidev-amd (approved):
Once solved Lancelot questions, I have nothing to add.

lumachad:
It seems we're hitting timeouts with the new test.

```
2026-08-13 17:22:16,582 - INFO: print {int}private_wave#0xffffffff
2026-08-13 17:22:16,582 - INFO: FAIL: gdb.rocm/aspace-nullptr.exp: access nullptr in aspace private_wave (timeout)
2026-08-13 17:22:16,582 - INFO: print {int}private_lane#0xffffffff
2026-08-13 17:22:16,582 - INFO: FAIL: gdb.rocm/aspace-nullptr.exp: access nullptr in aspace private_lane (timeout)
2026-08-13 17:22:16,582 - INFO: print {int}local#0xffffffff
2026-08-13 17:22:16,582 - INFO: FAIL: gdb.rocm/aspace-nullptr.exp: access nullptr in aspace local (timeout)
```

lumachad:
Could we please make sure this runs clean when O1~O3 and/or flto is used?

lumachad:
Could you please make sure this is clean when using O1~O3 and/or -flto in the testsuite?

PR #251, opened 2026-08-06
Merge remote-tracking branch 'origin/master' into amd-staging 20260806

aktemur:
There was one conflict which was resolved easily.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -147,3 +147,8 @@ repos:
         language: unsupported_script
         entry: gdb/contrib/check-file-mode.sh
         files: *gdb_files
+
+# Local Variables:
+# indent-tabs-mode: nil
+# tab-width: 2
+# End:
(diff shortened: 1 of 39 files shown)


lumachad:
General comment. We should either change the comment or add an exception to the pre-commit check so we don't keep hitting this renderD -> rendered spurious suggestion.

lancesix (approved):
Merge resolution looks good to me.  If CI is OK, all good to go.

lancesix:
> General comment. We should either change the comment or add an exception to the pre-commit check so we don't keep hitting this renderD -> rendered spurious suggestion.

Can't we add this exception in `gdb/contrib/codespell-ignore-words.txt`?

lumachad:
> > General comment. We should either change the comment or add an exception to the pre-commit check so we don't keep hitting this renderD -> rendered spurious suggestion.
> 
> Can't we add this exception in `gdb/contrib/codespell-ignore-words.txt`?

Probably. That's only downstream. I'll open a PR.

lumachad:
> > > General comment. We should either change the comment or add an exception to the pre-commit check so we don't keep hitting this renderD -> rendered spurious suggestion.
> > 
> > 
> > Can't we add this exception in `gdb/contrib/codespell-ignore-words.txt`?
> 
> Probably. That's only downstream. I'll open a PR.

https://github.com/ROCm/ROCgdb/pull/254

PR #254, opened 2026-08-06
Fix typos and add a codespell exception

lumachad:
Fix a couple typos caught by codespell and except a third one that is just the name pattern of AMD GPU device files.


lancesix (approved):
LGTM, thanks.

aktemur (approved):

PR #255, opened 2026-08-06
recognize DW_AT_address_space and use in pointers

aktemur:
## Motivation

Recognize the `DW_AT_address_space` attribute for pointer types. Use this information to dereference, assign, cast, etc. pointers.

## Technical Details

Add an attribute to pointer types and use it in various pointer-related operations.

## Test Plan

Via existing and new tests.

## Test Result

No ROCgdb regression. There are a few failures related to a dbgapi issue, which I'll handle separately.

### DIFF
--- a/gdb/NEWS
+++ b/gdb/NEWS
@@ -116,6 +116,9 @@
   intent to remove it in a future release.
   The s390 64-bit target (s390x-*) remains supported.
 
+* GDB now recognizes DWARF DW_AT_address_space attribute for pointer,
+  reference, and rvalue reference types.
+
 * Configure changes
 
 ** --with-babeltrace has been removed.  The babeltrace library was
--- a/gdb/amdgpu-tdep.c
+++ b/gdb/amdgpu-tdep.c
@@ -1598,6 +1598,71 @@ amdgpu_segment_address_from_core_address (CORE_ADDR addr)
   return addr & significant_bits;
 }
 
+/* Convert a given address in the given address space to an address in
+   another address space.  This is essentially a wrapper for a dbgapi
+   function.  PTID and SIMD_LANE are the contexts for which the
+   conversion should be made.  PTID can be NULL_PTID when the address
+   does not necessarily depend on wave context.  */
+
+static std::optional<addr_range>
+amdgpu_convert_address (gdbarch *gdbarch, ptid_t ptid, int simd_lane,
+			CORE_ADDR from_address,
+			arch_addr_space_id from_aspace_id,
+			arch_addr_space_id to_aspace_id)
+{
+  /* Argument should be already stripped off its address space id.  */
+  gdb_assert (amdgpu_address_space_id_from_core_address (from_address)
+	      == ARCH_ADDR_SPACE_ID_DEFAULT);
+
+  /* We should never get a valid ptid here that is not a gpu
+     thread.  */
+  gdb_assert (ptid == null_ptid || ptid_is_gpu (ptid));
+
+  const amd_dbgapi_wave_id_t wave_id
+    = (ptid != null_ptid)
+    ? get_amd_dbgapi_wave_id (ptid)
+    : AMD_DBGAPI_WAVE_NONE;
+
+  amd_dbgapi_architecture_id_t architecture_id;
+  if (ptid == null_ptid
+      && (amd_dbgapi_get_architecture (gdbarch_bfd_arch_info (gdbarch)->mach,
+				       &architecture_id)
+	  != AMD_DBGAPI_STATUS_SUCCESS))
+      error (_("amd_dbgapi_get_architecture failed"));
+  else if (ptid != null_ptid
+	   && (amd_dbgapi_wave_get_info (wave_id,
+					 AMD_DBGAPI_WAVE_INFO_ARCHITECTURE,
+					 sizeof (architecture_id),
+					 &architecture_id)
+	       != AMD_DBGAPI_STATUS_SUCCESS))
+    error (_("amd_dbgapi_wave_get_info failed"));
+
+  /* Get a dbgapi address space id for the original address space.  */
+  amd_dbgapi_address_space_id_t dbgapi_from_addr_space_id;
+  if (amd_dbgapi_dwarf_address_space_to_address_space
+	(architecture_id, from_aspace_id, &dbgapi_from_addr_space_id)
+      != AMD_DBGAPI_STATUS_SUCCESS)
+    error (_("amd_dbgapi_dwarf_address_space_to_address_space failed"));
+
+  /* Get the dbgapi address space id for the default address space.  */
+  amd_dbgapi_address_space_id_t dbgapi_to_addr_space_id;
+  if (amd_dbgapi_dwarf_address_space_to_address_space
+	(architecture_id, to_aspace_id, &dbgapi_to_addr_space_id)
+      != AMD_DBGAPI_STATUS_SUCCESS)
+    error (_("amd_dbgapi_dwarf_address_space_to_address_space failed"));
+
+  CORE_ADDR converted_address;
+  size_t contiguous_size;
+
+  if (amd_dbgapi_convert_address_space
+	(wave_id, simd_lane, dbgapi_from_addr_space_id, from_address,
+	 dbgapi_to_addr_space_id, &converted_address, &contiguous_size)
+      != AMD_DBGAPI_STATUS_SUCCESS)
+    return std::nullopt;
+
+  return addr_range {converted_address, contiguous_size};
+}
+
 /* Address class to address space mapping.
 
    TODO: This is just a quick fix to make the address class hand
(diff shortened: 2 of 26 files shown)


palves:
Is it really DW_AT_address_space?  Or DW_LLVM_AT_address_space?

aktemur:
> Is it really DW_AT_address_space? Or DW_LLVM_AT_address_space?

llvm-dwarfdump displays it as `DW_AT_LLVM_address_space`. https://dwarfstd.org/issues/260211.1.html is not officially accepted, and there is no binary code for the attribute, yet. So, I defined it in GDB as `DW_AT_LLVM_address_space`, but in the code I usually refer to it as `DW_AT_address_space`.

lumachad:
It seems we need a refresh to address conflicts.

aktemur:
> It seems we need a refresh to address conflicts.

I updated the branch. Thanks.

lumachad:
Thanks. Let me go through this one.

lumachad:
> [!NOTE]
> The following review comments were generated with automated assistance. Please use your own judgement on each point.

## Review

### Mandatory

**`gdb/arch-utils.h:427`** — Missing two-space indent before `(`:
```c
-extern arch_addr_space_id default_address_space_dwarf_to_id
-(gdbarch *gdbarch, ULONGEST dwarf_addr_space);
+extern arch_addr_space_id default_address_space_dwarf_to_id
+  (gdbarch *gdbarch, ULONGEST dwarf_addr_space);
```
Matches the style of the declaration two lines above.

---

**`gdb/amdgpu-tdep.c:1647`** — Comment was copy-pasted from `amdgpu_get_watchable_aliases` where the destination was always the default address space. Here `to_aspace_id` is caller-supplied and can be any address space:
```c
-/* Get the dbgapi address space id for the default address space.  */
+/* Get the dbgapi address space id for the destination address space.  */
```

---

**`gdb/amdgpu-tdep.c:1749`** — Two issues in the TODO comment:
1. "i.e." should be "i.e.," (comma after the abbreviation).
2. "In that case, we use it" is vague. Suggested: "In that case, we use the address space from the type."

---

**`gdb/amdgpu-tdep.c:1888`** — Wrong error message. The call that just failed is `amd_dbgapi_wave_get_info`, not `amd_dbgapi_get_architecture`:
```c
-error (_("amd_dbgapi_get_architecture failed"));
+error (_("amd_dbgapi_wave_get_info failed"));
```

---

**`gdb/dwarf2/read.c:12004`** — The comment says `DW_AT_address_space` but the function reads `DW_AT_LLVM_address_space` (the LLVM vendor extension). Suggested fix:
```c
-/* Helper function to read DW_AT_address_space attribute.  */
+/* Helper function to read the DW_AT_LLVM_address_space attribute
+   (the LLVM vendor extension for DW_AT_address_space).  */
```

---

**`gdb/gdbtypes.h:170`** — The value 15 is the maximum representable in the 4-bit `address_space` bitfield (2^4 - 1 = 15), but that connection is not obvious. Suggested:
```c
+  /* Maximum value storable in the address_space bitfield above.  */
   static constexpr arch_addr_space_id ADDRESS_SPACE_MAX = 15;
```

---

**`gdb/testsuite/gdb.rocm/aspace-pointer.exp:53`** — The `sizeof` block compares C-level sizeof against GDB's sizeof, which is a good indirect check for `address_space_pointer_size`. Consider adding a `maint print type` check to make the coverage explicit and easier to diagnose on regression, e.g.:
```tcl
gdb_test "maint print type local_ptr0" "length.*4"
```
(Adjust expected length to whatever the ABI specifies for LDS pointers.)

---

**`gdb/testsuite/gdb.rocm/aspace-pointer.exp:83`** — The "legal assignment" block tests reading through an assigned pointer but not writing through one. Consider adding a write-through test, e.g.:
```tcl
gdb_test_no_output "set *local_ptr0 = 999"
gdb_test "print *local_ptr0" "= 999" "write through local_ptr0"
```
`generic-address.exp` covers write-through for generic pointers; it would be good to have the same for typed address-space pointers.

---

**`gdb/testsuite/gdb.rocm/aspace-user-input.exp:120`** — Pattern `TYPE_ADDRESS_SPACE.*` is too loose; it would pass even if the wrong ID was stored. `local` is address space 3, so:
```tcl
-    "TYPE_ADDRESS_SPACE.*"
+    "TYPE_ADDRESS_SPACE\\(3\\).*"
```

---

### Optional

**`gdb/gdbarch_components.py:864`** — `params` uses bare `type *` without `struct`. The generated `gdbarch-gen.h` will follow suit, but all neighbouring declarations in that header use `struct type *`. Consider `"struct type *"` for consistency.

**`gdb/valops.c:313`** — "when there is, say" is informal. Suggested: "when, for example, there is a pointer to generic address space".

**`gdb/amdgpu-tdep.c:1818`** — `(gdb_byte *) &from_address` reinterprets a host `CORE_ADDR` as a raw byte buffer; `amdgpu_address_to_pointer` then writes it in target byte order via `store_unsigned_integer`. Reading `from_address` back as a host integer would produce wrong bytes on a big-endian host / little-endian target. ROCm only runs on LE hosts today so this is latent, but an alternative is to strip the address space bits directly via `amdgpu_segment_address_from_core_address` instead of the store/load round-trip.

**`gdb/gdbtypes.c:565`** — `gdbarch_address_space_pointer_size` is called via `type->arch()`. In a cross-debug scenario where `type->arch()` differs from the actual target gdbarch, the reported size could be wrong. Pre-existing type/arch coupling issue, not introduced here, but worth keeping in mind.

**`gdb/arch-utils.c:1154`** — The new implementation silently changes the behaviour of the default: the old version was a silent cast; this one validates and errors on unrecognized IDs. A comment noting the change would help future maintainers understand why the default is no longer a simple cast.

**`include/dwarf2.h:629`** — `DW_ASPACE_LLVM_AMDGPU_region` is defined but not referenced anywhere in the patch. A short inline comment would clarify intent:
```c
  DW_ASPACE_LLVM_AMDGPU_region = 2,  /* Defined for completeness; currently unused.  */
```

lumachad:
It looks like we have a number of failures in CI:
- gdb.rocm/runtime-core.exp
- gdb.dwarf2/dw2-llvm-info-address.exp
- gdb.rocm/aspace-watchpoint.exp
- gdb.rocm/aspace-pointer.exp

aktemur:
> It looks like we have a number of failures in CI:
> 
> * gdb.rocm/runtime-core.exp
> * gdb.dwarf2/dw2-llvm-info-address.exp
> * gdb.rocm/aspace-watchpoint.exp
> * gdb.rocm/aspace-pointer.exp

Please ignore these for the moment. They will go away when https://github.com/ROCm/rocm-systems/pull/10163 is merged.

lumachad:
> > It looks like we have a number of failures in CI:
> > 
> > * gdb.rocm/runtime-core.exp
> > * gdb.dwarf2/dw2-llvm-info-address.exp
> > * gdb.rocm/aspace-watchpoint.exp
> > * gdb.rocm/aspace-pointer.exp
> 
> Please ignore these for the moment. They will go away when [ROCm/rocm-systems#10163](https://github.com/ROCm/rocm-systems/pull/10163) is merged.

But gdb.dwarf2/dw2-llvm-info-address.exp is a cpu-based test that doesn't use rocdbgapi, right?

On ignoring the gdb.rocm ones, I'm not sure we can. If we have a situation where we need an atomic change between rocgdb and rocdbgapi, then we need to put guards on either or both so we can freely merge their changes into TheRock main without breaking builds/tests.

Is it the case here?

aktemur:
> But gdb.dwarf2/dw2-llvm-info-address.exp is a cpu-based test that doesn't use rocdbgapi, right?

That's right. I'll work on that.

> On ignoring the gdb.rocm ones, I'm not sure we can. If we have a situation where we need an atomic change between rocgdb and rocdbgapi, then we need to put guards on either or both so we can freely merge their changes into TheRock main without breaking builds/tests.
> 
> Is it the case here?

There is the atomic case, but it is being handled in another (and smaller) PR: https://github.com/ROCm/rocm-systems/pull/10163 Once the dbgapi PR is in, I'll adjust and rebase this PR, after which the gdb.rocm failures should go away.

lumachad:
> > But gdb.dwarf2/dw2-llvm-info-address.exp is a cpu-based test that doesn't use rocdbgapi, right?
> 
> That's right. I'll work on that.
> 
> > On ignoring the gdb.rocm ones, I'm not sure we can. If we have a situation where we need an atomic change between rocgdb and rocdbgapi, then we need to put guards on either or both so we can freely merge their changes into TheRock main without breaking builds/tests.
> > Is it the case here?
> 
> There is the atomic case, but it is being handled in another (and smaller) PR: [ROCm/rocm-systems#10163](https://github.com/ROCm/rocm-systems/pull/10163) Once the dbgapi PR is in, I'll adjust and rebase this PR, after which the gdb.rocm failures should go away.

Got it. Thanks for clarifying it.

aktemur:
> But gdb.dwarf2/dw2-llvm-info-address.exp is a cpu-based test that doesn't use rocdbgapi, right?

I fixed this in my local branch and will include it in the next revision.

aktemur:
> The following review comments were generated with automated assistance. Please use your own judgement on each point.

I went over them and addressed those that I thought are valid. They were actually very useful. Thank you.

lumachad:
Thanks. Let's wait for the right pieces to land so we can test this.

lumachad:
Conflicts need addressing. But I see a pattern of failing tests as well.
- gdb.rocm/runtime-core.exp
- gdb.rocm/aspace-watchpoint.exp
- gdb.rocm/aspace-pointer.exp

PR #256, opened 2026-08-06
[ci] Fetch build and CPU test runner dynamically from therock-ci-config

lumachad:
## Summary

- Replace hardcoded \`azure-linux-scale-rocm\` (Azure) in the build job and \`aws-linux-scale-rocm-prod\` (AWS) in the CPU test job with dynamic runner selection driven by \`ROCm/therock-ci-config\`'s \`runner-config.json\` (\`build_runners.linux.default\`, highest-weight entry).
- The \`setup\` job in \`therock-ci.yml\` checks out \`therock-ci-config\` and selects the runner via \`ci_config_api.py\`, passing it as a new \`build_runs_on\` input to \`therock-ci-linux.yml\`.
- \`therock-test-packages.yml\`'s \`configure_test_matrix\` job applies the same logic and exposes \`cpu_runs_on\` for the CPU component's \`test_runs_on\` expression.
- Both runner-selection steps fall back to \`aws-linux-scale-rocm-prod\` if \`therock-ci-config\` is unavailable (e.g., in forks).
- Runner-selection logic extracted into \`.github/scripts/select_runner.py\` to avoid duplication between the two workflows.

## Test plan

- [x] Verify CI triggers on this PR and the build job picks up the correct runner label from \`therock-ci-config\`
- [x] Verify CPU test component routes to the dynamically selected runner
- [x] Verify GPU test component still uses the per-component runner from \`fetch_test_configurations.py\`
- [x] Verify fallback behaviour by temporarily breaking the ci-config checkout (\`continue-on-error: true\` already set)

### DIFF
--- a/.github/workflows/therock-ci-linux.yml
+++ b/.github/workflows/therock-ci-linux.yml
@@ -9,6 +9,9 @@ on:
         type: string
       extra_cmake_options:
         type: string
+      build_runs_on:
+        type: string
+        default: aws-linux-scale-rocm-prod
 
 permissions:
   contents: read
@@ -19,7 +22,7 @@ env:
 jobs:
   therock-build-linux:
     name: Build Linux Packages
-    runs-on: azure-linux-scale-rocm
+    runs-on: ${{ inputs.build_runs_on }}
     permissions:
       id-token: write
     outputs:
--- a/.github/workflows/therock-test-packages.yml
+++ b/.github/workflows/therock-test-packages.yml
@@ -37,6 +37,7 @@ jobs:
     runs-on: ubuntu-24.04
     outputs:
       components: ${{ steps.patch-scripts.outputs.components }}
+      cpu_runs_on: ${{ steps.runner.outputs.cpu_runs_on }}
     steps:
       - name: Checkout TheRock repository
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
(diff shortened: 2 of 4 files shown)


lumachad:
@czidev-amd ping?

.github/workflows/therock-ci-linux.yml
```diff
@@ -19,7 +22,7 @@ env:
 jobs:
   therock-build-linux:
     name: Build Linux Packages
-    runs-on: azure-linux-scale-rocm
+    runs-on: ${{ inputs.build_runs_on }}
```
czidev-amd, line 25: This is a change from azure to aws. It seems to me the commit message doesn't clearly catch this. Maybe rephrase commit message, not a blocker.

.github/workflows/therock-test-packages.yml
```diff
@@ -57,6 +58,36 @@ jobs:
         with:
           python-version: "3.12"
 
+      - name: Select CPU runner
+        id: runner
+        run: |
+          python3 - <<'EOF'
```
czidev-amd: This python script seems very similar with the above one. Is it possible to have it only in one location?

czidev-amd (approved):
Maybe a new sentence in the commit message explaining what was the issue with the old runner. Also, if  it is possible, to dedup the python snip.
Otherwise it looks good to me.

.github/workflows/therock-ci-linux.yml
lumachad: Thanks. Done now.

.github/workflows/therock-test-packages.yml
lumachad: Yeah, because it lives in ROCgdb and not in TheRock. The code has been extracted to a separate python script now, and we do a sparse ROCgdb checkout to get it in the second job. Thanks.

PR #259, opened 2026-08-10
gdb.rocm/omp-target-basic: xfail firstprivate value test

lumachad:
The compiler emits `unsigned long` instead of `int` for the firstprivate
argument variable (AIROCGDB-649), causing GDB to print the wrong type.
The test may pass or fail depending on the surrounding values/memory.


lumachad:
Or maybe a kfail would be best.

lancesix:
> Or maybe a kfail would be best.

xfail is the right one.

KFAIL is for known GDB issues, XFAIL is for expected failures given the environment.  See https://github.com/ROCm/ROCgdb/blob/amd-staging/gdb/testsuite/README#L757-L776 for a discussion of the difference between the two.

lancesix (approved):
LGTM, thanks.

PR #261, opened 2026-08-10
gdb/testsuite/hip: pass --offload-arch flags in hip.exp

lumachad:
The hip board was not passing any --offload-arch flags to the compiler,
so HIP defaulted to the compiler's built-in target, which may not match
the actual GPU on the machine.

Fix this by calling hcc_amdgpu_targets in hip.exp's gdb_compile override
to add the correct --offload-arch flags, matching what gdb.exp already
does for regular HIP tests.

find_amdgpu_devices compiles a host-only probe program to obtain the
target list.  It passes hip_no_offload_arch to signal that no
--offload-arch is needed for that compile.  Without a corresponding
guard in our override, calling hcc_amdgpu_targets there would recurse
back into find_amdgpu_devices before its result is cached, causing an
infinite loop.  Fix this by skipping our override entirely when
hip_no_offload_arch is present.

Also fix an existing bug where lreplace was called with index -1 when
"c++" was not in the options list, silently removing the last element.

### DIFF
--- a/gdb/testsuite/boards/hip.exp
+++ b/gdb/testsuite/boards/hip.exp
@@ -265,7 +265,11 @@ proc gdb_compile {source dest type options} {
 
     set lang [hip_get_compile_language {$options}]
 
+    # Skip our overrides for host-only probe compilations done by
+    # find_amdgpu_devices.  Those pass hip_no_offload_arch to avoid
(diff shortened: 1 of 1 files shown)


lancesix (approved):
Looks reasonable to me.

PR #262, opened 2026-08-10
github: improve TheRock dependency update workflow

lumachad:
Move THEROCK_COMMIT_REF and the build container digest out of
therock-ci-linux.yml and into .github/configs.json, a plain data file
documented in .github/configs.md. A \`resolve\` job reads and validates
the pins and exposes them as outputs; the build job references those
outputs for \`container.image\` and \`THEROCK_COMMIT_REF\`. This lets the
automated update workflow push changes with only \`contents: write\` on
the GITHUB_TOKEN.

The update script is reworked accordingly:

- Fetch the TheRock branch HEAD and committer timestamp from the GitHub
  REST API (retried on transient failures) instead of \`git ls-remote\`
  and \`date.today()\`. Fetch the build image creation timestamp from the
  OCI config blob instead of \`date.today()\`. Both timestamps are stored
  in full ISO 8601 format and gated per pin so an unchanged value keeps
  its original timestamp. Timestamp fetches are skipped in dry-run mode.
- Validate all API response fields with descriptive errors; guard
  required config keys before any git side-effects.
- PR title is "Automated TheRock dependency update (\<date\>)"; body
  lists each bumped pin with its short ref and date.
- Print both tracked values on every run with clear update / unchanged
  labels; name what would change in dry-run mode.

Includes a comprehensive unit test suite covering all public functions,
including retry logic, error paths, timestamp gating, and the
run_update state machine.

### DIFF
--- a/.github/configs.json
+++ b/.github/configs.json
@@ -1,0 +1,6 @@
+{
+  "therock_commit_ref": "2ee541083481e6c93f2ea811f212b61047980b1a",
+  "therock_commit_created": "2026-07-30T10:07:58Z",
+  "build_image": "ghcr.io/rocm/therock_build_manylinux_x86_64@sha256:8616d086df21bcc835aed5112ff0d878530a0cf5433f8f42c7b28b973a75bb19",
+  "build_image_created": "2026-07-27T09:37:30.626203739Z"
+}
--- a/.github/scripts/update_therock_deps.py
+++ b/.github/scripts/update_therock_deps.py
@@ -1,20 +1,21 @@
 #!/usr/bin/env python3
 # Copyright Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
-"""Update TheRock references in ROCgdb's CI workflow.
+"""Update TheRock references in ROCgdb's CI dependency pin file.
 
 Fetches the latest TheRock commit and the latest therock_build_manylinux_x86_64
-container digest, updates .github/workflows/therock-ci-linux.yml, then pushes a
-branch and opens a ROCgdb PR labelled 'therock-deps'. If an update PR already
-carries that label, the run skips without touching git.
+container digest, updates .github/configs.json, then pushes a branch and
+opens a ROCgdb PR labelled 'therock-deps'. If an update PR already carries that
+label, the run skips without touching git.
 
 The test container is not updated here: it is pinned transitively through
-THEROCK_COMMIT_REF (TheRock's fetch_test_configurations.py supplies the test
+therock_commit_ref (TheRock's fetch_test_configurations.py supplies the test
 image), so bumping the commit ref updates it automatically.
 """
 
 import argparse
 import json
+import os
 import re
 import shlex
 import subprocess
@@ -31,9 +32,8 @@ from pathlib import Path
 
 ROCGDB_REPO = "ROCm/ROCgdb"
 BASE_BRANCH = "amd-staging"
-CI_LINUX = Path(".github/workflows/therock-ci-linux.yml")
+CONFIG_FILE = Path(".github/configs.json")
 
-THEROCK_REPO_URL = "https://github.com/ROCm/TheRock.git"
 BUILD_IMAGE = "ghcr.io/rocm/therock_build_manylinux_x86_64"
 
 UPDATE_LABEL = "therock-deps"
@@ -52,16 +52,24 @@ MAX_RETRIES = 3  # number of retries for network requests
 # ---------------------------------------------------------------------------
 
 
-def urlopen_with_retry(req, timeout: int = 30):
-    """Open a URL with retries on transient errors."""
+def urlopen_with_retry(req: str | urllib.request.Request, timeout: int = 30):
+    """Open a URL with retries on transient errors (5xx, timeout); raises immediately on 4xx."""
     for attempt in range(1, MAX_RETRIES + 1):
         try:
             return urllib.request.urlopen(req, timeout=timeout)
+        except urllib.error.HTTPError as exc:
+            if exc.code < 500:
+                raise
+            if attempt == MAX_RETRIES:
+                raise
+            print(f"[RETRY] attempt {attempt}/{MAX_RETRIES} failed: {exc}")
+            time.sleep(2**attempt)
         except (urllib.error.URLError, TimeoutError) as exc:
             if attempt == MAX_RETRIES:
                 raise
             print(f"[RETRY] attempt {attempt}/{MAX_RETRIES} failed: {exc}")
-            time.sleep(2 ** attempt)
+            time.sleep(2**attempt)
+    raise RuntimeError("urlopen_with_retry: unreachable")
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


lumachad:
@spatrang Could you please eyeball this one?

.github/scripts/update_therock_deps.py
```diff
@@ -186,49 +186,28 @@ def get_build_digest() -> tuple[str, str]:
...
-        raise ValueError(
-            f"Expected exactly 1 match for build image pattern, got {n}. "
-            "Check that the line has a trailing '# date' comment."
-        )
+def read_deps() -> dict[str, str]:
+    """Read and return the current dependency pins from CONFIG_FILE."""
+    return json.loads(CONFIG_FILE.read_text())
+
+
+def write_deps(
```
lancesix, line 198: I find this name missleading.  With `read_depts` reading a file, one would expect `write_deps` to write one, but this is not what this does.

lancesix (commented):
The 2nd commit msg says talks about a file is named .github/therock-deps.json, but code uses .config/configs.json.

.github/scripts/update_therock_deps.py
lumachad: Yeah. Changed it so it actually writes things to the file.

lumachad:
> The 2nd commit msg says talks about a file is named .github/therock-deps.json, but code uses .config/configs.json.

Missed this one. Had a change of heart and renamed it. Missed updating it.

.github/scripts/update_therock_deps.py
```diff
@@ -186,63 +186,43 @@ def get_build_digest() -> tuple[str, str]:
...
+    commit_date: str,
+    build_digest: str,
+    build_date: str,
+) -> None:
+    """Write updated dependency pins to CONFIG_FILE."""
+    updated = dict(pins)
+    updated["therock_commit_ref"] = commit
+    updated["therock_commit_date"] = commit_date
+    updated["build_image"] = f"{BUILD_IMAGE}@{build_digest}"
+    updated["build_image_date"] = build_date
```
spatrang: **Correctness:** these date fields are refreshed unconditionally. `commit_date` and `build_date` are both just `date.today()`, and the call site (line 330) passes them regardless of what actually changed. So if only the TheRock commit changed, `build_image_date` still gets bumped to today even though `build_image` is unchanged (and vice-versa). Since these `*_date` fields are meant to record when each pin was last actually updated, this makes them misleading. Suggest gating each date on its own change flag, e.g. only set `build_image_date` when `digest_changed` and `therock_commit_date` when `commit_changed`.

.github/scripts/update_therock_deps.py
```diff
@@ -186,63 +186,43 @@ def get_build_digest() -> tuple[str, str]:
...
-        f"{BUILD_IMAGE}@{build_digest} # {build_date}",
-        updated,
-    )
-    if n != 1:
-        raise ValueError(
-            f"Expected exactly 1 match for build image pattern, got {n}. "
-            "Check that the line has a trailing '# date' comment."
-        )
-    return updated
+def write_deps(
```
spatrang: Nit (follow-up to @lancesix): it now genuinely writes, good. But the signature `write_deps(pins, commit, commit_date, build_digest, build_date)` mixes the old dict with new scalars. Since you'll be touching this for the date fix above, a cleaner shape is to build the updated dict in the caller and let `write_deps(pins: dict)` just serialize it.

.github/configs.json
```diff
@@ -0,0 +1,6 @@
+{
```
spatrang, line 1: `configs.json` reads as a general-purpose config file, but it only holds TheRock pins - a future contributor will be tempted to dump unrelated settings here. Consider `therock-deps.json` (which is also what the PR description implies) for a more self-documenting name. Optional, since you resolved the earlier mismatch by keeping this name.

.github/workflows/therock-ci-linux.yml
```diff
@@ -13,20 +13,43 @@ on:
...
+      image: ${{ steps.read.outputs.image }}
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Read TheRock dependency pins
+        id: read
+        run: |
+          commit=$(python3 -c "import json,sys; d=json.load(open('.github/configs.json')); print(d['therock_commit_ref'])")
+          image=$(python3 -c "import json,sys; d=json.load(open('.github/configs.json')); print(d['build_image'])")
```
spatrang: A few small things on these two one-liners:

- **Path duplication (should fix):** `.github/configs.json` is hard-coded here twice and again as `CONFIG_FILE` in `update_therock_deps.py`. No single source of truth - a rename must touch all three. Consider one combined `python3 -c` that reads both keys in a single process.
- **Dead import (nit):** `import json,sys` - `sys` is unused in both one-liners.
- **Error clarity (nit):** a missing key dies with a raw `KeyError` traceback, unlike the friendly `echo ... >&2` messages you added for the regex failures below.

spatrang (commented):
Overall the design is sound - the `resolve` job is genuinely necessary since `container.image` must be known before the container starts, so the pins can't be read from inside the build container, and dropping the `workflows:` token scope is well motivated. A few items below. Note the commit-message/filename mismatch @lancesix raised is now resolved by the reworded message. One correctness bug (the date fields) is worth blocking on; the rest are should-fix / nits.

.github/scripts/update_therock_deps.py
lumachad: Fixed. Each `*_created` field is now gated on its own change flag. Also replaced `date.today()` with the actual event timestamps — committer date from the GitHub API for the commit, `created` from the OCI config blob for the image.

.github/scripts/update_therock_deps.py
lumachad: Fixed. `write_deps` now takes a single dict and just serializes it.

.github/configs.json
lumachad: Keeping `configs.json` — we plan to reuse it for multi-arch CI configuration. Added `.github/configs.md` documenting all fields.

.github/workflows/therock-ci-linux.yml
lumachad: All three fixed. Merged into one `python3 -c` process reading both keys at once. `sys` is now used for `sys.stderr`/`sys.exit`. Added `FileNotFoundError`, `json.JSONDecodeError`, and `KeyError` handlers with friendly messages, plus `|| { echo ...; exit 1; }` on the process substitution.

lumachad:
Improved a few things overall. We're now fetching the proper timestamps of commits and the build image. A bit more hardening across the board, more error checks, messages and graceful exits.

lumachad:
I went ahead and also added some unit tests to make sure things are stable and sane.

spatrang (approved):
LGTM.

Second-round review against `81ef9ca` — all four points from the first round are addressed:

- **Date fields** now gated per-change flag (`therock_commit_created`/`build_image_created` only refresh when their own pin changes), and they carry real event timestamps (GitHub committer date + OCI config `created`) instead of `date.today()`.
- **`write_deps(pins)`** takes a single dict and just serializes.
- **`configs.json`** kept intentionally (multi-arch reuse) and now documented in `configs.md`.
- **`resolve` one-liner** consolidated into a single `python3 -c` reading both keys, with `FileNotFoundError`/`JSONDecodeError`/`KeyError` handling and friendly messages.

Also verified: the new `test_update_therock_deps.py` suite passes (81/81) with good coverage (retry/backoff, 4xx-vs-5xx, dry-run timestamp skipping, per-flag date gating), the workflow diff is clean (only the `resolve` job + wiring), and the `configs.md` reference to `therock-deps-update.yml` resolves. Nice hardening overall.

PR #263, opened 2026-08-11
github: remove push triggers, run CI only on PRs

lumachad:
## Summary
- Remove `push` trigger from `therock-ci.yml` — TheRock CI was running
  redundantly on every push to `amd-staging` in addition to per-PR runs
- Remove `push` trigger from `pre-commit.yml` for the same reason, and
  simplify the base ref logic now that push events are no longer handled
- Removing the `push` trigger from `pre-commit.yml` reparents the existing
  `branches` filter onto `pull_request`, restricting pre-commit to PRs
  targeting `amd-staging` and `amd-staging-rocgdb-*`, consistent with
  `therock-ci.yml`

## Test plan
- [ ] Verify TheRock CI and pre-commit run on new PRs targeting `amd-staging`
- [ ] Verify no CI runs are triggered on direct pushes to `amd-staging`

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -5,7 +5,6 @@ name: pre-commit
 
 on:
   pull_request:
-  push:
     branches:
       - amd-staging
       - 'amd-staging-rocgdb-*'
@@ -24,12 +23,7 @@ jobs:
       - name: Determine base ref
         id: base-ref
         run: |
-          if [ "${{ github.event_name }}" = "pull_request" ]; then
-            BASE_REF="${{ github.event.pull_request.base.ref }}"
-          else
-            # For push events, use the current branch name
-            BASE_REF="${{ github.ref_name }}"
-          fi
+          BASE_REF="${{ github.event.pull_request.base.ref }}"
           echo "base_ref=${BASE_REF}" >> $GITHUB_OUTPUT
           echo "Base ref: ${BASE_REF}"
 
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-commit.yml
```diff
@@ -5,7 +5,6 @@ name: pre-commit
 
 on:
   pull_request:
```
spatrang, line 7: Removing the `push:` line reparents the `branches:` filter onto `pull_request:`. Previously `pull_request:` had no branch filter (it ran on PRs targeting *any* base branch), while `branches:` belonged to `push:`. After this change, pre-commit only runs on PRs targeting `amd-staging` / `amd-staging-rocgdb-*`.

That's a behavior change beyond just removing the push trigger. If it's intended (it does make this consistent with `therock-ci.yml`), please mention it in the PR description. If not, drop the `branches:` block entirely so it stays:

```yaml
on:
  pull_request:
```

lumachad: Intentional. We want pre-commit to run only on PRs targeting `amd-staging` and `amd-staging-rocgdb-*`, consistent with `therock-ci.yml`. Will update the PR description.

spatrang (approved):
LGTM. The branch-filter reparenting in `pre-commit.yml` is confirmed intentional and now documented in the PR description, making it consistent with `therock-ci.yml`. Both push-trigger removals and the base-ref simplification look correct.

PR #265, opened 2026-08-11
gdb.rocm/omp-target-basic: handle firstprivate compiler type bug gracefully

lumachad:
The firstprivate test previously used setup_xfail "*-*-*" + gdb_test,
which unconditionally marks the result as an expected failure.  This
means:

- a PASS becomes an XPASS (noisy when the compiler is fixed)
- a FAIL becomes an XFAIL (correct, but the mechanism hides any future
  regression)

Replace it with gdb_test_multiple that accepts "= 42" as a clean PASS
and any other integer as an XFAIL.  When the compiler bug is fixed the
test starts passing without requiring any testsuite change.

### DIFF
--- a/gdb/testsuite/gdb.rocm/omp-target-basic.exp
+++ b/gdb/testsuite/gdb.rocm/omp-target-basic.exp
@@ -60,11 +60,18 @@ with_rocm_gpu_lock {
     # Still stopped at the first statement in the region (out_arr is
     # not written yet), so the data-sharing values are deterministic.
     with_test_prefix "locals" {
-	# The compiler emits unsigned long instead of int for the
-	# firstprivate argument variable.  The test may pass or fail
-	# depending on the surrounding values/memory.
-	setup_xfail "*-*-*"
-	gdb_test "print firstpriv_val" "= 42" "firstprivate value"
+	# The compiler may emit unsigned long instead of int for the
+	# firstprivate argument variable, producing an unpredictable
+	# value.  Accept either the correct value or any integer.
+	gdb_test_multiple "print firstpriv_val" "firstprivate value" {
+	    -re -wrap "= 42" {
+		pass $gdb_test_name
+	    }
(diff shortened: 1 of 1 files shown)


gdb/testsuite/gdb.rocm/omp-target-basic.exp
```diff
@@ -60,11 +60,18 @@ with_rocm_gpu_lock {
...
+	# The compiler may emit unsigned long instead of int for the
+	# firstprivate argument variable, producing an unpredictable
+	# value.  Accept either the correct value or any integer.
+	gdb_test_multiple "print firstpriv_val" "firstprivate value" {
+	    -re -wrap "= 42" {
+		pass $gdb_test_name
+	    }
+	    -re -wrap "= $::decimal" {
+		# Compiler type bug: value is unreliable.
+		xfail $gdb_test_name
```
aktemur: If there is a compiler ticket we can refer to, doing this as
```
setup_xfail "*-*-*" "JIRASPACE=1234"
fail $gdb_test_name
```
would be better for tracking. When the fix is there, we would also notice a change in test results from XFAIL to XPASS. If the ticket information is not available, this looks good to me, too.

aktemur (approved):

gdb/testsuite/gdb.rocm/omp-target-basic.exp
lumachad: Thanks Let me add the ticket number.

The reason behind this ticket is we don't want to produce an XPASS. @amd-shahab reported having his testsuite runs returning non-zero due to it (a quirk in gdb's testsuite?). So this sidesteps the XPASS and only reports the XFAIL to dodge that issue.

Alternatively we'd have another test that verifies the type the compiler is generating is correct, but that might not be desirable since it's not really testing rocgdb but the compiler.

amd-shahab:
It works fine on my end. thanks!

PR #266, opened 2026-08-11
gdb.rocm/alu-exceptions.exp: prevent ROCr core dumps on exceptions

lumachad:
Use GDB's "set environment" to pass HSA_DISABLE_COREDUMP_ON_EXCEPTION=1
to the inferior, so the ROCr runtime does not generate a core dump when
the ALU exception causes the process to abort.  Using "set environment"
keeps the setting scoped to the inferior and avoids contaminating other
parallel test processes.

### DIFF
--- a/gdb/testsuite/gdb.rocm/alu-exceptions.exp
+++ b/gdb/testsuite/gdb.rocm/alu-exceptions.exp
@@ -67,6 +67,8 @@ proc do_test { exception_name precise_alu_exceptions_supported } {
 
 	with_rocm_gpu_lock {
 	    gdb_test_no_output "set args $exception_name"
+	    # Prevent ROCr from generating a GPU core dump before calling abort.
+	    gdb_test_no_output "set environment HSA_DISABLE_COREDUMP_ON_EXCEPTION=1"
 	    if {![runto_main]} {
 		return
 	    }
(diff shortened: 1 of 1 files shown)


lumachad:
Hmmm, thinking further. We may disrupt parallel tests that may be trying to generate core files. Back to the drawing board...

lancesix:
> Hmmm, thinking further. We may disrupt parallel tests that may be trying to generate core files. Back to the drawing board...

I do not expect this should be an issue.  In parallel, all gdb run in their own shell.

But we can always use `gdb_test_no_output "set environment HSA_DISABLE_COREDUMP_ON_EXCEPTION=1 ` which would remove any doubt.

lumachad:
> > Hmmm, thinking further. We may disrupt parallel tests that may be trying to generate core files. Back to the drawing board...
> 
> I do not expect this should be an issue. In parallel, all gdb run in their own shell.
> 
> But we can always use `gdb_test_no_output "set environment HSA_DISABLE_COREDUMP_ON_EXCEPTION=1 ` which would remove any doubt.

That's exactly what I'm testing.

lumachad:
Ok. With the "set environment" approach I still see reliable and quicker runs.

gdb/testsuite/gdb.rocm/alu-exceptions.exp
```diff
@@ -67,6 +67,8 @@ proc do_test { exception_name precise_alu_exceptions_supported } {
 
 	with_rocm_gpu_lock {
 	    gdb_test_no_output "set args $exception_name"
+	    # Prevent ROCr from generating a core dump when the ALU exception causes abort.
```
lancesix: nit picking, I would rephrase this a tiny bit.  ROCr creates the GPU side of the core dump (not a full dump), and this is after GDB forwards the ALU exception and before calling abort.

"Prevent ROCr from generating a GPU core dump before calling abort"

lancesix (approved):
Just a minor comment, but overall LGTM.  Thanks.

gdb/testsuite/gdb.rocm/alu-exceptions.exp
lumachad: Done. Thanks.

PR #268, opened 2026-08-12
[amd-staging-rocgdb-18] gdb, amdgpu: add a CHANGELOG note about '#' operator in Fortran

aktemur:
Cherry-pick a patch from amd-staging to amd-staging-rocgdb-18.


lumachad:
We have other things we need to cherry-pick to both 16 and/or 18 branches first, so I think we should do it in bunk later on.

lumachad:
> We have other things we need to cherry-pick to both 16 and/or 18 branches first, so I think we should do it in bunk later on.

Actually, I can work around that. So this should be fine.

lumachad (approved):
OK. Thanks.

PR #271, opened 2026-08-12
gdb, testsuite: fix 'no such variable' error in gdb.rocm/hip-catch-er…

aktemur:
Running gdb.rocm/hip-catch-errors.exp in an environment where the `__hipOnError` symbol is not defined gives

```
  ERROR: tcl error sourcing <path-to>/gdb/testsuite/gdb.rocm/hip-catch-errors.exp.
  ERROR: can't read "res": no such variable
      while executing
  "return $res"
      ("uplevel" body line 27)
        invoked from within
  "uplevel 1 $body"
      invoked from within
  "with_test_prefix preparation_test {
      with_rocm_gpu_lock {
          clean_restart ${::testfile}.dbg
          if {![runto_main]} {
            return false
          }
          gdb_test_mul..."
```

Fix this by defining "res" at an earlier point.

The problem here is in fact deeper.  The code body that is used in `preparation_test` has `return` statements.  One would expect the `return` statements to actually return from `preparation_test` early and not even reach `return $res`, but that's not happening because of how `with_rocm_gpu_lock` and `with_lock` are defined and use Tcl's uplevel. Fixing them will be more involved and should be addressed separately.

### DIFF
--- a/gdb/testsuite/gdb.rocm/hip-catch-errors.exp
+++ b/gdb/testsuite/gdb.rocm/hip-catch-errors.exp
@@ -197,10 +197,14 @@ proc check_hip_launch_kernel_err {cmd} {
 # using the test program.  Return false if anything goes wrong.
 
 proc_with_prefix preparation_test {} {
+    # The return statements below return to this level.  Hence keep
+    # the state in a variable.
+    set res false
+
     with_rocm_gpu_lock {
 	clean_restart ${::testfile}.dbg
 	if {![runto_main]} {
-	    return false
+	    return
 	}
 	gdb_test_multiple "info symbol __hipOnError" \
 	    "check if __hipOnError symbol exists" {
@@ -209,7 +213,7 @@ proc_with_prefix preparation_test {} {
 	    }
 	    -re -wrap ".*" {
 		unsupported "no __hipOnError symbol was found"
-		return false
+		return
 	    }
 	}
 
@@ -219,7 +223,7 @@ proc_with_prefix preparation_test {} {
 	    gdb_test_no_output "set args ref"
 	    if {![runto [gdb_get_line_number \
 			    "Break after reference initialization"]]} {
-		return false
+		return
 	    }
 	    set res [init_reference_hip_errors]
 	}


lumachad:
Out of curiosity, what is the difference in the environment where "__hipOnError" doesn't exist?

aktemur:
> Out of curiosity, what is the difference in the environment where "__hipOnError" doesn't exist?

My installed version of the runtime apparently does not have the symbol, yet.

lumachad:
> > Out of curiosity, what is the difference in the environment where "__hipOnError" doesn't exist?
> 
> My installed version of the runtime apparently does not have the symbol, yet.

If that's from the official release, we might want to correct the assumption in the testcase.

aktemur:
> > > Out of curiosity, what is the difference in the environment where "__hipOnError" doesn't exist?
> > 
> > 
> > My installed version of the runtime apparently does not have the symbol, yet.
> 
> If that's from the official release, we might want to correct the assumption in the testcase.

Which assumption are you referring to? The testcase checks if the symbol is available and if not, aims to exit early with UNSUPPORTED. The problem is, a `return` statement inside the with-gpu-lock code block does not make it return from the enclosing procedure, but rather it returns from the with-gpu-lock procedure. This is rather a Tcl issue and not addressed by this particular PR.

lumachad:
> > > > Out of curiosity, what is the difference in the environment where "__hipOnError" doesn't exist?
> > > 
> > > 
> > > My installed version of the runtime apparently does not have the symbol, yet.
> > 
> > 
> > If that's from the official release, we might want to correct the assumption in the testcase.
> 
> Which assumption are you referring to? The testcase checks if the symbol is available and if not, aims to exit early with UNSUPPORTED. The problem is, a `return` statement inside the with-gpu-lock code block does not make it return from the enclosing procedure, but rather it returns from the with-gpu-lock procedure. This is rather a Tcl issue and not addressed by this particular PR.

Right, my reasoning is if this is a newer release we should always have such symbol. If it is not there, we might have a problem.

gdb/testsuite/gdb.rocm/hip-catch-errors.exp
```diff
@@ -197,6 +197,8 @@ proc check_hip_launch_kernel_err {cmd} {
 # using the test program.  Return false if anything goes wrong.
 
 proc_with_prefix preparation_test {} {
+    set res false
```
amd-shahab, line 202: Please add a comment above this `set` that why this initialization is needed.

amd-shahab (approved):
.

aktemur:
If we are going to use a result variable, it makes sense to rely on it consistently. Therefore I replaced all `return false` statements in the code block with a simple `return`. @amd-shahab, please re-review.

lancesix (approved):
LGTM, thanks.

amd-shahab:
> @amd-shahab, please re-review.

Looks good. Thanks!

PR #272, opened 2026-08-12
github: test_rocgdb.py: add --fraction/-f parallel job scaling

lumachad:
Adds a `--fraction`/`-f` option to `test_rocgdb.py` that lets callers specify a fraction of `os.cpu_count()` to use as the parallel job count, rather than a fixed `-j N`. The two options are mutually exclusive.

The resulting job count is computed as `max(1, int(cpu_count * F))` where `F` is in `(0.0, 1.0]`. This is useful in CI environments where you want to throttle parallelism relative to available cores without hard-coding a number.

### DIFF
--- a/.github/scripts/test_rocgdb.py
+++ b/.github/scripts/test_rocgdb.py
@@ -329,6 +329,30 @@ def _timeout_value(value: str) -> int:
     return ivalue
 
 
+def _parallel_fraction(value: str) -> float:
+    """
+    Validate argument is a float in the range (0.0, 1.0].
+
+    Args:
+        value: String value from command line.
+
+    Returns:
+        Float fraction value.
+
+    Raises:
+        argparse.ArgumentTypeError: If value is not in (0.0, 1.0].
+    """
+    try:
+        fvalue = float(value)
+    except ValueError:
+        raise argparse.ArgumentTypeError(f"{value!r} is not a valid number")
+    if not (0.0 < fvalue <= 1.0):
+        raise argparse.ArgumentTypeError(
+            f"--fraction must be in the range (0.0, 1.0], got {fvalue}"
+        )
+    return fvalue
+
+
 def _positive_nonzero_int(value: str) -> int:
     """
     Validate argument is a positive non-zero integer.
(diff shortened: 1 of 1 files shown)


lumachad:
This will get used in CI to parallelize our runs across different types of machines.

.github/scripts/test_rocgdb.py
```diff
@@ -618,11 +657,15 @@ def parse_arguments() -> argparse.Namespace:
...
-    # Validate that -j/--jobs is only used with --parallel.
+    # Validate that -j/--jobs and --fraction are only used with --parallel.
     if args.jobs is not None and not args.parallel:
         _log_error_and_exit(
             "-j/--jobs can only be specified when --parallel is provided."
         )
+    if args.fraction is not None and not args.parallel:
+        _log_error_and_exit(
+            "-f/--fraction can only be specified when --parallel is provided."
+        )
```
aktemur, line 668: I don't see where we check mutual exclusion of `-j` and `-f`. This place seems like good location to do that check and error out if both are given.

lumachad: We create a mutually exclusive group...

```
    parallel_jobs_group = parser.add_mutually_exclusive_group()
    parallel_jobs_group.add_argument(
        "-j",
        "--jobs",
        type=_positive_nonzero_int,
        default=None,
        metavar="N",
        help="Number of parallel jobs for make (e.g., -j4). Only valid with --parallel. "
        "Mutually exclusive with --fraction. "
        "If --parallel is used without -j or --fraction, defaults to os.cpu_count().",
    )
    parallel_jobs_group.add_argument(
        "-f",
        "--fraction",
        type=_parallel_fraction,
        default=None,
        metavar="F",
        help="Fraction of os.cpu_count() to use as the number of parallel jobs "
        "(e.g., 0.5 for half the CPUs). Must be in the range (0.0, 1.0]. "
        "Only valid with --parallel. Mutually exclusive with -j/--jobs. "
        "The resulting job count is max(1, int(cpu_count * F)).",
    )
```

If both -j and -f are passed, we get...

`test_rocgdb.py: error: argument -j/--jobs: not allowed with argument -f/--fraction`

Or...

`test_rocgdb.py: error: argument -f/--fraction: not allowed with argument -j/--jobs`

... depending on the order that it was passed.

aktemur: Ah, great, thanks.

aktemur (commented):
I haven't tested the feature but I read the code changes. They make sense and look good to me.

lumachad:
To give this a bit more context, we want to parallelize our CI test runs, but the amount of cores are variable across runners. If we pass a fraction like 0.25, that should be reasonable for all of the runners. We will have to experiment a bit to see what fraction works best.

lancesix (approved):
Looks reasonable to me.

PR #276, opened 2026-08-13
github: improve TheRock dependency update workflow

lumachad:
Move `THEROCK_COMMIT_REF` and the build container digest out of
`therock-ci-linux.yml` and into `.github/configs.json`, a plain data
file documented in `.github/configs.md`. A `resolve` job reads and
validates the pins and exposes them as outputs; the build job references
those outputs for `container.image` and `THEROCK_COMMIT_REF`. This lets
the automated update workflow push changes with only `contents: write`
on the `GITHUB_TOKEN`.

The update script is reworked accordingly:

- Fetch the TheRock branch HEAD and committer timestamp from the GitHub
  REST API (retried on transient failures) instead of `git ls-remote`
  and a separate commit lookup.
- Fetch the build container digest and creation timestamp via the OCI
  Distribution API, removing the dependency on `skopeo`.
- Skip opening a new PR if one labelled `therock-deps` is already open.
- Write a dry-run summary to stdout instead of exiting early.

A new `.github/scripts/config.py` helper reads and validates named
fields from `configs.json` and writes them to `GITHUB_OUTPUT`, following
the same pattern as `select_runner.py`. This avoids inline Python in the
workflow YAML, which breaks GitHub Actions' parser for reusable
workflows.

Fixes the YAML parse error that caused TheRock CI to fail for all PRs
after the original commit was merged.

### DIFF
--- a/.github/scripts/config.py
+++ b/.github/scripts/config.py
@@ -1,0 +1,76 @@
+# Copyright (c) Advanced Micro Devices, Inc., or its affiliates.
+# SPDX-License-Identifier: MIT
+"""Read and validate fields from .github/configs.json.
+
+Usage: python3 config.py <field> [<field> ...]
+
+Validates each requested field and writes field=value to $GITHUB_OUTPUT.
+Exits non-zero if the config file is missing, malformed, or any field
+fails validation.
+"""
+
+import json
+import os
+import re
+import sys
+from pathlib import Path
+
+# Maps each known field name to a (description, validator) pair.
+_VALIDATORS = {
+    "therock_commit_ref": (
+        "40-character lowercase hex SHA",
+        re.compile(r"^[0-9a-f]{40}$").match,
+    ),
+    "build_image": (
+        "ghcr.io image reference with sha256 digest",
+        re.compile(r"^ghcr\.io/[^@]+@sha256:[0-9a-f]{64}$").match,
+    ),
+}
+
+CONFIGS_PATH = Path(".github/configs.json")
+
+
+def main():
+    if len(sys.argv) < 2:
+        print("Usage: config.py <field> [<field> ...]", file=sys.stderr)
+        sys.exit(1)
+
+    fields = sys.argv[1:]
+
+    for field in fields:
+        if field not in _VALIDATORS:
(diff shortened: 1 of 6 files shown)


.github/scripts/config.py
```diff
@@ -0,0 +1,76 @@
...
+        sys.exit(1)
+    except json.JSONDecodeError as e:
+        print(f"{CONFIGS_PATH} is not valid JSON: {e}", file=sys.stderr)
+        sys.exit(1)
+
+    results = {}
+    for field in fields:
+        description, validate = _VALIDATORS[field]
+        value = d.get(field, "")
+        if not validate(value):
```
spatrang: Nit (non-blocking): if a field is present but non-string (e.g. `"therock_commit_ref": null` or a number), `validate(value)` calls `re.match` on a non-str and raises an uncaught `TypeError` traceback instead of the friendly `Invalid '...' (expected ...)` message just below. It still fails closed (non-zero exit), and `configs.json` is machine-written, so this is only a hand-edit hazard. Optional guard:

```python
if not isinstance(value, str) or not validate(value):
```

spatrang (approved):
LGTM.

This correctly fixes the CI-breaking YAML parse error from #262 by moving the inline `python3 -c` out of the `resolve` step into a standalone `.github/scripts/config.py` invoked as a one-liner. Reviewed:

- **`config.py`**: field allow-list validates the two CI-consumed pins with the same regexes as the old inline check; good fail-fast ordering (unknown-field check before file read), friendly `FileNotFoundError`/`JSONDecodeError` handling, writes to `$GITHUB_OUTPUT` and works locally. Functionally tested (valid fields, unknown field, no args, missing file all behave correctly).
- **Workflow**: `resolve` outputs (`therock_commit_ref`/`build_image`) are consistently wired into `container.image`, the `THEROCK_COMMIT_REF` env, the checkout `ref`, and the `therock_ref` output; redundant `export-ref` step removed.
- The other four files are unchanged from the already-approved #262.

One non-blocking nit inline. Approving.

.github/scripts/config.py
lumachad: Good catch, thanks. Applied the `isinstance(value, str)` guard.

PR #278, opened 2026-08-14
test_rocgdb.py: drop ROCM_PATH when --try-rocm-path is not passed

lumachad:
## Summary

- A system-wide `ROCM_PATH` can redirect LLVM tool lookups away from the artifacts tree configured by `setup_environment`, causing hard-to-diagnose test failures.
- `setup_environment` now accepts a `try_rocm_path` parameter and strips `ROCM_PATH` from the subprocess environment whenever the caller has not opted in via `--try-rocm-path`.

## Test plan

- [ ] Run `test_rocgdb.py` without `--try-rocm-path` on a system with `ROCM_PATH` set and verify LLVM tools resolve from the artifacts tree.
- [ ] Run `test_rocgdb.py` with `--try-rocm-path` and verify `ROCM_PATH` is still honoured.

### DIFF
--- a/.github/scripts/test_rocgdb.py
+++ b/.github/scripts/test_rocgdb.py
@@ -1368,12 +1368,16 @@ def setup_sanitizer_environment(env_vars: Dict[str, str], testsuite_dir: Path) -
     _log_aligned_fields(fields)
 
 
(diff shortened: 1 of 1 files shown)


lancesix (approved):
LGTM, thanks.

PR #279, opened 2026-08-14
[amd-staging][14-08-26] merge upstream-master

lancesix:
Merge upstream/master into amd-staging.

 Conflicts in bfd/version.h and gdb/python/python-internal.h are trivial.
 Just used upstream verison.
 
 Also adjust gdb.rocm/step-abort.exp to handle downstream differences
 (different outputs, and "allow_hip_tests" which replaced
 "allow_hipcc_tests").


lumachad (approved):
Conflict resolution looks OK to me.

Ok if CI is clean.

PR #282, opened 2026-08-14
gdb, amdgpu: pass process id to amd_dbgapi_convert_address_space

aktemur:
This is the companion PR to https://github.com/ROCm/rocm-systems/pull/10163.

### DIFF
--- a/gdb/configure.ac
+++ b/gdb/configure.ac
@@ -327,7 +327,7 @@ if test "$gdb_require_amd_dbgapi" = true \
   # stability until amd-dbgapi hits 1.0, but for convenience, still check for
   # greater or equal that version.  It can be handy when testing with a newer
   # version of the library.
-  PKG_CHECK_MODULES([AMD_DBGAPI], [amd-dbgapi >= 0.80.0],
+  PKG_CHECK_MODULES([AMD_DBGAPI], [amd-dbgapi >= 0.81.0],
 		    [has_amd_dbgapi=yes], [has_amd_dbgapi=no])
 
   if test "$has_amd_dbgapi" = "yes"; then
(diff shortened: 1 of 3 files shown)


gdb/configure.ac
```diff
@@ -327,7 +327,7 @@ if test "$gdb_require_amd_dbgapi" = true \
   # stability until amd-dbgapi hits 1.0, but for convenience, still check for
   # greater or equal that version.  It can be handy when testing with a newer
   # version of the library.
-  PKG_CHECK_MODULES([AMD_DBGAPI], [amd-dbgapi >= 0.80.0],
+  PKG_CHECK_MODULES([AMD_DBGAPI], [amd-dbgapi >= 0.81.0],
```
lancesix, line 330: I think we will need a transition phase for this…  otherwise CI will never pass on ROCm-systems.

i.e. we need to first have ROCgdb which has a `#if` to select between the old and new dbgapi interface.  Land this, and get it in TheRock.  Once this is picked up, we can proceed and merge the dbgapi part (with passing CI), and later on we can do cleanup in GDB and remove the `#if`.

This is a real nuisance, but unfortunately, I do not think we have much of a way around this.

Ideally, in GDB's patch (at least the initial, this can be removed from the cleanup), GDB's configury would check the API and select based on this rather than a version number (using something like `AC_TRY_COMPILE `).  This way, we do not need to assume what order patches will land in dbgapi.

lumachad: Could we introduce runtime checks instead of relying on a hardcoded minimum version?

lancesix:
Isn't this PR superseded by #289 ?

lumachad:
> Isn't this PR superseded by #289 ?

@aktemur Could you please confirm/deny?

aktemur:
> > Isn't this PR superseded by #289 ?
> 
> @aktemur Could you please confirm/deny?

Confirmed. This PR is obsolete. Closing.

PR #284, opened 2026-08-17
gdb.rocm/hip-builtin-completions: consider multi choices for completions

amd-shahab:
```
Current implementation of the test assumes one "suggestion" per
completion.  This can go awry if there are symbols that have
commonality in their names with the symbols that the test cares
about, like "threadIdx" and "threadIndex" [1].

This change adjust the checks as such to look for the desired
symbol name in the output of "complete" command, assuming that
there can be one or more entries.

[1]
Due to the missing step of stripping ".symtab" from libLLVM.so, a
"threadIndex" variable comes into GDB's scope.  This collides with
the expected tab-completion of "threadI" to become "threadIdx".
The symbol from libLLVM.so is loaded through these shared objects
dependencies:

  test
    libamdhip64.so
      libamd_comgr.so
        libclang-cpp.so
          libLLVM.so

Bug: LCOMPILER-2585
```

### DIFF
--- a/gdb/testsuite/gdb.rocm/hip-builtin-completions.exp
+++ b/gdb/testsuite/gdb.rocm/hip-builtin-completions.exp
@@ -51,8 +51,29 @@ proc all_hip_completions {} {
     foreach var {threadIdx blockIdx blockDim gridDim warpSize} {
 	# threadI blockI blockD gridD warpSi
 	set substr [string range $var 0 [expr {[string length $var] - 3}]]
-	gdb_test "complete print $substr" "print $var"
-	gdb_test "complete break main if $substr" "break main if $var"
+
+	set seen false
+	gdb_test_multiple "complete print $substr" "print $substr" {
+	    -re "print $var\r\n" {
+		set seen true
+		exp_continue
+	    }
+	    -re "$::gdb_prompt $" {
+		gdb_assert {$seen} $gdb_test_name
+	    }
+	}
+
+	set seen false
+	gdb_test_multiple "complete break main if $substr" \
+	    "break main if $substr" {
+	    -re "break main if $var\r\n" {
+		set seen true
+		exp_continue
+	    }
+	    -re "$::gdb_prompt $" {
+		gdb_assert {$seen} $gdb_test_name
+	    }
+	}
     }
 
     # ".x", ".y" and ".z" completions for three-dimensional built-ins.


gdb/testsuite/gdb.rocm/hip-builtin-completions.exp
```diff
@@ -48,12 +48,18 @@ proc no_hip_completions {} {
...
+    foreach var {blockIdx blockDim gridDim warpSize} {
+	# blockI blockD gridD warpSi
 	set substr [string range $var 0 [expr {[string length $var] - 3}]]
 	gdb_test "complete print $substr" "print $var"
 	gdb_test "complete break main if $substr" "break main if $var"
     }
+    # Look for completion of "threadId" into "threadIdx".
+    # Unlike the above tests, this one is narrower, becasue it can
+    # collide with libLLVM's parallel::threadIndex
+    # (llvm/lib/Support/Parallel.cpp).
```
lancesix: Didn't we discuss having the test check that `print threadIdx` was one of the proposed completion (and implicitly accept having other symbols show up)?

Also, it is really not clear to someone without context why libLLVM would come here.  Might be worth mentionning that this is a rocr -> comgr -> libllvm dependency.

lumachad: becasue -> because

I don't think we should mention the build/llvm bug here. We're just protecting ourselves from potential deviations from symbol availability.

amd-shahab:
> Didn't we discuss having the test check that `print threadIdx` was one of the proposed completion (and implicitly accept having other symbols show up)?

We did, ~but I don't know how to do it. Otherwise,~ it could be a final solution (not being affected by the environment).

UPDATE: Now it is done as we've discussed.

gdb/testsuite/gdb.rocm/hip-builtin-completions.exp
amd-shahab: > I don't think we should mention the build/llvm bug here. We're just protecting ourselves from potential deviations from symbol availability.

With this "solution", the bug is the sole reason of this change. We're not protecting ourselves from future issues this way. "Narrow"ing down is not the way to go. What if tomorrow there's another variable added, this time called `threadIdintity`!?

amd-shahab:
- Use `gdb_test_multiple` to look for our symbol-of-interest amongst a likely multi-line suggestion.
- Update the commit message to reflect the latest changes and the shared object dependencies.

lancesix (approved):
LGTM, thanks.

ronlieb:
@amd-shahab  has this been CP'ed to develop ?

lumachad:
> @amd-shahab has this been CP'ed to develop ?

No. It needs cherry-picking to amd-staging-rocgdb-* first then a rocgdb bump to reach TheRock main.

PR #286, opened 2026-08-18
github: add rocgdb-corefile to the test workflow

lumachad:
Add rocgdb-corefile to the list of projects tested in the TheRock test
packages workflow, alongside rocgdb-cpu and rocgdb-gpu.


lancesix (approved):
LGTM, thanks.

PR #287, opened 2026-08-18
handle clusters (alternative)

aktemur:
This is the same as PR #206, except that it uses the alternative dbgapi in PR https://github.com/ROCm/rocm-systems/pull/10241.

### DIFF
--- a/gdb/testsuite/gdb.rocm/cluster-barrier.cpp
+++ b/gdb/testsuite/gdb.rocm/cluster-barrier.cpp
@@ -1,0 +1,64 @@
+/* Copyright (C) 2026 Free Software Foundation, Inc.
+   Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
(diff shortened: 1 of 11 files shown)


gdb/testsuite/gdb.rocm/cluster-barrier.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+  /* Index of the calling block within cluster.  */
+  auto block_in_cluster = c.block_index ();
+
+  int bx = blockIdx.x;
+  int by = blockIdx.y;
+  int bz = blockIdx.z;
+
+  if (bx == 0)
+    NOP (1); /* BP1: before barrier arrive.  */
```
lancesix: might be worth having a sleep before the breakpoint, to ensure that other waves have plenty of time to reach the barrier_arrive.

aktemur:
The force-push above is a rebase.

PR #290, opened 2026-08-18
github: remove auto-assign PR workflow

lumachad:
Remove the auto-assign workflow that randomly assigns a reviewer to new PRs.


lumachad:
I think the auto-assignment workflow has served its purpose. Let's drop it and keep going with folks reviewing things as they can.

aktemur (approved):
No concerns from my side.

PR #291, opened 2026-08-19
testsuite: hip: fix -Wpointer-bool-conversion warnings in hip-driver.cc

lumachad:
Fix compilation warnings on Ubuntu 24.04 with newer Clang:

  warning: address of function 'm' will always evaluate to 'true'
  [-Wpointer-bool-conversion]

The kernel() function in hip-driver.cc checks which overload of the
weak symbol gdb_hip_test_main is defined by casting it and testing
the result in an if condition.  The original code used function
references (int (&)(...)), but a reference to a function can never
be null, so the condition always evaluates to true.

Fix this by using function pointers (int (*)(...)) instead.  A
function pointer for an undefined weak symbol is null, so the
if/else if chain correctly identifies the defined overload.


lancesix (approved):
LGTM, thanks.

PR #292, opened 2026-08-19
github: schedule therock deps update to run weekly on Mondays

lumachad:
## Summary

- Enable the scheduled trigger in `therock-deps-update.yml`, set to run every Monday at 03:17 UTC.
- Gate the `--dry-run` flag explicitly on `workflow_dispatch` so scheduled runs always execute for real.

## Test plan

- [ ] Verify the workflow appears under the scheduled triggers in the Actions tab after merge.
- [x] Trigger manually with `dry_run: true` to confirm the dry-run path still works.
- [x] Trigger manually with `dry_run: false` to confirm the real-run path works.

### DIFF
--- a/.github/workflows/therock-deps-update.yml
+++ b/.github/workflows/therock-deps-update.yml
@@ -1,9 +1,8 @@
 name: Update TheRock Dependencies
 
 on:
-  # schedule: # runs twice daily (03:17 and 15:17 UTC)
-  #   - cron: '17 3 * * *'
-  #   - cron: '17 15 * * *'
+  schedule: # runs weekly on Mondays at 03:17 UTC
+    - cron: '17 3 * * 1'
   workflow_dispatch:
     inputs:
       dry_run:
@@ -38,4 +37,4 @@ jobs:
 
       - name: Update TheRock references
         run: |
-          python3 .github/scripts/update_therock_deps.py ${{ inputs.dry_run && '--dry-run' || '' }}
+          python3 .github/scripts/update_therock_deps.py ${{ github.event_name == 'workflow_dispatch' && inputs.dry_run && '--dry-run' || '' }}


.github/workflows/therock-deps-update.yml
```diff
@@ -38,4 +37,4 @@ jobs:
 
       - name: Update TheRock references
         run: |
-          python3 .github/scripts/update_therock_deps.py ${{ inputs.dry_run && '--dry-run' || '' }}
+          python3 .github/scripts/update_therock_deps.py ${{ github.event_name == 'workflow_dispatch' && inputs.dry_run && '--dry-run' || '' }}
```
lancesix, line 40: shouldn't this be `( github.event_name == 'workflow_dispatch' || inputs.dry_run)` && '--dry-run'`?

i.e. use `--dry-run` if either dry_run is part of the inputs, or if dealing with `workfloaw_dispatch`?

lumachad: I don't think so. With your suggestion I think we'd have something like this:

- Manual dispatch + dry_run=false → (true || false) && '--dry-run' = --dry-run gets passed, which is wrong.

lancesix: indeed. But it feels odd to have the event type override the parameter.

lumachad: Feels odd but acceptable or feels odd and you want it changed? :-)

spatrang, line 40: Logic is right. A slightly clearer form that always compares the string input:
```
${{ github.event_name == 'workflow_dispatch' && github.event.inputs.dry_run == 'true' && '--dry-run' || '' }}
```
Not a blocker — both manual paths were already tested.

spatrang (approved):

.github/workflows/therock-deps-update.yml
lumachad: Thanks. Let's go with the current logic for now.

PR #296, opened 2026-08-20
gdb.rocm/step-schedlock-spurious-waves: fix failure at -O3

lumachad:
At -O3 the compiler collapses kern's loop body to fewer source line
mappings, causing the stepped wave to reach end_of_kernel in only 2
next commands instead of 5.  The test's next loop then fails to match
its expected pattern when the end_of_kernel breakpoint fires.

Apply __attribute__ ((optnone)) to kern, mirroring the same attribute
already used on end_of_kernel, to keep the loop intact under higher
optimization levels and ensure there are enough source lines to step
over.

### DIFF
--- a/gdb/testsuite/gdb.rocm/step-schedlock-spurious-waves.cpp
+++ b/gdb/testsuite/gdb.rocm/step-schedlock-spurious-waves.cpp
@@ -31,7 +31,9 @@ end_of_kernel ()
 {
 }
 
-__global__ void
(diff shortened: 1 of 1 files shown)


lumachad:
I've spotted this failing more recently. We should fix this to keep optimization test runs clean.

lumachad:
Dropping in favor of https://github.com/ROCm/ROCgdb/pull/302.

PR #299, opened 2026-08-20
testsuite: rocm.exp: Limit wave output logged by info_threads_get_wave_count

akondrat-amd:
On GPUs with many wavefronts, info_threads_get_wave_count floods the log with thousands of lines.  Print only the first and last 20 wave lines, replacing the omitted middle with a single summary line "... N waves omitted ...".

Key changes:

- log_user 0/1 suppresses raw expect output during matching
- Lines are collected into wave_lines instead of being logged on the fly
- After matching, first 20 and last 20 are emitted via verbose -log; middle rows get a single summary line

### DIFF
--- a/gdb/testsuite/lib/rocm.exp
+++ b/gdb/testsuite/lib/rocm.exp
@@ -421,18 +421,44 @@ proc gdb_compile_ocl {source dest type options} {
 }
 
 # Run "info threads INF_NUM", return the number of "AMDGPU Wave" threads found.
+# Print only the first and last 20 wave lines if there are more than 40 waves,
+# replacing the omitted middle with a single summary line "... N lines omitted ...".
 
 proc info_threads_get_wave_count { inf_num } {
     set wave_count 0
+    set wave_lines {}
+    set test_name "unknown"
+
+    log_user 0
     gdb_test_multiple "info threads ${inf_num}.*" "" {
 	-re "AMDGPU Wave\[^\r\n\]+\r\n" {
 	    incr wave_count
+	    lappend wave_lines [string trimright $expect_out(buffer) "\r\n"]
+	    exp_continue
+	}
+
+	-re "\[^\r\n\]+\r\n" {
+	    lappend wave_lines [string trimright $expect_out(buffer) "\r\n"]
 	    exp_continue
 	}
 
 	-re "$::gdb_prompt " {
-	    pass $gdb_test_name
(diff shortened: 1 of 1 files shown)


lumachad:
What problem exactly are we trying to solve? A timeout? A expect buffer overflow?

akondrat-amd:
> What problem exactly are we trying to solve? A timeout? A expect buffer overflow?

We count waves at least three times in a couple of tests, it adds thousands of lines to the log, particularly, 3 x 16,384. It's a nightmare to scroll through it.

lumachad:
Honestly I think we need a command to give us the number of waves we have. It is not feasible to count waves by going through a wall of text. For instance, if we run this with check-read1, it will be very slow.

gdb/testsuite/lib/rocm.exp
```diff
@@ -421,19 +421,45 @@ proc gdb_compile_ocl {source dest type options} {
 }
 
 # Run "info threads INF_NUM", return the number of "AMDGPU Wave" threads found.
+# Print only the first and last 20 wave lines if there are more than 40 waves,
+# replacing the omitted middle with a single summary line "... N lines omitted ...".
 
 proc info_threads_get_wave_count { inf_num } {
     set wave_count 0
+    set wave_lines {}
+    set test_name "unknown"
```
akondrat-amd, line 430: I use this variable to store the name of the test in order to print "pass" after we print the "wave_lines".

akondrat-amd:
This is how the output looks:
Screenshot 2026-08-21 at 10 45 38 AM

akondrat-amd:
> Honestly I think we need a command to give us the number of waves we have. It is not feasible to count waves by going through a wall of text. For instance, if we run this with check-read1, it will be very slow.

Running with check-read1 takes ~30 seconds longer with 3,000 waves regardless of this fix. 
We should have a line like "Total Waves:" with a default option to omit the lines in the middle.
```log
... (first 20 wave lines)
... 49,344 waves omitted ...
Total Waves: 49,384
... (last 20 wave lines)

lumachad:
@amd-shahab Thoughts here? I do not think we should try to handle this in TCL.

lumachad:
@akondrat-amd Please check if https://github.com/ROCm/ROCgdb/pull/327 solves this for you.

lumachad:
@lancesix Without #327, this is still a problem that I don't think we want to solve with the code in this PR.

PR #300, opened 2026-08-21
ci: restructure TheRock CI and add multi-arch GPU testing

lumachad:
## Summary

- Restructures TheRock CI to reduce redundancy: a single multiarch build artifact is shared across all test jobs
- Adds `ci:run-all-archs` label support: when set, the GPU test matrix expands from `gfx94X` only to all supported GFX families (`gfx90a`, `gfx950`, `gfx110X`, `gfx1151`, `gfx120X`)
- Adding or removing the label now re-triggers CI immediately without requiring a new push
- Default behavior (no label): unchanged, GPU tests run on `gfx94X` only

## Test plan

- [x] Default (unlabeled) CI runs on gfx94X only
- [x] `ci:run-all-archs` label triggers multi-arch test matrix
- [x] Failing arch legs block the PR (intended)

### DIFF
--- a/.github/workflows/therock-ci-linux.yml
+++ b/.github/workflows/therock-ci-linux.yml
@@ -84,6 +84,8 @@ jobs:
 
       - name: Fetch sources
         run: |
+          # debug-tools only requires core build infrastructure; the excluded groups
+          # (rocm-libraries, ml-frameworks, media-libs, math-libraries) are not needed.
           ./build_tools/fetch_sources.py --jobs 12 --depth 1 \
             --no-include-rocm-libraries \
             --no-include-ml-frameworks \
--- a/.github/workflows/therock-ci-linux.yml
+++ b/.github/workflows/therock-ci-linux.yml
@@ -12,38 +12,31 @@ on:
       build_runs_on:
         type: string
         default: aws-linux-scale-rocm-prod
+      therock_commit_ref:
+        type: string
+        required: true
+      build_image:
+        type: string
+        required: true
+    outputs:
+      therock_ref:
+        value: ${{ jobs.therock-build-linux.outputs.therock_ref }}
 
 permissions:
   contents: read
 
 jobs:
-  resolve:
-    name: Resolve TheRock references
-    runs-on: ubuntu-24.04
-    outputs:
-      therock_commit_ref: ${{ steps.read.outputs.therock_commit_ref }}
-      build_image: ${{ steps.read.outputs.build_image }}
-    steps:
-      - name: Check out repository
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-
-      - name: Read TheRock dependency pins
-        id: read
-        run: python3 .github/scripts/config.py therock_commit_ref build_image
-
   therock-build-linux:
     name: Build Linux Packages
-    needs: resolve
     runs-on: ${{ inputs.build_runs_on }}
     permissions:
       contents: read
       id-token: write
     outputs:
-      therock_ref: ${{ needs.resolve.outputs.therock_commit_ref }}
+      therock_ref: ${{ inputs.therock_commit_ref }}
     container:
-      image: ${{ needs.resolve.outputs.build_image }}
+      image: ${{ inputs.build_image }}
     env:
-      THEROCK_COMMIT_REF: ${{ needs.resolve.outputs.therock_commit_ref }}
       AMDGPU_FAMILIES: ${{ inputs.amdgpu_families }}
       TEATIME_FORCE_INTERACTIVE: 0
       CACHE_DIR: ${{ github.workspace }}/.container-cache
@@ -54,7 +47,7 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           repository: "ROCm/TheRock"
-          ref: ${{ needs.resolve.outputs.therock_commit_ref }}
+          ref: ${{ inputs.therock_commit_ref }}
 
       - name: Checkout CI config
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
@@ -91,7 +84,11 @@ jobs:
 
       - name: Fetch sources
         run: |
-          ./build_tools/fetch_sources.py --jobs 12
+          ./build_tools/fetch_sources.py --jobs 12 --depth 1 \
+            --no-include-rocm-libraries \
+            --no-include-ml-frameworks \
+            --no-include-media-libs \
+            --no-include-math-libraries
 
       - name: Configure Projects
         env:
@@ -141,17 +138,6 @@ jobs:
         run: |
           python3 build_tools/github_actions/post_build_upload.py \
             --run-id ${{ github.run_id }} \
-            --artifact-group ${{ env.AMDGPU_FAMILIES }} \
+            --artifact-group ${{ inputs.artifact_group }} \
             --build-dir build \
             --upload
-
-  therock-test-linux:
-    name: "Run Tests"
-    if: ${{ inputs.amdgpu_families == 'gfx94X-dcgpu' }}
-    needs: [therock-build-linux]
-    uses: ./.github/workflows/therock-test-packages.yml
-    with:
-      amdgpu_families: ${{ inputs.amdgpu_families }}
-      artifact_group: ${{ inputs.artifact_group }}
-      therock_ref: ${{ needs.therock-build-linux.outputs.therock_ref }}
-      artifact_run_id: ${{ github.run_id }}
--- a/.github/workflows/therock-ci.yml
+++ b/.github/workflows/therock-ci.yml
@@ -33,49 +33,106 @@ jobs:
         id: configure
         run: python .github/scripts/therock_configure_ci.py
 
-      - name: Checkout CI config
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-        with:
-          repository: ROCm/therock-ci-config
-          ref: main
-          path: ci-config
-        continue-on-error: true
-
       - name: Select build runner
         id: runner
         run: python3 .github/scripts/select_runner.py build_runs_on
 
-  therock-ci-linux:
-    name: TheRock CI Linux
+  resolve:
+    name: Resolve TheRock references
     needs: setup
+    if: ${{ needs.setup.outputs.enable_therock_ci == 'true' }}
+    runs-on: ubuntu-24.04
+    outputs:
+      therock_commit_ref: ${{ steps.read.outputs.therock_commit_ref }}
+      build_image: ${{ steps.read.outputs.build_image }}
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Read TheRock dependency pins
+        id: read
+        run: python3 .github/scripts/config.py therock_commit_ref build_image
+
+  therock-build-linux:
+    name: Build Linux Packages
+    needs: [setup, resolve]
     if: ${{ needs.setup.outputs.enable_therock_ci == 'true' }}
     permissions:
       contents: read
       id-token: write
-    strategy:
-      fail-fast: false
-      matrix:
-        amdgpu_family: [gfx94X-dcgpu]
     uses: ./.github/workflows/therock-ci-linux.yml
     with:
-      amdgpu_families: ${{ matrix.amdgpu_family }}
-      artifact_group: ${{ matrix.amdgpu_family }}
+      amdgpu_families: dcgpu-all;dgpu-all;igpu-all
+      artifact_group: gfx94X
       build_runs_on: ${{ needs.setup.outputs.build_runs_on }}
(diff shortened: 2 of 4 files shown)


.github/workflows/therock-ci.yml
```diff
@@ -33,49 +34,114 @@ jobs:
...
       - name: Select build runner
         id: runner
         run: python3 .github/scripts/select_runner.py build_runs_on
 
-  therock-ci-linux:
-    name: TheRock CI Linux
+      - name: Select GPU test families
+        id: families
+        run: |
+          if [[ "${{ contains(github.event.pull_request.labels.*.name, 'ci:run-all-archs') }}" == "true" ]]; then
```
spatrang, line 47: **[High]** When `ci:run-all-archs` is set, this expands the GPU test matrix — but note the trigger at L4–7 only fires on `opened`, `synchronize`, and `reopened`. Adding/removing the label alone will **not** re-run CI until the next push.

Consider adding `types: [opened, synchronize, reopened, labeled, unlabeled]` (or at least `labeled`), or document prominently that the label only takes effect on the next push.

**[High — same step]** When the label is set, any failing GPU matrix leg fails the entire `test-gpu` job and the PR check. That matches a probe, but the label cannot be used for informational-only multi-arch runs. Consider `continue-on-error: true` on `test-gpu` when probing, or document expected red CI during probing.

.github/workflows/therock-ci.yml
```diff
@@ -33,49 +34,114 @@ jobs:
         id: configure
         run: python .github/scripts/therock_configure_ci.py
 
-      - name: Checkout CI config
```
spatrang, line 36: **[Low]** Removed `ci-config` checkout from `setup`. Runner selection still passed in CI, so probably fine — please confirm `select_runner.py` no longer depends on this checkout.

.github/workflows/therock-ci.yml
```diff
@@ -33,49 +34,114 @@ jobs:
...
-      fail-fast: false
-      matrix:
-        amdgpu_family: [gfx94X-dcgpu]
     uses: ./.github/workflows/therock-ci-linux.yml
     with:
-      amdgpu_families: ${{ matrix.amdgpu_family }}
-      artifact_group: ${{ matrix.amdgpu_family }}
+      amdgpu_families: dcgpu-all;dgpu-all;igpu-all
+      artifact_group: gfx94X
       build_runs_on: ${{ needs.setup.outputs.build_runs_on }}
```
spatrang, line 80: **[Medium]** `artifact_group: gfx94X` is hardcoded here and again at L104, L118, and L132. The multi-arch build + single upload group design looks correct, but the magic string is duplicated.

Consider defining `artifact_group` once in `setup` (e.g. from `config.py`) and threading it through all callers.

.github/workflows/therock-ci.yml
```diff
@@ -33,49 +34,114 @@ jobs:
...
+      therock_commit_ref: ${{ needs.resolve.outputs.therock_commit_ref }}
+      build_image: ${{ needs.resolve.outputs.build_image }}
       extra_cmake_options: >
-        -DTHEROCK_ENABLE_ALL=OFF 
-        -DTHEROCK_BUILD_TESTING=ON 
+        -DTHEROCK_AMDGPU_DIST_BUNDLE_NAME=multiarch
+        -DTHEROCK_ENABLE_ALL=OFF
```
spatrang, line 86: **[Medium]** `-DTHEROCK_AMDGPU_DIST_BUNDLE_NAME=multiarch` is a behavioral change worth a brief comment explaining that GPU tests consume this bundle via the fixed `artifact_group: gfx94X` upload.

.github/workflows/therock-ci.yml
```diff
@@ -33,49 +34,114 @@ jobs:
...
+    needs: [setup, therock-build-linux]
+    if: ${{ needs.setup.outputs.enable_therock_ci == 'true' }}
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        amdgpu_family: ${{ fromJSON(needs.setup.outputs.test_families) }}
+    uses: ./.github/workflows/therock-test-packages.yml
+    with:
```
spatrang, line 105: **[Medium]** No `secrets: inherit` on this reusable workflow call (same for `therock-build-linux`, `test-cpu`, and `test-corefile`). Current CI run succeeded for artifact fetch, so not blocking today — but if `setup_test_environment` later needs secrets, these calls will fail opaquely.

Suggest adding `secrets: inherit` on all `uses:` calls for consistency.

.github/workflows/therock-ci.yml
```diff
@@ -33,49 +34,114 @@ jobs:
...
+  test-corefile:
+    name: Test corefile (gfx942)
+    needs: [setup, therock-build-linux]
+    if: ${{ needs.setup.outputs.enable_therock_ci == 'true' }}
+    permissions:
+      contents: read
+    uses: ./.github/workflows/therock-test-packages.yml
+    with:
+      amdgpu_families: gfx942
+      artifact_group: gfx94X
```
spatrang: **[Medium]** `amdgpu_families: gfx942` here vs `gfx94X` / family aliases elsewhere in this file. Likely intentional (only gfx942 supports core dumps), but a one-line comment would help future editors avoid "fixing" this to `gfx94X`.

.github/workflows/therock-ci.yml
```diff
@@ -33,49 +34,114 @@ jobs:
...
   therock_ci_summary:
     name: TheRock CI Summary
     if: always()
     needs:
       - setup
-      - therock-ci-linux
+      - therock-build-linux
+      - test-gpu
+      - test-cpu
+      - test-corefile
```
spatrang, line 149: **[High]** When `test-gpu` fails, the summary step only reports `test-gpu` as a failed job — not which matrix leg (e.g. gfx90a vs gfx110X) failed. Operators must dig into nested reusable-workflow logs.

Consider extending the summary to surface per-arch failures.

.github/workflows/therock-ci-linux.yml
```diff
@@ -12,38 +12,31 @@ on:
       build_runs_on:
         type: string
         default: aws-linux-scale-rocm-prod
+      therock_commit_ref:
```
spatrang, line 15: **[Medium]** New required inputs (`therock_commit_ref`, `build_image`) — any other caller of this reusable workflow must now supply them; resolve was previously internal. Fine if ROCgdb is the only caller; worth confirming.

.github/workflows/therock-ci-linux.yml
```diff
@@ -91,7 +84,11 @@ jobs:
 
       - name: Fetch sources
         run: |
-          ./build_tools/fetch_sources.py --jobs 12
+          ./build_tools/fetch_sources.py --jobs 12 --depth 1 \
+            --no-include-rocm-libraries \
```
spatrang, line 90: **[Medium]** Aggressive source-fetch pruning (`--no-include-rocm-libraries`, etc.) is a good speed win if debug-tools never pull those trees. Risk: a future TheRock dependency edge could break the build with a non-obvious missing-repo error.

Worth noting in the PR description which component groups debug-tools requires.

.github/workflows/therock-ci-linux.yml
```diff
@@ -141,17 +138,6 @@ jobs:
         run: |
           python3 build_tools/github_actions/post_build_upload.py \
             --run-id ${{ github.run_id }} \
-            --artifact-group ${{ env.AMDGPU_FAMILIES }} \
+            --artifact-group ${{ inputs.artifact_group }} \
```
spatrang, line 143: **[Positive]** Using `inputs.artifact_group` instead of `env.AMDGPU_FAMILIES` prevents uploading under `dcgpu-all;dgpu-all;igpu-all` after the multi-arch build change. This is an important fix — please keep.

.github/workflows/therock-test-packages.yml
```diff
@@ -108,6 +116,7 @@ jobs:
       artifact_run_id: ${{ inputs.artifact_run_id }}
       therock_ref: ${{ inputs.therock_ref }}
       amdgpu_families: ${{ inputs.amdgpu_families }}
+      artifact_group: ${{ inputs.artifact_group }}
```
spatrang, line 119: **[Medium]** Forwarding `artifact_group` here is required for the decoupling design. Without it, tests would look up artifacts under the wrong group when `amdgpu_families` differs from the build upload group.

.github/workflows/therock-test-component.yml
```diff
@@ -65,7 +67,7 @@ jobs:
         uses: './.github/actions/setup_test_environment'
         with:
           ARTIFACT_RUN_ID: ${{ env.ARTIFACT_RUN_ID }}
-          ARTIFACT_GROUP: ${{ inputs.amdgpu_families }}
+          ARTIFACT_GROUP: ${{ inputs.artifact_group != '' && inputs.artifact_group || inputs.amdgpu_families }}
```
spatrang, line 70: **[Medium]** Fallback to `inputs.amdgpu_families` when `artifact_group` is empty is correct for backward compatibility. All new callers pass `gfx94X` explicitly, so the fallback path is mostly legacy — looks fine.

spatrang (commented):
## Review summary

**Verdict: Approve with suggestions.** The build-once / test-many split is sound, and decoupling `artifact_group` from `amdgpu_families` (especially in `therock-ci-linux.yml`) prevents a real bug with the multi-arch build.

Default-path CI (gfx94X only) should stay green.

### Also noted (no diff hunk for inline comment)
- **`therock-ci.yml` L4–7:** `pull_request` trigger lacks `labeled` / `unlabeled` — see inline comment on the label check step.
- **`therock-test-component.yml` L73:** Pre-existing invalid `IS_PR_FROM_FORK` input to `setup_test_environment` (CI warning in logs). Worth a follow-up fix.

.github/workflows/therock-ci.yml
lumachad: Fixed — added `types: [opened, synchronize, reopened, labeled, unlabeled]` to the `pull_request:` trigger so adding/removing `ci:run-all-archs` re-runs CI immediately without requiring a push.

For the second part: intentional. We want failures on any arch to block the PR — the label is not meant for informational-only runs.

.github/workflows/therock-ci.yml
lumachad: `select_runner.py` treats `ci-config` as optional — it falls back to a default runner when the directory is not present (see the script's fallback path). Removing the checkout from `setup` is safe.

.github/workflows/therock-ci.yml
lumachad: Fixed — `artifact_group` is now emitted once from the `setup` job and referenced via `needs.setup.outputs.artifact_group` in all callers.

.github/workflows/therock-ci.yml
lumachad: Added a comment: a multi-arch build cannot use the default bundle name, which is derived from `amdgpu_families` and would not be usable as an artifact group name when building for multiple families simultaneously.

.github/workflows/therock-ci.yml
lumachad: Intentionally omitted. The reusable workflows have no secret dependencies — AWS credentials go through OIDC. Adding blanket `secrets: inherit` would unnecessarily expand the exposure surface. We will add specific secret forwarding if a concrete need arises.

.github/workflows/therock-ci.yml
lumachad: Added a comment: `gfx942` is intentional — we only have a `gfx942` corefile runner for rocgdb corefile testing.

.github/workflows/therock-ci.yml
lumachad: Acknowledged. Leaving as-is for now — the matrix leg names are visible in the GitHub Actions job list even if the summary only names the top-level job.

.github/workflows/therock-ci-linux.yml
lumachad: Confirmed — ROCgdb is the only caller of this reusable workflow.

.github/workflows/therock-ci-linux.yml
lumachad: Added a comment in the workflow: `debug-tools` only requires core build infrastructure; none of the pruned groups are needed.

.github/workflows/therock-ci-linux.yml
lumachad: Thanks for the confirmation.

.github/workflows/therock-test-packages.yml
lumachad: Thanks for confirming.

.github/workflows/therock-test-component.yml
lumachad: Thanks for confirming.

spatrang (approved):
**Approved.** The build-once / test-many restructure looks good and the follow-up push addresses the main review feedback (label trigger, centralized `artifact_group`, workflow comments).

Remaining non-blocking follow-ups:
1. Pre-existing invalid `IS_PR_FROM_FORK` input in `therock-test-component.yml` — fix in a follow-up PR.
2. Optional: richer summary output when `test-gpu` matrix legs fail.

The `ci:run-all-archs` probe behavior (red CI when any arch fails) is acceptable for now given the stated intent.

PR #301, opened 2026-08-21
gdb.rocm/hip-builtin-completions: simplify the checks

amd-shahab:
```
In a previous patch [1], gdb_test_multiple constructs were added
to handle cases that there could be more than one suggestion for
the "complete" command's output.

This change achieves the same goal with "gdb_test".

[1]
b5151b0b376 in amd-staging
```


lancesix (approved):
Should work as well.  Thanks for the simplification.

PR #307, opened 2026-08-24
Update CHANGELOG_AMD.md for therock-10.0 release

amd-shahab:
With this change the CHANGELOG_AMD file is updated to have more accurate entries.

### DIFF
--- a/CHANGELOG_AMD.md
+++ b/CHANGELOG_AMD.md
@@ -3,13 +3,24 @@
 Full documentation for ROCgdb is available at
 [rocm.docs.amd.com/rocgdb](https://rocm.docs.amd.com/projects/ROCgdb/en/latest/index.html)
 
-## ROCgdb-X for ROCm-next
+## ROCgdb-16-3 for ROCm-10.0
+
+### Added
+
+- The address space operator '#' is recognized in Fortran programs too.
+  This allows evaluating expressions like 'private_lane#0x08' in Fortran
+  applications that offload kernels to an AMD GPU.
+
+## ROCgdb-16-3 for ROCm-7.14
 
 ### Added
 
 - Dumping core of AMD GPU programs with the "gcore" command is now
   significantly faster, particularly for kernels that use small
   amounts of VRAM.
+- New "catch hiperr" command that stops the inferior when a HIP API
+  call returns an error.  The convenience variable `$_hiperr` holds
+  the error code at the catchpoint.
 
 ## ROCgdb-16-3 for ROCm-7.13
 


lumachad:
What is release/therock-10.0?

lumachad:
We should really update amd-staging-rocgdb-16, our stable release branch if we're missing entries there. After that is done, we should then cherry-pick those into this branch (with -x) . This should avoid the releases containing things our stable branch doesn't and creating divergence.

amd-shahab:
> We should really update amd-staging-rocgdb-16, our stable release branch if we're missing entries there. After that is done, we should then cherry-pick those into this branch (with -x) . This should avoid the releases containing things our stable branch doesn't and creating divergence.

but that part will still have `ROCgdb-X for ROCm-next`. That line should not end-up in a branch like this.

amd-shahab:
> What is release/therock-10.0?

I was hoping you'd know. It was brought to my attention by @SwRaw .

lumachad:
> > We should really update amd-staging-rocgdb-16, our stable release branch if we're missing entries there. After that is done, we should then cherry-pick those into this branch (with -x) . This should avoid the releases containing things our stable branch doesn't and creating divergence.
> 
> but that part will still have `ROCgdb-X for ROCm-next`. That line should not end-up in a branch like this.

Yeah, we should fix that first then.

HereThereBeDragons:
> What is release/therock-10.0?

when we branch of for therock we have a branching script that creates a for all the commits of all the subprojects of therock at the given point of the branching.

you then need cherry picks + bumps if you want to update the code. i dont think that doc classifies for a cherry pick for that particular release

amd-shahab:
> you then need cherry picks + bumps if you want to update the code

Isn't this supposed to remain frozen? More important question: qualifications and testing are done based on this branch, right?

marbre (commented):
> What is release/therock-10.0?

@lumachad, When we branch a ROCm release we cut branches in `rocm-systems`, `rocm-libraries`, our LLVM fork as well as in the ROCgdb fork. This is for the case we need to cherry-pick patches into a release (or later into a patch release) but so far there was no need to do this in ROCgdb, thus I assume it was unnoticed so far. If we cherry-pick into a release branch, this branch needs to be bumped in the corresponding release branch in TheRock. For ROCm 10, release candidates are out and the proposed change will not propagate into the final release. IMHO there is not much value in adding this here, assuming it it not used to generate public documentation.

> Isn't this supposed to remain frozen? More important question: qualifications and testing are done based on this branch, right?

@amd-shahab, This branch points to the code base used in the ROCm 10.0.0 release candidates, yes. The branch is not necessarily frozen but your commit would not go into the release as it is too late in the game. It would only go in if we would create a 10.0.1 release which we don't have plans for.

lumachad:
@SwRaw can clarify, but I think this is going to be used to generate external public documentation. We really don't need this change in code, but the current process is we take these from CHANGELOG_AMD.md.

So we may not even need this if we point @SwRaw to the proper changes? @marbre

amd-shahab:
based on @marbre 's explanations, this PR is closed.

PR #308, opened 2026-08-25
Rectify CHANGELOG_AMD entries

amd-shahab:
The "catch hiperr" MI feature did not make it to the ROCm-10.0. Therefore, its entry is moved out of "ROCgdb-16.3 for ROCm-10.0" and put under "ROCgdb-X for ROCm-next" section.


lumachad (approved):
This is OK. Thanks.

PR #311, opened 2026-08-27
.github: automerge: unshallow repo before merge-base

lumachad:
## Problem

The automerge workflow uses `actions/checkout` with the default
`fetch-depth: 1`, producing a shallow clone.  The script reuses that
repo (`--repo $WORKSPACE`), fetches full `sourceware/master` history,
then calls:

```
git merge-base origin/amd-staging sourceware/master
```

This fails with exit code 1 because the shallow clone does not contain
the common ancestor.  Example failure:
https://github.com/ROCm/ROCgdb/actions/runs/33000292878/job/98280035701

## Fix

After fetching `origin/TARGET_BRANCH`, check whether the repo is shallow
(`git rev-parse --is-shallow-repository`) and unshallow it if so.  When
the script manages its own clone the repo is never shallow, so this is a
no-op on that path.

### DIFF
--- a/.github/scripts/automerge.py
+++ b/.github/scripts/automerge.py
@@ -454,6 +454,21 @@ def main() -> None:
         print(f"Fetching origin/{TARGET_BRANCH}…")
         run_net(["git", "fetch", "origin", TARGET_BRANCH], cwd=repo)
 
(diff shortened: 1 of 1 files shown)


lumachad:
Self-approving as it is parallel to CI and a simple fix.

PR #316, opened 2026-09-01
test_rocgdb.py: add --emulate for GPU-less test runs

spatrang:
AIROCGDB-639

## What

`--emulate GPU` runs the gdb test suite on the [rocjitsu](https://github.com/ROCm/TheRock)
userspace GPU emulator instead of real hardware, so `gdb.rocm` can run on any
x86-64 Linux host with no AMD GPU, no `/dev/kfd` and no root.

```bash
python .github/scripts/test_rocgdb.py --emulate mi350x --gpu-tests --parallel
```

`GPU` is a mirage profile (`mi300x`, `mi350x`, `mi450x`) or the gfx target it
emulates (`gfx942`, `gfx950`, `gfx1250`), validated up front against the targets
the ROCm tree under test was built for.

## How it works

rocjitsu interposes the KFD ioctls in userspace, so an emulated run needs
nothing more than each `make check` launched inside a `mirage run` session. The
harness itself stays outside: configure, result parsing, ignore lists, retries
and exit codes are the hardware path, unchanged.

Session scope follows the execution mode:

* `--one-by-one`: a session per test, so no test can be affected by GPU state an
  earlier one left behind.
* `--parallel`: a session per job, by pointing dejagnu's `RUNTEST` at a wrapper,
  with each job given its own `GDB_LOCK_DIR` so the testsuite's
  `with_rocm_gpu_lock` stays in place but stops serializing unrelated jobs.
* neither: one session for the whole `make check`, as on hardware.

`--emulate` also defaults `--default-timeout` to 600s, since dejagnu's 10 second
per-testcase default fails the slower emulated tests for want of time rather
than for a real reason; ignores `--sanity-check` with a warning, since an
emulated GPU is built fresh per test and cannot wedge the way that probe exists
to catch (and `dmesg` is silent about a device living in userspace); and runs
`mirage cleanup` after a wall-clock timeout, which leaves mirage no chance to
tear its own session down.

Support code lives in a new module, `rocjitsu_emulator.py`, which resolves the
profile, finds `mirage` on the run's PATH (which starts with the tree's `bin`,
so a tree built with the emulation tools brings its own), and wraps a command in
a session; mirage then loads `librocjitsu.so` itself from the tree's lib on
`LD_LIBRARY_PATH`. It is imported lazily, so a hardware run never touches it and
an emulated run reports clearly when it is absent.

## The hardware path is untouched

The change to `test_rocgdb.py` is 152 added lines and 0 removed. No existing line
is modified, and every addition is reachable only through `--emulate`; with the
flag absent, the new code is `if emulator is not None` guards that fall through.

This was verified by running the upstream copy of the harness and this one side
by side from the same directory, across 13 invocation shapes: `--parallel`,
`-f 0.5`, `-j 4 --check-type check-read1`, `--one-by-one --sanity-check
--gpu-tests`, `--timing --parallel`, `--cpu-tests`, `--gpu-tests --quiet
--no-xfail`, and the invalid combinations that must be rejected (`-j` without
`--parallel`, `--one-by-one --parallel`, `--sanity-check` alone, `--sanity-check
--parallel`, `-f` with `-j`). Every one produced identical output. The only
difference anywhere is argparse re-wrapping its own usage line, because
`--emulate GPU` now appears in it.

`--fraction` is deliberately left exactly as it resolves for hardware.

## Results

Full `gdb.rocm` suite (76 files) against a gfx950 TheRock tree, on a CPU-only
host:

| Run | Wall clock | Expected passes |
|---|---|---|
| serial, one session | 923s | 2044 |
| `--parallel -j16`, one session | 746s | 2044 |
| `--parallel -j16`, session per job | **204s** | 2044 |

No new failures. Because a job now holds an emulated GPU rather than waiting for
one, memory scales with the job count (about 7GB, plus a little under 1GB per
job) while wall clock stops improving well before it does (307s at 4 jobs, 238s
at 8, 207s at 16, 196s at 32). So under `--emulate` an unspecified `--jobs` is
capped at 16 or at what memory allows, rather than left at the host's CPU count.
A `--jobs` you pass is always honoured, but flagged when the arithmetic says it
will not fit, because the OOM killer takes tests down in a way that reads as
failures rather than as an out-of-memory.

## Requirements

A ROCm tree carrying the emulator: a TheRock build configured with
`-D THEROCK_ENABLE_EMULATION=ON` (off by default, and not implied by
`THEROCK_ENABLE_ALL`), or a nightly dist tarball, which ships
`lib/librocjitsu.so`, `bin/mirage` and `tests/rocgdb` and so needs nothing built.
rocgdb, `amdclang++`, the emulator and the testsuite all come from that one tree,
so an emulated run is version-matched by construction.

`.github/scripts/EMULATED-TESTS.md` documents usage, requirements and
troubleshooting.

## Known follow-ups

* TheRock's install rule for `test_rocgdb.py` (`debug-tools/rocgdb`) does not
  ship `rocjitsu_emulator.py` yet, so `--emulate` works from a ROCgdb checkout
  but not from an installed tree. The harness says exactly that if you try. A
  small TheRock change closes it.
* TheRock's `rocgdb-gpu-emulated` matrix entry can call `test_rocgdb.py
  --emulate <profile>` directly, which makes the external
  `run_rocgdb_rocjitsu.py` launcher redundant.
* Emulator-versus-hardware gaps in individual tests should be seeded into an
  emulator-specific ignore list before any emulated run becomes a required PR
  check.

## Test plan

* [x] Full `gdb.rocm` suite emulated, serially and at `-j16`, on a gfx950 tree:
  2044 expected passes both ways.
* [x] Individual tests emulated in both `--one-by-one` and `--parallel` modes.
* [x] Profile and gfx-target spellings, and rejection of a profile the tree was
  not built for.
* [x] Emulator and mirage resolved from a nightly dist tarball, with no TheRock
  checkout present.
* [x] Hardware path output compared against upstream across the 13 invocation
  shapes listed above.
* [x] `black`, `flake8` (`gdb/setup.cfg`) and `codespell` clean; all added
  content is ASCII.
* [ ] Emulated run on a machine other than the author's, ideally in CI.


.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,104 @@
+# Running the ROCgdb tests on the rocjitsu emulator
```
lumachad: ## EMULATED-TESTS.md: too long

This doc is 104 lines explaining implementation details that mostly belong in the code or PR description. A reader reaching for it wants to know how to invoke the flag, what it requires, and what to do when something breaks.

Suggested structure (~40 lines):

```markdown
# Running the ROCgdb tests without hardware

`--emulate GPU` runs `gdb.rocm` on the rocjitsu userspace emulator with no AMD GPU.
GPU is a mirage profile (`mi300x`, `mi350x`, `mi450x`) or its gfx target (`gfx942`,
`gfx950`, `gfx1250`); the tree must have been built for it.

## Usage

```bash
python test_rocgdb.py --emulate mi350x --gpu-tests
python test_rocgdb.py --emulate mi350x --gpu-tests --parallel
```

## Requirements

- x86-64 Linux. No AMD GPU, no `/dev/kfd`, no root.
- A ROCm tree with the emulator: TheRock built with `-D THEROCK_ENABLE_EMULATION=ON`,
  or a nightly dist tarball (ships `librocjitsu.so` and `mirage` already).
- `dejagnu` and a C compiler, as on hardware.

| Profile  | gfx target |
|----------|------------|
| `mi300x` | `gfx942`   |
| `mi350x` | `gfx950`   |
| `mi450x` | `gfx1250`  |

## Notes

- `--default-timeout` is raised to 600 s automatically; emulated tests are slower.
- `--parallel` is recommended: each job gets its own emulated GPU.
  Memory scales with job count (~8 GB base + ~1 GB/job); `--jobs` is capped
  accordingly and warned when it will not fit.
- `--sanity-check` is ignored (not meaningful for an emulated device).

## Troubleshooting

- **`--emulate needs rocjitsu_emulator.py`**: use a ROCgdb checkout, not an installed tree.
- **`no librocjitsu.so`**: rebuild with `-D THEROCK_ENABLE_EMULATION=ON`.
- **`no mirage found`**: set `$MIRAGE_BIN` or put `mirage` on `$PATH`.
- **Tests time out**: raise `--default-timeout`.
```

Specific cuts:
- **Intro**: drop the KFD ioctl interposition sentence. That is implementation detail for `rocjitsu_emulator.py`'s module docstring.
- **Nightly tarball paragraph**: move to a single bullet under Requirements.
- **"What `--emulate` changes"**: keep as bullets but shorten each. The `mirage cleanup` detail after a wall-clock timeout is already in `Emulator.cleanup()`'s docstring.
- **"Parallel runs" section**: drop entirely. The GPU-lock mechanism and memory arithmetic are covered by `parallel_jobs()`'s docstring. The user only needs to know the headline numbers, which fit in one bullet under Notes.
- **`librocjitsu.so` / `mirage` resolution paragraph** (lines 63-65): redundant with the Troubleshooting bullets.

File: .github/scripts/EMULATED-TESTS.md [hunk 2]
```diff
@@ -0,0 +1,104 @@
...
+## Usage
+
+```bash
+# from an installed ROCm tree, which is the harness's own location
+python <rocm>/tests/rocgdb/test_rocgdb.py --emulate mi350x --gpu-tests
+
+# from a ROCgdb checkout, testing a tree built elsewhere
+ROCM_PATH=~/therock/build/dist/rocm \
+    python .github/scripts/test_rocgdb.py --emulate mi350x --try-rocm-path \
+    --gpu-tests --parallel
```
lumachad: Let's not use ROCM_PATH but rather OUTPUT_ARTIFACT_DIR.

.github/scripts/test_rocgdb.py
```diff
@@ -17,10 +17,18 @@
...
 from pathlib import Path
 from typing import Dict, List, NoReturn, Optional, Set, Tuple, Union
 
+if typing.TYPE_CHECKING:
+    # Annotations only: the emulator support module is imported lazily, so that
+    # a hardware run never needs it (see _load_emulator_support).
+    from types import ModuleType
+
+    from rocjitsu_emulator import Emulator
+
```
lumachad: Is this required here? Why?

.github/scripts/test_rocgdb.py
```diff
@@ -550,6 +586,22 @@ def parse_arguments() -> argparse.Namespace:
...
+        "AMD GPU and no /dev/kfd needed. GPU is a mirage profile (mi300x, "
+        "mi350x, mi450x) or the gfx target it emulates (gfx942, gfx950, "
+        "gfx1250), and the ROCm tree under test has to have been built for it. "
+        "The emulator (librocjitsu.so) comes from that tree; mirage from "
+        "$MIRAGE_BIN, else the tree, else a cargo build in the TheRock "
+        "checkout holding the tree, else $PATH. Each `make check` runs in "
+        "its own emulated GPU session, so with --one-by-one every test gets a "
+        "clean GPU. Implies a longer --default-timeout, since emulated tests "
+        "are much slower than on hardware.",
+    )
```
lumachad: Let's trim this down. It is too verbose.

.github/scripts/test_rocgdb.py
```diff
@@ -1976,6 +2081,22 @@ def run_tests(
...
                 cmd += ["FORCE_PARALLEL=1", f"-j{jobs}"]
 
+            if emulator is not None:
+                if args.parallel:
+                    # Each parallel job gets an emulated GPU (and a GPU lock) of
+                    # its own, instead of all of them queueing on the single GPU
+                    # of one session. `make` itself stays outside: it only
+                    # creates directories and collates results. The path is
+                    # quoted because make expands $(RUNTEST) into a shell
+                    # recipe, and $TMPDIR decides where it lives.
```
lumachad: Let's trim some of these comments so they are to the point and meaningful.

.github/scripts/test_rocgdb.py
```diff
@@ -2253,6 +2374,27 @@ def main() -> None:
...
+                args.emulate,
+                rocm_dir,
+                checkout=emulator_support.find_checkout(rocm_dir),
+            )
+        except emulator_support.EmulatorError as e:
+            _log_error_and_exit(str(e))
+        emulator.apply_env(env_vars)
+        print_section("Emulated GPU (no hardware in use)")
+        _log_aligned_fields(emulator.summary())
+
```
lumachad: I think we should only look for mirage. If that's available in our PATH, like clang++ etc, then we should be safe to assume rocjitsu is available. Let's add this check next to the check for other pre-requisite executables.

.github/scripts/rocjitsu_emulator.py
```diff
@@ -0,0 +1,425 @@
...
+    """Return the TheRock checkout at or above `start`, if there is one.
+
+    A build directory and the dist tree inside it both live in the checkout, so
+    the checkout that owns a path is that path or one of its parents.
+    """
+    start = Path(start).resolve()
+    for directory in [start, *start.parents]:
+        if (directory / CHECKOUT_MARKER).is_file():
+            return directory
+    return None
```
lumachad: We should locate mirage in the ROCm tree the user pointed at.

.github/scripts/rocjitsu_emulator.py
```diff
@@ -0,0 +1,425 @@
...
+        return script
+
+    def apply_env(self, env_vars: Dict[str, str]) -> None:
+        """Add what the emulated run needs to an environment.
+
+        ROCJITSU_LIB is the tie between the two halves: mirage resolves the
+        emulator library from its own environment rather than the session's, so
+        a mirage from outside the tree has to be told which rocjitsu to load.
+        """
+        env_vars["ROCJITSU_LIB"] = str(self.rocjitsu)
```
lumachad: With the right paths set, I'd expect mirage to be able to discover the library automatically. Do we really need this env var?

.github/scripts/rocjitsu_emulator.py
```diff
@@ -0,0 +1,425 @@
...
+        env_vars["HCC_AMDGPU_TARGET"] = self.arch
```
lumachad: Do we really need this env var to tell rocjitsu what to emulate?

.github/scripts/rocjitsu_emulator.py
```diff
@@ -0,0 +1,425 @@
...
+            "# emulated GPU of its own, with a GPU lock of its own.\n"
+            f"lock_dir=$(mktemp -d {template}) || exit 1\n"
+            "trap 'rm -rf \"$lock_dir\"' HUP INT TERM\n"
+            f'{session} "GDB_LOCK_DIR=$lock_dir" "$@"\n'
+            "status=$?\n"
+            'rm -rf "$lock_dir"\n'
+            "exit $status\n",
+            encoding="utf-8",
+        )
+        script.chmod(0o755)
```
lumachad: Let's not do this here. If we need execution permission, let's configure the file like so.

.github/scripts/rocjitsu_emulator.py
```diff
@@ -0,0 +1,425 @@
...
+            "#!/bin/sh\n"
+            "# Generated by rocjitsu_emulator.py: run one dejagnu job on an\n"
+            "# emulated GPU of its own, with a GPU lock of its own.\n"
+            f"lock_dir=$(mktemp -d {template}) || exit 1\n"
+            "trap 'rm -rf \"$lock_dir\"' HUP INT TERM\n"
+            f'{session} "GDB_LOCK_DIR=$lock_dir" "$@"\n'
+            "status=$?\n"
+            'rm -rf "$lock_dir"\n'
+            "exit $status\n",
+            encoding="utf-8",
```
lumachad: What is this used for?

lumachad (commented):
A few early comments. I'll give this another pass later.

.github/scripts/EMULATED-TESTS.md
spatrang: Done. Rewritten to your structure, 104 lines down to 57: intro, Usage,
Requirements with the profile table, Notes, Troubleshooting.

The KFD interposition sentence, the resolution paragraph and the whole
"Parallel runs" section are gone. I kept the headline parallel numbers as one
bullet under Notes, since that is what tells a reader whether --parallel is
worth their time, and two troubleshooting entries beyond your list because both
cost real debugging time: a tree with no emulator, and mirage failing to seed
its config when $HOME is not writable.

.github/scripts/EMULATED-TESTS.md
spatrang: Done, with one detail: the variable the harness reads is OUTPUT_ARTIFACTS_DIR
(plural "ARTIFACTS"), so that is what the examples use. They now export it once
at the top, which also let --try-rocm-path drop out, since it was only there to
make ROCM_PATH take effect.

.github/scripts/test_rocgdb.py
spatrang: Not required, and removed along with "import typing". It existed only so
Optional["Emulator"] and the ModuleType return annotation resolved for a type
checker; nothing evaluates them at runtime, since the module is imported lazily.

Two small consequences: _load_emulator_support() now returns types.ModuleType
with a plain "import types", and the two emulator parameters lost their
annotations, because pyflakes reports F821 for a string annotation naming a
class it cannot see. Both stay documented in the existing Args docstrings.

.github/scripts/test_rocgdb.py
spatrang: Trimmed from ten lines to six, and it now points at EMULATED-TESTS.md instead
of repeating it. The mirage resolution order is gone from the help text, which
your later comments made obsolete anyway.

.github/scripts/test_rocgdb.py
spatrang: Trimmed: seven lines down to three for the parallel branch, three down to one
for the serial one. What is kept is the two things the code cannot show, namely
that a session per job is what stops the jobs queueing on one emulated GPU, and
why the path is quoted (make expands $(RUNTEST) into a shell recipe).

.github/scripts/test_rocgdb.py
spatrang: Done, and it removed more code than I expected: 58 lines out of the module.

mirage is now appended to required_executables and checked by
check_executables() alongside make, amdclang++ and runtest, so a missing mirage
is reported in the same section and format as any other missing tool. Emulator
resolution moved to just after that check. The librocjitsu.so lookup went with
it, as you suggested, so find_rocjitsu() is deleted; mirage's own diagnostic is
good when the library is genuinely absent, since it reports how many locations
it searched and every variable that would fix it.

One detail: $MIRAGE_BIN can name a mirage that is not on PATH at all, which is
how TheRock CI uses it, so the PATH check is skipped when that variable is set.
Without that the override stopped working, which I only caught by testing it.

.github/scripts/rocjitsu_emulator.py
spatrang: This is already what happens, and it is why the previous comment collapses so
neatly. setup_environment() prepends <tree>/bin to the PATH used for the run,
so a single shutil.which("mirage", path=env_vars["PATH"]) finds the tree's own
mirage first and falls back to the caller's PATH only when the tree ships none.
A tree with bin/mirage, such as a nightly dist tarball, therefore always wins.

find_checkout() and the cargo build discovery are deleted. The consequence is
that a TheRock build tree, which does not install mirage today, now needs
mirage on PATH or $MIRAGE_BIN, and says so. That seems right: TheRock
installing mirage is the proper fix, not this harness guessing at cargo target
directories.

.github/scripts/rocjitsu_emulator.py
spatrang: You are right, removed. I checked it rather than reasoning about it:

    # no ROCJITSU_LIB, LD_LIBRARY_PATH=<tree>/lib
    $ mirage run --profile mi350x -- rocminfo
      Name: gfx950
      Name: amdgcn-amd-amdhsa--gfx950:sramecc+:xnack-

    # no ROCJITSU_LIB, no LD_LIBRARY_PATH
    $ mirage run --profile mi350x -- rocminfo
    error: rocjitsu: KMD preload library (librocjitsu.so) not found [...]
    Set one of: ROCJITSU_LIB=..., LD_LIBRARY_PATH=..., ROCM_HOME=..., ROCM_PATH=...

setup_environment() already prepends <tree>/lib to LD_LIBRARY_PATH, which is
one of the mechanisms mirage itself names, so the variable was redundant.

.github/scripts/rocjitsu_emulator.py
spatrang: No, and it never told rocjitsu anything: it tells the testsuite which target to
compile the HIP programs for. hcc_amdgpu_targets in lib/rocm.exp uses it when
set and otherwise calls find_amdgpu_devices, which enumerates what it can see.
Inside the session that finds the emulated GPU, so the right target is
discovered without being told.

Removed and verified: the gdb.log of an emulated run shows offload-arch=gfx950
chosen by the probe, with neither variable appearing anywhere in it. That
emptied apply_env(), so it is deleted too.

.github/scripts/rocjitsu_emulator.py
spatrang: Your suggestion did not come through on my side (no diff attached), so I took
the intent as setting the mode when creating the file rather than patching it
afterwards:

    with os.fdopen(
        os.open(script, os.O_WRONLY | os.O_CREAT | os.O_EXCL, 0o755),
        "w", encoding="utf-8",
    ) as out:
        out.write(...)

The file is created executable in one step, O_EXCL refuses to write over
anything already there, and the separate chmod is gone. One behaviour change
worth naming: this honours umask, where chmod forced 0755 unconditionally. That
is harmless, since dejagnu execs the script as the user that created it. If you
meant another form, please re-attach the suggestion and I will follow it.

.github/scripts/rocjitsu_emulator.py
spatrang: It is what makes --parallel worth anything under emulation.

The testsuite serializes GPU-touching test bodies with with_rocm_gpu_lock,
because on hardware there is one GPU for all jobs to share. Wrapping the whole
make check in one session recreates exactly that: 16 jobs, one emulated GPU and
the same lock, which measured 746s against 923s serially, so almost no gain.

An emulated GPU costs a process rather than a socket, so each job can have one.
Dejagnu runs one $(RUNTEST) per test file, so pointing RUNTEST at this script
means every job starts its own session, and the GDB_LOCK_DIR line gives each
job its own lock directory, so the lock still protects a job from itself but no
longer serializes unrelated jobs. Same suite at -j16: 204s, with the same 2044
expected passes.

The runtest_wrapper() docstring says this; happy to shorten it or move the
explanation into EMULATED-TESTS.md if you would rather it lived there.

.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,57 @@
+# Running the ROCgdb tests without hardware
+
+`--emulate GPU` runs `gdb.rocm` on the rocjitsu userspace emulator with no AMD
```
lumachad: I think it's misleading to say we run gdb.rocm. This is determined by whatever --tests or similar options are passed to test_rocgdb.py.

.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,57 @@
...
+GPU. GPU is a mirage profile (`mi300x`, `mi350x`, `mi450x`) or its gfx target
+(`gfx942`, `gfx950`, `gfx1250`); the tree must have been built for it.
+
+## Usage
+
+`--emulate` needs `rocjitsu_emulator.py`, which ships with the harness in a
+ROCgdb checkout but not in an installed tree, so the ROCm tree under test is
+named by `OUTPUT_ARTIFACTS_DIR` rather than found next to the script:
```
lumachad: We will need to have this in an install tree going forward.

.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,57 @@
...
+
+```bash
+export OUTPUT_ARTIFACTS_DIR=~/therock/build/dist/rocm
```
lumachad: This is correct, but a better practice is to not set this globally but to pass it for the specific command.

.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,57 @@
...
+
+python .github/scripts/test_rocgdb.py --emulate mi350x --gpu-tests
+python .github/scripts/test_rocgdb.py --emulate mi350x --gpu-tests --parallel
+```
+
+## Requirements
+
+- x86-64 Linux. No AMD GPU, no `/dev/kfd`, no root.
```
lumachad: Do we have plans to support it on Windows at some point?

.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,57 @@
...
+- A ROCm tree with the emulator: TheRock built with
+  `-D THEROCK_ENABLE_EMULATION=ON` (off by default, and not implied by
+  `THEROCK_ENABLE_ALL`), or a nightly dist tarball, which ships
```
lumachad: This is wrong. THEROCK_ENABLE_ALL ON definitely enables rocjitsu.

.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,57 @@
...
+| `mi300x` | `gfx942` |
+| `mi350x` | `gfx950` |
+| `mi450x` | `gfx1250` |
+
+## Notes
+
+- `--default-timeout` is raised to 600s automatically; emulated tests are
+  slower.
+- `--parallel` is recommended: each job gets its own emulated GPU, which on the
+  76-file `gdb.rocm` suite meant 204s at `-j16` against 923s serially, with the
```
lumachad: Let's not mention absolute numbers that get stale quickly.

.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,57 @@
...
+  same 2044 expected passes. Memory scales with job count (~8GB base, ~1GB per
+  job), so `--jobs` is capped accordingly and warned when it will not fit.
+- `--one-by-one` gives each test a session, and so a clean GPU, of its own.
+- `--sanity-check` is ignored (not meaningful for an emulated device).
+
+## Troubleshooting
+
+- **`--emulate needs rocjitsu_emulator.py`**: use a ROCgdb checkout, not an
+  installed tree.
```
lumachad: That's not what we're aiming for. We want to make these available in the ROCm tree after a build or a nightly tarball. We don't want to have to checkout sources to be able to run emulated tests.

.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,57 @@
...
+- **`no mirage found`**: use a tree that ships `bin/mirage`, put one on `$PATH`,
+  or set `$MIRAGE_BIN`.
```
lumachad: Misleading, we should not require the user setting PATH entries manually or other env variables other than OUTPUT_ARTIFACTS_DIR. test_rocgdb.py should be able to configure everything for the user.

.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,57 @@
...
+- **mirage cannot find `librocjitsu.so`**: the tree has no emulator; rebuild it
+  with `-D THEROCK_ENABLE_EMULATION=ON`.
```
lumachad: As mentioned above, this is incorrect. The emulator should always be available, unless we're aiming at a custom minimal ROCm tree build to work on debug-tools. In the latter case we need to build TheRock with -DTHEROCK_ENABLE_EMULATION=ON.

.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,57 @@
...
+- **mirage cannot write its config**: it seeds profiles on first run under
+  `$XDG_CONFIG_HOME/mirage`; set `$MIRAGE_CONFIG` somewhere writable when
+  `$HOME` is not.
```
lumachad: Is there a realistic scenario where this might happen? We always want to run mirage as regular users in our home directories. There shouldn't be access restrictions there.

.github/scripts/test_rocgdb.py
```diff
@@ -550,6 +579,18 @@ def parse_arguments() -> argparse.Namespace:
...
+        "--emulate",
+        type=str,
+        default=None,
+        metavar="GPU",
+        help="Run the tests on an emulated GPU (rocjitsu) instead of real "
+        "hardware; no AMD GPU needed. GPU is a mirage profile (mi300x, "
+        "mi350x, mi450x) or the gfx target it emulates (gfx942, gfx950, "
+        "gfx1250), which the ROCm tree must have been built for. Raises "
+        "--default-timeout, since emulated tests are slower. "
+        "See EMULATED-TESTS.md.",
```
lumachad: We should reference EMULATED-TESTS.md from the emulator file, not this one.

.github/scripts/test_rocgdb.py
```diff
@@ -2283,8 +2388,26 @@ def main() -> None:
     required_executables = ["make", *toolchain_executables, "runtest"]
     if has_rocm_tests:
         required_executables.insert(1, "amdclang++")
+    # $MIRAGE_BIN names a mirage outright, and so need not be on PATH at all.
+    if args.emulate is not None and not os.getenv("MIRAGE_BIN"):
+        required_executables.append("mirage")
     check_executables(required_executables, env_vars)
 
+    # Resolve the emulated GPU, if one was asked for. mirage comes from
+    # $MIRAGE_BIN, else the PATH checked above, which starts with the bin of
```
lumachad: I think MIRAGE_BIN here is an implementation detail of the emulator module and shouldn't be mentioned here.

.github/scripts/EMULATED-TESTS.md
spatrang: Good point, thanks. The doc now says it runs the tests on the emulator
instead of real hardware, and that which tests run is still up to `--tests`,
`--gpu-tests` and the rest, exactly as on hardware.

.github/scripts/EMULATED-TESTS.md
spatrang: Agreed on the goal, and thanks for raising it. The shape of it is small:
TheRock's `debug-tools/rocgdb/CMakeLists.txt` already installs
`test_rocgdb.py` into `tests/rocgdb` and, when present,
`rocgdb_ignore_list.json` beside it, so `rocjitsu_emulator.py` would want the
same optional-file block:

```cmake
if(EXISTS "${ROCGDB_SOURCE_DIR}/.github/scripts/rocjitsu_emulator.py")
  install(
    FILES ${ROCGDB_SOURCE_DIR}/.github/scripts/rocjitsu_emulator.py
    DESTINATION tests/rocgdb
    COMPONENT tests
  )
endif()
```

Nothing would change on this side when that lands: `_load_emulator_support()`
imports from the script's own directory, so the module works as soon as it sits
next to the harness, and an installed tree then needs no configuration at all,
since it is the tree the harness lives in.

If you agree, I would prefer to propose it here rather than raise it right
away. The rocjitsu side is still moving, and installing the module into every
ROCm tree commits us to an interface I would like to see settle first. One
thing worth noting either way: a tree installs whatever harness it was built
from, and the `tests/rocgdb/test_rocgdb.py` in my current tree predates
`--emulate` and `--one-by-one` both, so emulated runs from an installed tree
will need a tree built from a ROCgdb carrying this change, not only the install
rule.

If you would rather have it working from an installed tree sooner, two other
options: fold the module's contents into `test_rocgdb.py`, which works today
because that file is already installed, at the cost of some 340 lines in the
shared harness; or have TheRock's emulated CI leg invoke the harness from a
ROCgdb checkout, which is how it runs today in any case. My preference is the
install rule once rocjitsu settles, but I am happy to follow whichever you
prefer.

.github/scripts/EMULATED-TESTS.md
spatrang: Fair point, thanks. It is now `OUTPUT_ARTIFACTS_DIR=... python ...` on one
line, with no `export`.

.github/scripts/EMULATED-TESTS.md
spatrang: Not as things stand, I am afraid. In TheRock's `BUILD_TOPOLOGY.toml` the
`rocjitsu`, `rocjitsu-hotswap` and `mirage` artifacts are all
`disable_platforms = ["windows"]`, and rocjitsu emulates a GPU by interposing
KFD ioctls, which has no Windows equivalent. The doc now simply says Linux and
makes no claim about the future, so it will not need revisiting if that
changes. Happy to raise it with the rocjitsu folks if it would be useful to
know their plans.

.github/scripts/EMULATED-TESTS.md
spatrang: You are quite right, and thank you for the correction. TheRock's
`CMakeLists.txt` has:

```cmake
option(THEROCK_ENABLE_ALL "Enables building of all feature groups" ON)
option(THEROCK_ENABLE_EMULATION "Enable building of emulation tools" "${THEROCK_ENABLE_ALL}")
```

so emulation is on in any normal build. The requirement now reads that a normal
TheRock build and the nightly tarballs both carry the emulation tools, and that
only a minimal tree built to work on the debug tools alone needs
`-D THEROCK_ENABLE_EMULATION=ON`.

.github/scripts/EMULATED-TESTS.md
spatrang: Agreed, thanks. I have taken it a little further than the line you marked:
the timings are out of the module's comments as well, not only the doc. What
remains is the memory arithmetic the job cap actually computes, and the numbers
in its messages, which are measured on the machine at run time rather than
written down.

.github/scripts/EMULATED-TESTS.md
spatrang: Understood, and agreed as the goal. The "use a checkout, not an installed
tree" bullet is gone, and the doc now says plainly that once a ROCm tree carries
the module the run needs nothing configured.

To be straight about today: only `test_rocgdb.py` is installed into
`tests/rocgdb`, so `--emulate` from an installed tree still reports the module
as missing. The doc therefore documents what works now and names the end state
rather than claiming it. My reply on the install-tree comment above has the
TheRock block that would close the gap, and my thinking on when to raise it.

.github/scripts/EMULATED-TESTS.md
spatrang: Agreed, and thanks: this deleted the knob rather than the sentence.
`$MIRAGE_BIN` is gone from the module, from `test_rocgdb.py` and from the doc.
mirage is found only on the PATH the harness itself prepares, which starts with
the bin of the tree under test, so a normal tree needs nothing set, and
overriding a mirage is now the same as overriding `clang++`, namely PATH.

One correction to what I told you yesterday, with apologies: I said
`$MIRAGE_BIN` was how TheRock's CI names an out-of-tree mirage. That was wrong.
TheRock never sets it; the convention comes from rocjitsu's own demo scripts.
Nothing in TheRock or this repo sets it, which is what made it safe to
remove.

.github/scripts/EMULATED-TESTS.md
spatrang: Agreed, thanks, and this follows from the `THEROCK_ENABLE_ALL` correction.
The librocjitsu troubleshooting entry is gone, and the mirage one now says the
run is either pointed at a tree that is not a ROCm build carrying the emulation
tools, or at one built without them, which wants
`-D THEROCK_ENABLE_EMULATION=ON`. That is the minimal-build case you
describe.

.github/scripts/EMULATED-TESTS.md
spatrang: Not for the way we run it, no, so I have removed it. I ran into it in a
container with no writable `$HOME`, which on reflection is not a case worth
documenting here.

.github/scripts/test_rocgdb.py
spatrang: Agreed, thanks. The reference is now the last line of
`rocjitsu_emulator.py`'s module docstring, and out of the `--emulate` help
text.

.github/scripts/test_rocgdb.py
spatrang: Quite right, thanks, and it is fixed at the root: the variable no longer
exists, so neither the comment nor the condition it guarded is there.
`test_rocgdb.py` simply appends `mirage` to the prerequisites under `--emulate`
and leaves the module to resolve it.

.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,60 @@
...
+
+GPU is a mirage profile (`mi300x`, `mi350x`, `mi450x`) or the gfx target it
+emulates (`gfx942`, `gfx950`, `gfx1250`), and the ROCm tree under test has to
+have been built for it.
+
+| Profile | gfx target |
+|---|---|
+| `mi300x` | `gfx942` |
+| `mi350x` | `gfx950` |
+| `mi450x` | `gfx1250` |
```
lumachad, line 16: Can we point users to some location in rocjitsu listing the available entries? Then we can have just a link and we wouldn't need to bother to keep this list updated.

.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Usage
+
+`--emulate` needs `rocjitsu_emulator.py` beside `test_rocgdb.py`, which a
+ROCgdb checkout has, so name the tree under test for the command:
```
lumachad, line 21: Moving forward we want rocjitsu_emulator.py to be installed alongside test_rocgdb.py such that users won't need to handle this.

.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,60 @@
...
+```bash
+OUTPUT_ARTIFACTS_DIR=~/therock/build/dist/rocm \
+    python .github/scripts/test_rocgdb.py --emulate mi350x --gpu-tests
+
+OUTPUT_ARTIFACTS_DIR=~/therock/build/dist/rocm \
+    python .github/scripts/test_rocgdb.py --emulate mi350x --gpu-tests --parallel
+```
+
+Once a ROCm tree installs the module alongside the harness, a run from inside
+that tree will need nothing configured at all.
```
lumachad, line 32: I think we can now assume this is going to be the case.

.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,60 @@
...
+- `dejagnu` and a C compiler, as on hardware.
+
+## Notes
+
+- `--default-timeout` is raised automatically, since emulated tests are slower.
+- `--parallel` is worth using: each job gets an emulated GPU of its own, rather
+  than queueing for one. Memory scales with the job count, so an unspecified
+  `--jobs` is capped at what memory allows and a `--jobs` that will not fit is
+  flagged.
+- `--one-by-one` gives each test a session, and so a clean GPU, of its own.
```
lumachad, line 49: Phrasing a bit off here. Should be able to reformat it to read better.

.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,60 @@
...
+- `--sanity-check` is ignored (not meaningful for an emulated device).
+
+## Troubleshooting
+
+- **`mirage NOT found on PATH`**: the run is pointed at a tree that is not a
+  ROCm build carrying the emulation tools, or at one built without them, which
+  wants `-D THEROCK_ENABLE_EMULATION=ON`.
+- **`--emulate needs rocjitsu_emulator.py`**: the harness was run from a tree
+  that does not carry the module yet; run it from a ROCgdb checkout, naming the
+  tree under test as above.
```
lumachad, line 59: I think it is going to be a temporary situation where we won't have this module installed alongside test_rocgdb.py, so let's just assume it is going to be there.

.github/scripts/EMULATED-TESTS.md
```diff
@@ -0,0 +1,60 @@
...
+- **Tests time out**: raise `--default-timeout`.
```
lumachad, line 60: ... or genuinely the testcase / emulation / GPU being emulated has issues.

.github/scripts/test_rocgdb.py
```diff
@@ -372,6 +373,32 @@ def _positive_nonzero_int(value: str) -> int:
...
 
 
+def _load_emulator_support() -> types.ModuleType:
+    """
+    Import the rocjitsu emulator support module, needed only for --emulate.
+
+    Imported lazily so this file keeps working as the single script it is
+    installed as (see the install rule in TheRock's debug-tools/rocgdb):
+    hardware runs never reach this, and an emulated run says exactly what is
+    missing instead of failing at import time.
```
lumachad, line 383: Let's not mention things outside the context of the script like TheRock install rules and paths.

.github/scripts/test_rocgdb.py
```diff
@@ -677,6 +717,24 @@ def parse_arguments() -> argparse.Namespace:
             "--one-by-one-log-dir / --one-by-one-test-timeout require --one-by-one."
         )
 
+    # The probe exists to catch a GPU that has wedged and would fail every
+    # remaining test, so that the run aborts with a dmesg.log instead. Neither
+    # half of that applies to an emulated GPU: it is built fresh for each test,
+    # so no state survives to poison the next one, and dmesg says nothing about
+    # a device implemented in userspace. Ignore the option instead of rejecting
+    # it, so the same command line serves hardware and the emulator.
```
lumachad, line 725: The comment is fine, but I think it should be simplified a bit, to make it to the point.

.github/scripts/test_rocgdb.py
```diff
@@ -703,6 +761,36 @@ def parse_arguments() -> argparse.Namespace:
     if args.rocgdb_bin is not None:
         args.rocgdb_bin = validate_path(args.rocgdb_bin, is_file=True)
 
+    # Reject an unknown emulated GPU here rather than after the environment is
+    # set up, and raise the timeout: emulated tests run far slower than on
+    # hardware, where dejagnu's own 10 second per-testcase default is enough.
```
lumachad, line 766: Let's not mention absolute values. Suffices to say dejagnu's default timeout is not enough for tests under emulation.

.github/scripts/test_rocgdb.py
```diff
@@ -2716,6 +2840,10 @@ def _execute_one_by_one(
             logger.warning(
                 f"{STATUS_WARN} {test}: wall-clock timeout after {wall_clock_timeout}s"
             )
+            # The kill denied mirage the chance to tear its session down, so
+            # reclaim it before the next test brings up its own.
```
lumachad, line 2844: Maybe rephrase this comment a bit to make it more clear.

.github/scripts/rocjitsu_emulator.py
```diff
@@ -0,0 +1,345 @@
...
+import tempfile
+from pathlib import Path
+from typing import Dict, List, NamedTuple, Optional, Sequence, Tuple
+
+logger = logging.getLogger(__name__)
+
+# The emulated GPU profiles mirage ships, and the GPU each one emulates. The
+# profile decides what the test programs are compiled for, so it has to match
+# what the ROCm tree under test was built for.
+PROFILE_ARCH = {"mi300x": "gfx942", "mi350x": "gfx950", "mi450x": "gfx1250"}
```
lumachad, line 42: I wonder if we have a better way of keeping this list in sync without much manual intervention.

.github/scripts/rocjitsu_emulator.py
```diff
@@ -0,0 +1,345 @@
...
+ARCH_PROFILE = {arch: profile for profile, arch in PROFILE_ARCH.items()}
+
+# Emulated tests run far slower than on hardware, so dejagnu's 10 second
+# default per-testcase timeout is what makes the slow ones fail rather than any
+# real problem.
```
lumachad, line 47: Just mention dejagnu's default timeout. Don't mention absolute numbers.

.github/scripts/rocjitsu_emulator.py
```diff
@@ -0,0 +1,345 @@
...
+        A run tears its own session down on exit, but SIGKILL (which is how the
+        harness enforces its wall-clock timeouts) leaves no code of mirage's to
+        run at all. `mirage cleanup` is the documented remedy; it leaves live
+        sessions alone, so calling it mid-run is safe.
+
+        Not every mirage has the subcommand: it landed in August 2026, and a
+        dist tarball carries whatever mirage its ROCm tree was pinned to. Say
+        so rather than failing, since the run itself is unaffected -- what is
+        left behind is a stranded emulated GPU, costing memory until the host
+        is rebooted or a newer mirage cleans up after it.
```
lumachad, line 315: We might want to assume this cleanup is already there, since we haven't integrated these changes yet. What do you think? Then we can avoid having to cater to multiple compatibility scenarios.

.github/scripts/EMULATED-TESTS.md
spatrang: There isn't a single rocjitsu page that lists these. `mirage profile list` prints only the names (and it mixes in the non-MI ones too), the friendly mi300x/mi350x/mi450x names live in mirage's build.rs, and the raw configs under emulation/rocjitsu/configs carry the _kmd/multi-gpu variants. So a link can't really stand in for the table here.

What I'll do instead is drop the hand-maintained names and pull them from `mirage --json profile list` at runtime, and keep the small table in the doc just for the profile -> gfx mapping (the CLI doesn't expose that). That keeps the part that actually goes stale out of our hands.

.github/scripts/rocjitsu_emulator.py
spatrang: Yeah, I don't love hardcoding it either. Plan is to source the accepted profile names from `mirage --json profile list` at runtime and keep this dict only for the gfx mapping, since the CLI doesn't hand that back. New profiles then don't need an edit here. I'll wire it up once we're settled on that direction (same thread as the doc one).

spatrang:
Thanks for the pass. Two of these I want to push back on:

**`GDB_LOCK_DIR`**: this is actually working as intended, not a stray positional arg. `runtest` treats `GDB_LOCK_DIR=<dir>` as a dejagnu command-line assignment and sets the Tcl global `$::GDB_LOCK_DIR`, which is exactly what `lock_dir()` in `gdb/testsuite/lib/gdb-utils.exp` reads. Passing it as `--env` would set a process env var that `lock_dir()` never looks at, so the per-job lock dir would silently fall back to the default and jobs would collide. I've added a short comment at that line so this doesn't get re-flagged.

**`THEROCK_ENABLE_EMULATION` in the doc**: I think this is already covered. Requirements says a normal TheRock build and the nightlies both ship the emulation tools, and that a minimal debug-tools-only build needs `-D THEROCK_ENABLE_EMULATION=ON`; the troubleshooting entry repeats that. Happy to reword if you think it's still unclear.

The `PROFILE_ARCH` / `mirage --json profile list` suggestion is a good one and I'm doing that in the two inline threads.

.github/scripts/rocjitsu_emulator.py
lumachad: > Yeah, I don't love hardcoding it either. Plan is to source the accepted profile names from `mirage --json profile list` at runtime and keep this dict only for the gfx mapping, since the CLI doesn't hand that back. New profiles then don't need an edit here. I'll wire it up once we're settled on that direction (same thread as the doc one).

If there isn't a good way to do it, let's not push ourselves. We should request that feature from rocjitsu so it makes the lives of tools easier.

.github/scripts/EMULATED-TESTS.md
spatrang: Done, dropped the "must sit beside test_rocgdb.py" caveat and just assume it ships with the harness.

.github/scripts/EMULATED-TESTS.md
spatrang: Done, took the caveat out and assume it's the case now.

.github/scripts/EMULATED-TESTS.md
spatrang: Reworded that bullet so it reads better.

.github/scripts/EMULATED-TESTS.md
spatrang: Done, removed the troubleshooting entry and assume the module is there.

.github/scripts/EMULATED-TESTS.md
spatrang: Good point, broadened it to cover a genuine test/emulation/GPU issue, not just the harness.

.github/scripts/test_rocgdb.py
spatrang: Done, dropped the TheRock install-rule mention and kept it to the script.

.github/scripts/test_rocgdb.py
spatrang: Trimmed it down to the point.

.github/scripts/test_rocgdb.py
spatrang: Done, dropped the absolute number and just say dejagnu's default isn't enough under emulation.

.github/scripts/test_rocgdb.py
spatrang: Rephrased it to read more clearly.

.github/scripts/rocjitsu_emulator.py
spatrang: Done, removed the absolute value and just point at dejagnu's default.

.github/scripts/rocjitsu_emulator.py
spatrang: Agreed, assumed the cleanup subcommand is there and dropped the compat handling.

.github/scripts/rocjitsu_emulator.py
spatrang: Sounds good, let's not force it then. The CLI only gives names, not the gfx mapping, so there's no clean way to drop this dict right now. I'll keep it as-is and raise a request with rocjitsu to expose the profile to arch mapping in `mirage --json profile list` so we can retire it later.

lumachad:

## Review notes

**Cosmetic / style**

- `rocjitsu_emulator.py:218-220` — `Emulator.wrap()` uses a bare `Sequence` without a type argument while the rest of the file uses typed `Dict` and `List`. Minor inconsistency.

**Typos / comments**

- `EMULATED-TESTS.md:47` — "first raise \`--default-timeout\`" reads slightly awkwardly; "try raising \`--default-timeout\` first" flows better.

- PR body description says `find_mirage()` locates mirage from \`\$MIRAGE_BIN\`, a cargo build, or a TheRock checkout path, but the actual implementation only does \`shutil.which("mirage", path=path)\`. The description is stale relative to the code. The module docstring in \`rocjitsu_emulator.py\` is correct; the issue is in the PR body only.

**Bugs**

- `rocjitsu_emulator.py:81-84` (`parallel_jobs`) — `fits = max(1, int((available_gb - BASE_MEMORY_GB) / JOB_MEMORY_GB))` clamps to 1 even when available memory is below `BASE_MEMORY_GB`. For example, with `available_gb = 7.5` and `BASE_MEMORY_GB = 8.0`: `int((7.5 - 8.0) / 1.0) = 0`, `max(1, 0) = 1`, so `fits = 1`. In the `requested is not None` branch, if the user passes `--jobs 1`, the check `requested > fits` evaluates to `1 > 1 = False`, so no OOM warning fires even though the machine is below the threshold for even one job. The `low_memory` guard that catches this only runs in the default (no `requested`) branch. Suggested fix: use `max(0, int(...))` and treat `fits == 0` as "even 1 job will not fit", emitting the warning regardless of `requested`.

- `test_rocgdb.py` (`parse_arguments`, `--emulate` block) — when `--fraction` is used with `--emulate --parallel`, the memory advice path is skipped entirely (`if args.fraction is None` guard). The job count is resolved later in `run_tests` as `max(1, int(os.cpu_count() * fraction))`, bypassing `parallel_jobs()`, so no OOM warning fires even if the resolved count exceeds what memory can hold. For example, on a 32-core machine with 12 GB available, `--fraction 0.5` resolves to 16 jobs, needing roughly 24 GB, with no warning. The gap could be closed by resolving `args.fraction` to a job count early (in the `args.emulate is not None` block) and passing it through `parallel_jobs()` as the `requested` argument, then storing the result in `args.jobs` and clearing `args.fraction`. That way the memory check applies and the rest of the code, which already handles `args.jobs`, needs no changes.

**Nits / optional**

- `test_rocgdb.py:711-724` — when `--emulate` silently clears `args.sanity_check_timeout`, a debug-level log line (e.g. "also clearing sanity_check_timeout") would make future diagnostics easier when someone wonders why their `--sanity-check-timeout` had no effect.

- `EMULATED-TESTS.md` — the table uses `|---|---|` with no padding; `| --- | --- |` is the more common convention in this project's docs (minor).

spatrang:
Thanks, good catches. All addressed in the latest push:

**Bugs**
- `parallel_jobs` low-memory clamp: switched `fits` to `max(0, ...)` and added a `below_one_job` flag so the OOM warning now fires even on the explicit `--jobs 1` path, not just the auto one. The job count itself still floors at 1 so a run never drops to zero jobs.
- `--fraction` under `--emulate --parallel`: it's now resolved to a concrete count up front and fed through `parallel_jobs()` as the request, stored in `args.jobs` with `args.fraction` cleared, so the same memory check and OOM warning cover it. One thing to note: like an explicit `--jobs`, a `--fraction` request is honoured past the wall-clock ceiling (it was never ceiling-capped before either), just memory-warned now. Called that out in the commit log.

**Style / typos**
- Typed `wrap()`'s param as `Sequence[str]`.
- `EMULATED-TESTS.md`: table separator is now `| --- | --- |`, and reworded the troubleshooting bullet to "try raising `--default-timeout` first".
- Fixed the stale `find_mirage()` description in the PR body: it now says mirage is found on the run's PATH via `shutil.which`, matching the code, and drops the old `$MIRAGE_BIN`/cargo/TheRock chain.

**Nit**
- Added a `logger.debug` when `--emulate` clears `--sanity-check-timeout`, so it's traceable when someone wonders why their timeout had no effect.

lumachad:

## Review summary
Well-scoped, thoroughly documented change that adds `--emulate GPU` support via a new, cleanly separated `rocjitsu_emulator.py` module. The hardware code path is genuinely untouched (guarded behind `emulator is not None` / `args.emulate is not None`), argument validation is consistent with existing `--fraction`/`--jobs` semantics, and the shell-script generation in `runtest_wrapper()` is careful about quoting. One real gap: orphaned emulated-GPU sessions aren't reclaimed outside the one-by-one path. A few small nits below.

## Bugs
- `.github/scripts/test_rocgdb.py:2070-2082` (in `run_tests`, non-one-by-one branch) — no `emulator.cleanup()` is called if the `make check` invocation is interrupted. `_run_command()` already SIGKILLs the whole process group on `KeyboardInterrupt`/any escaping exception (`test_rocgdb.py:160-176`), which kills `mirage run` itself without giving it a chance to tear down its session, exactly the scenario `Emulator.cleanup()` exists for (`rocjitsu_emulator.py:402-429`), but that call only happens in `_execute_one_by_one`'s wall-clock-timeout path (`test_rocgdb.py:2845-2848`). A Ctrl-C during a `--parallel` or single-session emulated run (or any per-job `runtest-in-session` script killed the same way) will strand an emulated GPU consuming memory until someone runs `mirage cleanup` manually or the next `--emulate` invocation happens to call it. Consider calling `emulator.cleanup()` from a `finally`/`atexit` around the `make check` invocation in `run_tests()` too, not just in `_execute_one_by_one`.

## Nits / optional
- `rocjitsu_emulator.py:315-318` — `Emulator.resolve()`'s return type is annotated as the string literal `"Emulator"`, but the file already has `from __future__ import annotations` (line 90), which makes the quoting unnecessary (can just be `Emulator`). Purely cosmetic.
- `.github/scripts/test_rocgdb.py:623-624` / `rocjitsu_emulator.py:250-266` — `mirage` is looked up twice: once by `check_executables()` (which only checks presence) and again by `find_mirage()` inside `Emulator.resolve()` (which returns the resolved path). Harmless (same `PATH`, same result) but slightly redundant; could pass the already-resolved path through instead of re-searching.
- `rocjitsu_emulator.py:228-234` — `parallel_jobs()`'s auto-selection: `reason` is computed once from `jobs == JOB_CEILING`, then unconditionally overwritten if the memory cap kicks in (`fits < jobs`), so the "ceiling" phrasing is effectively dead in that branch. Not a bug (the memory-derived message is correct), just an unused code path worth a comment or removal for clarity.

## Tests
The PR body documents extensive manual verification (13 invocation shapes diffed against upstream, full suite runs, profile/arch validation, missing-tree/missing-mirage error paths), which is solid, but none of that is captured as an automated unit test for the new pure logic in `rocjitsu_emulator.py` (e.g., `normalize_profile`, `parallel_jobs`, `check_profile_arch`), these are self-contained and easy to unit-test (no GPU/mirage needed) and would guard against regressions in the memory-ceiling arithmetic or profile-alias resolution without needing an emulator-capable machine to run CI. Worth considering, though not a blocker given the manual test matrix already covers the behavior.

No security, GNU-standards (N/A, Python), or typo/comment issues found. The docstrings and `EMULATED-TESTS.md` are clear and accurate against the code. `black --check` also comes back clean on the new/changed files.

lumachad:
Other than those, this is looking good to me. Thanks!

Let me know once you do another push and we can go from there.

spatrang:
Pushed another round addressing the latest review:

- `run_tests()` now registers `emulator.cleanup()` via `atexit`, so a `--parallel` or single-session run reclaims its emulated GPU if the harness is interrupted, not just on the one-by-one timeout path.
- Dropped `mirage` from `check_executables()` since `Emulator.resolve()` already looks it up (and gives a better error), so it isn't searched twice.
- Removed the now-needless string quoting on `Emulator.resolve()`'s return type, and clarified the `parallel_jobs()` auto branch where the memory cap overrides the CPU/ceiling reason.
- Added `test_rocjitsu_emulator.py`: 22 unit tests for the pure logic (`normalize_profile`, `parallel_jobs`, `check_profile_arch`), no GPU or mirage needed.

Also ran it against a real rocjitsu tree: profile/arch validation, single-session and `--parallel` runs, the `--fraction` job-count path, and the low-memory OOM warning all behave as expected. black and flake8 are clean.

lumachad:

## Bugs
- `.github/scripts/EMULATED-TESTS.md:49` — the troubleshooting entry quotes `mirage NOT found on PATH`, a string that can never appear for mirage since it's deliberately excluded from `check_executables()`'s list (test_rocgdb.py:2389-2394). The real error is `EmulatorError`'s "no mirage found: ..." text from `find_mirage()` (rocjitsu_emulator.py:198-203). Worth fixing so users searching the docs for the actual message find it.

## Nits / optional
- `rocjitsu_emulator.py:266` — `Emulator.wrap()`'s `env` parameter and its `--env` construction (lines 277-278) have no current caller; could be dropped until something needs it, or left if it's anticipated for a near-term follow-up.

lumachad (approved):
Mostly we just need a quick update on the EMULATED-TESTS.md. Otherwise this looks good to me. Thanks!

spatrang:
Small follow-up for the last review:

- Fixed the troubleshooting entry in `EMULATED-TESTS.md`: it quoted `mirage NOT found on PATH`, which can't occur anymore now that mirage is resolved by `find_mirage()`. It now uses the actual message, `no mirage found`.
- Dropped `wrap()`'s unused `env`/`--env` argument (GDB_LOCK_DIR goes through the generated runtest script, not `--env`), and removed the now-unused `Dict` import.

Unit tests, black, and flake8 still clean.

PR #317, opened 2026-09-01
[testsuite] gdb.rocm/omp-target-multi-kernel: fix optimized-out x

lumachad:
At -O3 the compiler optimizes away the `x` parameter in `square_dev`
and `cube_dev`, causing `print x` to report `<optimized out>` and the
test to fail in the `-O3` and `-O3 -flto` CI jobs.

Mark both functions with `__attribute__((optnone))` to keep `x` live
and readable in the debugger.

Fixes: https://github.com/ROCm/ROCgdb/actions/runs/33192378358/job/98929943116


amd-bfilipov (approved):
looks good

PR #318, opened 2026-09-01
gdb.rocm: xfail coop-group-grid-sync on gfx1151

spatrang:
## Summary
- AIROCGDB-654: on gfx1151, `hipLaunchCooperativeKernel` fails under the debugger with HIP 401 because KFD refuses GWS allocation while debug trap is enabled. The same binary runs without GDB.
- XFAIL the continue in `coop-group-grid-sync.exp` on gfx1151 instead of FAIL or UNSUPPORTED. Drop the xfail when this starts XPASS (kernel/DKMS with the GWS-debug fix).

## Test plan
- [x] Validated on Strix Halo (gfx1151): continue is XFAIL, no unexpected failures (GCC and LLVM).

### DIFF
--- a/gdb/testsuite/gdb.rocm/coop-group-grid-sync.exp
+++ b/gdb/testsuite/gdb.rocm/coop-group-grid-sync.exp
@@ -40,6 +40,31 @@ if {[build_executable "failed to prepare" $testfile $srcfile {debug hip}]} {
     return
 }
 
+# On gfx1151, hipLaunchCooperativeKernel fails under the debugger
+# with HIP error 401.  KFD refuses GWS allocation while debug trap
+# is enabled.  The program works without GDB so this is an expected
+# failure rather than UNSUPPORTED.  Drop the xfail once a kernel
+# with the GWS debug fix is in use.
+
+proc target_has_xfail {} {
+    set xfail_arches {gfx1151}
+
+    set targets [find_amdgpu_devices]
+    if {[llength $targets] == 0} {
+	return 0
+    }
+
+    # The test runs on the first visible HIP device.
+    set target [lindex $targets 0]
+    return [expr {[lsearch -exact $xfail_arches $target] != -1}]
+}
+
+proc maybe_xfail {} {
+    if {[target_has_xfail]} {
+	setup_xfail "*-*-*" "AIROCGDB-654"
+    }
+}
+
 # Test debugging a single-device cooperative kernel.  Stop at a
 # breakpoint after grid.sync () -- which fires once all waves have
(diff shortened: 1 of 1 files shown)


gdb/testsuite/gdb.rocm/coop-group-grid-sync.exp
```diff
@@ -40,6 +40,32 @@ if {[build_executable "failed to prepare" $testfile $srcfile {debug hip}]} {
...
+    set xfail_arches {gfx1151}
+
+    set targets [find_amdgpu_devices]
+    if {[llength $targets] == 0} {
+	return 0
+    }
+
+    # The test will run on GPU-0, so it should be the first of the list.
+    set target [lindex $targets 0]
+    return [expr {[lsearch -exact $xfail_arches $target] != -1}]
```
lumachad: I'm slightly worried about this one. When we go fully parallel, we might not want to run things on specific GPU indexes. But the test is already this way, right?

spatrang: Right. This matches watchpoint-basic.exp and watchpoint-at-end-of-shader.exp. find_amdgpu_devices returns HIP visible devices, so index 0 is the first device this process sees. When ROCR_VISIBLE_DEVICES pins one GPU per job, that is the job GPU, not a hard coded machine index. I can reword the comment to say first visible HIP device if that is clearer.

gdb/testsuite/gdb.rocm/coop-group-grid-sync.exp
```diff
@@ -40,6 +40,32 @@ if {[build_executable "failed to prepare" $testfile $srcfile {debug hip}]} {
     return
 }
 
+# On gfx1151, hipLaunchCooperativeKernel currently fails under the
+# debugger with HIP 401 (KFD refuses GWS while debug trap is enabled;
+# AIROCGDB-654 / ROCM-9951).  The same binary runs correctly without
+# GDB, so this is an expected failure rather than UNSUPPORTED.  When a
+# kernel with the KFD fix is in use, the continue below should XPASS
+# and this xfail can be dropped.
```
lumachad: I requested an adjustment of the comment on a side chat.

lumachad (approved):
With the comment fix, this looks OK to me. Thanks.

gdb/testsuite/gdb.rocm/coop-group-grid-sync.exp
spatrang: Reworded the comments.

PR #324, opened 2026-09-03
Add CHANGELOG entry for __builtin_verbose_trap support

amd-shahab:


lumachad (approved):
Leaving my approval here. Looks good to me as-is, but could get more clarifications.

amd-bfilipov (approved):
Simple enough, describes the change well. Looks good

CHANGELOG_AMD.md
```diff
@@ -7,11 +7,15 @@ Full documentation for ROCgdb is available at
 
 ### Added
 
+- For AMD64 and AMDGPU targets, print backtrace information for
+  Clang's __builtin_verbose_trap.  The compiler appends a message in
+  debug information that comes from the builtin and ROCgdb prints it
+  as the context of frame #0: 
+  #0 ... __clang_trap_msg$<category>$<message> () at ...
```
palves, line 15: It's both category and message, and those come from the built-in call.   I'd tweak it like this:

The debug information emited by the compiler describes the category and message arguments passed to the built-in as a call to an inlined function, and ROCgdb presents it as the context of frame #0: ...

palves (approved):

PR #327, opened 2026-09-03
Add "info waves", "info waves count", -wave-info, -wave-count

lumachad:
Add CLI commands "info waves" and "info waves count" and their MI
equivalents -wave-info and -wave-count for querying GPU wave state.

"info waves [INF]" displays a summary of active waves broken down by
state (running, single-stepping, stopped) for the given inferior, or
the current inferior if omitted.  "info waves count [INF]" displays just
the total wave count.  The optional inferior number argument allows
cross-inferior queries in non-stop sessions without switching inferiors.

The MI commands always query the current inferior.

Includes documentation and tests across four scenarios: before GPU
activity, all-stop, non-stop with single-stepping, and multi-inferior
non-stop.

## Test plan

- [ ] info-waves-cli.exp passes (all-stop, non-stop, multi-inferior)
- [ ] info-waves-mi.exp passes (all-stop, non-stop, multi-inferior)
- [ ] Existing tests that use info_threads_get_wave_count still pass

### DIFF
--- a/gdb/amd-dbgapi-target.c
+++ b/gdb/amd-dbgapi-target.c
@@ -4058,6 +4058,7 @@ info_agents_command (const char *args, int from_tty)
 }
 
 static struct cmd_list_element *queue_list;
+static struct cmd_list_element *wave_list;
 
 static void
 info_queues_command (const char *args, int from_tty)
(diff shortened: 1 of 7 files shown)


lumachad:
Re-purposed the PR and went the direction of having a single maint command that counts waves, as suggested at some point in the PR. Still a draft for now as we want #225 to land first so we can stash the maint command under amdgpu.

PR #332, opened 2026-09-10
gdb.rocm: select precise-memory fault insn by XNACK mode

spatrang:
## Summary

`gdb.rocm/precise-memory.exp` failed on MI325X while passing on MI300X, even though both are gfx942. The divergence is driven by XNACK mode, not architecture: with XNACK enabled the faulting `s_load_dword` is replayable, so the wave halts on that instruction; with XNACK disabled the PC advances to the following `s_nop`.

## Change

- Gate the expected instruction on the effective XNACK mode instead of hard-coding `s_nop`.
- Add a `hip_device_xnack_on` helper to `lib/rocm.exp` that reads `hipDeviceProp_t::gcnArchName`, whose feature suffix (e.g. `gfx942:sramecc+:xnack+`) reflects the runtime mode set by `HSA_XNACK` and device config.

Distinct per-mode expectations are kept (rather than accepting both) to preserve the test's ability to catch a fault-PC regression.

## Test

- MI300X (xnack- default): PASS
- MI300X (`HSA_XNACK=1`): PASS
- MI325X (xnack+ default): PASS


gdb/testsuite/gdb.rocm/precise-memory.exp
```diff
@@ -83,12 +83,23 @@ proc do_test { } {
 	    gdb_test_no_output "set amdgpu precise-memory on"
 	}
 
-	# If precise-memory is available, run until a SIGSEGV is reported.  At
-	# that point, the PC should point to the s_nop instruction (the one
-	# following the one which caused the memory violation).
+	# If precise-memory is available, run until a SIGSEGV is reported.
+	# The PC then points at or just after the faulting instruction,
+	# depending on the XNACK mode (see below).
```
lancesix, line 88: I am pretty sure the documentation for GDB states that it should be the next instruction

from gdb.texinfo:

```
@kindex set amdgpu precise-memory                                       
@cindex AMD GPU precise memory event reporting                       
@item set amdgpu precise-memory @var{mode}                   
                                                                   
Controls how @acronym{AMD} @acronym{GPU} devices detect memory violations and
address watch events, where @var{mode} can be:
                                                                    
@table @code               
                                 
@item off                                       
The program location may not be immediately after the instruction that
caused the memory violation or address watch event.  This is the
default.         
                  
@item on                     
Requests that the program location will be immediately after the
instruction that caused a memory violation or address watch event.
Enabling this mode may make the @acronym{AMD} @acronym{GPU} device execution
significantly slower as it has to wait for each memory operation to complete
before executing the next instruction.                               
                                                                     
@end table                                                          
```

So either we need to update the documentation, or change gdb or dbgapi? to adjust the PC based on the xnack mode.

lancesix (commented):
If the device supports it, should we try to test both the HSA_XNACK modes?

gdb/testsuite/gdb.rocm/precise-memory.exp
lumachad: I guess it depends on where is the source of truth here. Do we want to shield gdb from these behavior differences?

spatrang: Good point on the manual. I went digging in dbgapi and it doesn't actually guarantee "immediately after" anywhere. The header only says the PC "may not be positioned at the instruction that caused the memory violation", so stopping on the faulting load with xnack+ is fine as far as dbgapi is concerned. Doing the fixup there wouldn't really help us either, we'd still be stuck handling the core-file case ourselves.

Talked it over with Luis and we'll keep the test change here and just fix the manual. I've reworded the precise-memory "on" entry to say the PC lands on the faulting instruction with xnack+ and right after it with xnack-.

spatrang:
Updated (single amended commit): folded a `gdb/doc/gdb.texinfo` change into the patch.

- `set amdgpu precise-memory on` now documents that the reported PC is left **at** the faulting instruction under xnack+ (replayable) or **immediately after** it under xnack-, instead of unconditionally promising the following instruction.
- The test expectation remains gated on the effective XNACK mode via the new `hip_device_xnack_on` helper.

No dbgapi change: its contract already permits this (the header only states the PC "may not be positioned at the instruction that caused the memory violation"), and a dbgapi-side fixup would still require core-file handling in rocgdb.

lumachad:
Looks good to me as-is, but Lancelot suggested we check if both modes are supported and proceed to try them both to make sure they do the expected thing.

lumachad:

`gdb/testsuite/gdb.rocm/precise-memory.cpp:30-31`: The inline comment is now inaccurate after this patch. It says "With precise-memory, the exception should be reported on the s_nop", but that is only true under XNACK disabled; under XNACK enabled it stops on `s_load_dword`. Suggested replacement:

```
/* Simple kernel which loads from address 0 to trigger a pagefault.
   With precise-memory disabled the fault is reported after the
   s_nop.  With precise-memory enabled, the location depends on the
   XNACK mode: on the s_load_dword itself (XNACK enabled) or on the
   following s_nop (XNACK disabled).  */
```

`gdb/doc/gdb.texinfo` (`@item on` section): "immediately after it" is ambiguous -- "it" could refer to the instruction or to the event. Suggest replacing:

> the program location is left either at the instruction that caused the event or immediately after it

with:

> the program location is left either at the faulting instruction or immediately after it

`gdb/testsuite/gdb.rocm/precise-memory.exp:97` (nit): The pattern `"s_load_dword\[^\r\n\]*"` also matches `s_load_dwordx2`, `s_load_dword_d16`, etc. Since the kernel hard-codes `s_load_dword s12, [s10, s11]`, a tighter pattern such as `"s_load_dword\[ \t\]"` would catch a future ISA change more visibly. Not blocking.

spatrang:
Updated the PR based on the review feedback.

- The test now exercises both XNACK modes instead of only the device default. A new hip_device_xnack_modes helper in lib/rocm.exp reports which modes the device supports, and the test forces HSA_XNACK for each one and checks the expected instruction (s_nop for off, s_load_dword for on).
- Fixed the stale kernel comment and reworded the precise-memory doc entry to match the XNACK-dependent location.
- Tightened the s_load_dword pattern so it no longer matches wider variants like s_load_dwordx2, while still consuming the operands to end of line so the match anchors correctly.

Validated on both machines, both modes pass:
- MI300X (default xnack-): off and on PASS
- MI325X (default xnack+): off and on PASS

Both boxes can be forced to either mode, so both iterations run and pass on each.

lumachad:

**`lib/rocm.exp:hip_device_xnack_modes` -- bool vs list**

The helper returns a list of supported modes (`{off}`, `{off on}`) to handle the theoretical case where `HSA_XNACK=0` is ignored by the runtime and a device is stuck in XNACK-on. If that case cannot happen in practice, a simpler bool (`hip_device_xnack_supported`) would express the intent more directly, and the XNACK-off iteration could run unconditionally:

```tcl
if {[hip_device_xnack_supported]} {
    with_test_prefix "xnack on"  { do_test 1 {s_load_dword[ \t][^\r\n]*} }
}
with_test_prefix "xnack off" { do_test 0 {s_nop[ \t]+0} }
```

Worth confirming with the driver/hardware team whether a device can ever be locked to XNACK-on such that `HSA_XNACK=0` has no effect. If not, the list is unnecessary complexity.

**`precise-memory.exp:107` -- `foreach` flat-list iteration (nit)**

The triple-variable `foreach` over a flat list is valid Tcl but uncommon enough that a brief comment would help readers who expect single-variable iteration:

```tcl
# Iterate {mode request expected_insn} triples.
foreach { mode request expected_insn } { ... }
```

spatrang:
Simplified per the review. Replaced the list-returning helper with a bool hip_device_xnack_supported that only probes whether XNACK can be turned on. The test now runs the XNACK off path unconditionally (every device can run with it off) and adds the XNACK on path only when supported. Also updated the commit message to match.

Re-validated on both machines, both paths pass:
- MI300X (default xnack-): off and on PASS
- MI325X (default xnack+): off and on PASS

lumachad (approved):
OK if CI is clean.

PR #336, opened 2026-09-11
gdb: Add generic AMDGPU target testing support

lumachad:
## Summary

- Add \`gdb/testsuite/boards/rocm-generic.exp\`, a new DejaGnu board that runs the \`gdb.rocm/\` testsuite with kernels compiled for generic AMDGPU targets rather than the concrete arch detected on the system (e.g. gfx942 is compiled as \`gfx9-4-generic\`).
- Add \`generic_target_for\` proc to \`gdb/testsuite/lib/rocm.exp\` that maps a concrete gfx arch to its generic equivalent, making the mapping reusable by individual tests.
- Add a \`::rocm_use_generic_targets\` flag consumed by \`hcc_amdgpu_targets\` so the board can activate generic-target substitution without replacing the proc, keeping the \`hcc-amdgpu-targets.exp\` unit test unaffected.
- Fix \`amdgpu_address_scope\` to use the wave's own architecture id when a concrete wave is available, rather than the gdbarch mach of the generic code object. This is required for \`amd_dbgapi_address_dependency\` to succeed when debugging generic code objects on concrete hardware.

Addresses AIROCGDB-660.

## Test plan

- [ ] Run \`make check RUNTESTFLAGS="--target_board=rocm-generic"\` on a system with a GPU that has a known generic target (e.g. gfx942, gfx1100, gfx1030)
- [ ] Verify tests compile against the generic arch and run correctly on the concrete hardware
- [ ] Verify tests are skipped cleanly on hardware with no generic target mapping
- [ ] Verify \`gdb.rocm/hcc-amdgpu-targets.exp\` passes under both the default and \`rocm-generic\` boards

### DIFF
--- a/gdb/amdgpu-tdep.c
+++ b/gdb/amdgpu-tdep.c
@@ -1762,10 +1762,26 @@ amdgpu_address_scope (struct gdbarch *gdbarch, ptid_t ptid, CORE_ADDR address)
     = amdgpu_segment_address_from_core_address (address);
 
   amd_dbgapi_architecture_id_t architecture_id;
-  if (amd_dbgapi_get_architecture
-      (gdbarch_bfd_arch_info (gdbarch)->mach, &architecture_id)
-      != AMD_DBGAPI_STATUS_SUCCESS)
-    error (_("amd_dbgapi_get_architecture failed"));
+  if (ptid_is_gpu (ptid))
+    {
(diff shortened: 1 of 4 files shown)


lancesix (approved):
Tested on gfx1031 (using gfx10-3-generic), and everything works fine.

Thanks.

PR #339, opened 2026-09-15
ci: add optional generic-board GPU testing job

lumachad:
Add a `test-gpu-generic` job that runs the `gdb.rocm/` testsuite using
the `rocm-generic` DejaGnu board (`--target_board=rocm-generic`).  The
job fans out across the same `amdgpu_family` matrix as `test-gpu` so
that generic-compiled code is verified on every concrete GPU family
already under test.

The job is opt-in via the `test:generic-gfx-board` PR label.  A new
`run_generic_board` output on the `setup` job propagates the flag from
the label check in the `families` step.  The `therock_ci_summary` job
lists `test-gpu-generic` in its needs; the existing `result!="skipped"`
filter means the summary still passes when the label is absent.

Combined with `ci:run-all-archs`, adding `test:generic-gfx-board` fans
out generic-board testing across all six architecture families in
parallel.


.github/workflows/therock-ci.yml
```diff
@@ -49,6 +50,11 @@ jobs:
           else
             echo 'test_families=["gfx94X"]' >> $GITHUB_OUTPUT
           fi
+          if [[ "${{ contains(github.event.pull_request.labels.*.name, 'test:generic-gfx-board') }}" == "true" ]]; then
```
spatrang, line 53: Nit: the other gating labels in this file use the `ci:` prefix (`ci:skip`, `ci:run-all-archs`). For consistency, consider `ci:generic-gfx-board` rather than `test:generic-gfx-board` (and the matching job `if` / comment). Not blocking. If `test:` is deliberate to keep a separate label group, feel free to ignore.

.github/workflows/therock-ci.yml
```diff
@@ -232,6 +238,31 @@ jobs:
...
+  # Generic-board jobs. Run the gdb.rocm/ testsuite with kernels compiled for
+  # generic AMDGPU targets (rocm-generic board). Runs on every concrete GPU
+  # family that test-gpu covers, so we verify generic-compiled code works on
+  # real hardware. Opt-in via the 'test:generic-gfx-board' PR label.
+  test-gpu-generic:
+    name: Test GPU (generic board)
+    needs: [setup, therock-build-linux]
+    if: >-
+      ${{ needs.setup.outputs.enable_therock_ci == 'true' &&
+          needs.setup.outputs.run_generic_board == 'true' }}
```
spatrang, line 250: This is the only job in this file whose `if` is a folded multi-line `${{ }}` expression. It is valid (GitHub Actions ignores the newline inside `${{ }}`). If you want to stay uniform with the other jobs, you can collapse it:

```yaml
    if: ${{ needs.setup.outputs.enable_therock_ci == 'true' && needs.setup.outputs.run_generic_board == 'true' }}
```

Optional.

.github/workflows/therock-ci.yml
```diff
@@ -232,6 +238,31 @@ jobs:
...
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        amdgpu_family: ${{ fromJSON(needs.setup.outputs.test_families) }}
```
spatrang, line 256: Sharing `test_families` with `test-gpu` is the right default for families that have a generic mapping (`gfx942`/`gfx950` → `gfx9-4-generic`, RDNA3/4 → `gfx11-generic` / `gfx12-generic`).

`generic_target_for` in `gdb/testsuite/lib/rocm.exp` has **no** entry for `gfx90a`. With both `test:generic-gfx-board` and `ci:run-all-archs`, this matrix still schedules a generic-board job on a `gfx90a` runner. `hcc_amdgpu_targets` will drop that arch, so the suite either skips or spends a GPU runner without testing generic code.

Consider a slightly narrower family list for this job (omit `gfx90a`) or a short comment here that the empty mapping is intentional and the board self-skips. Not a merge blocker for the default `gfx94X` path, which already passed.

.github/workflows/therock-ci.yml
```diff
@@ -232,6 +238,31 @@ jobs:
...
+    uses: ./.github/workflows/therock-test-packages.yml
+    with:
+      amdgpu_families: ${{ matrix.amdgpu_family }}
+      artifact_group: ${{ needs.setup.outputs.artifact_group }}
+      therock_ref: ${{ needs.therock-build-linux.outputs.therock_ref }}
+      artifact_run_id: ${{ github.run_id }}
+      projects_to_test: rocgdb-gpu
+      test_options: "--runtestflags=--target_board=rocm-generic"
```
spatrang, line 264: This matches the hip-board pattern (`--runtestflags=--target_board=...`) and is appended onto the `rocgdb-gpu` `test_script` by `therock-test-packages.yml`. With no `--tests`, the default `gdb.rocm` set runs, which matches the job comment.

Tiny style nit: hip-board uses a folded scalar (`>-`) for the same flag. Either form is fine.

.github/workflows/therock-ci.yml
```diff
@@ -246,6 +277,7 @@ jobs:
       - test-gpu-check-read1
       - test-gpu-check-readmore
       - test-gpu-hip-board
+      - test-gpu-generic
```
spatrang, line 280: This is correct. When the label is absent the job is `skipped`, and the existing `jq` filter drops `skipped` results, so `therock_ci_summary` still passes. When the label is present and the job fails, it is included and fails the summary.

spatrang (commented):
## Review summary

Clean, low-risk CI addition. `test-gpu-generic` mirrors `test-gpu` (matrix, inputs, permissions, `fail-fast: false`). The `rocm-generic` board exists, `contains()` matches the `ci:run-all-archs` pattern, `test_options` is appended by `therock-test-packages.yml`, and `therock_ci_summary` still passes when the job is skipped (`result != "skipped"`). `therock_ref` comes through `therock-build-linux`, so `resolve` does not need to be an extra `needs` entry.

Default-family CI is green (`gfx94X` generic-board job passed in the latest run). The earlier `ci:run-all-archs` run was cancelled; several non-generic GPU jobs failed there too, so those failures do not look unique to this job.

No correctness or security blockers. Comments below are nits plus one matrix note about `gfx90a`.

### Commit / PR prose

Please drop the semicolon in the commit body and split it into two sentences, and remove the `Generated with Claude Code` trailer from the PR description.

Current:
> The therock_ci_summary job lists test-gpu-generic in its needs; the existing result!="skipped" filter means the summary still passes when the label is absent.

Suggested:
> The therock_ci_summary job lists test-gpu-generic in its needs. The existing result!="skipped" filter means the summary still passes when the label is absent.

### Optional (cannot pin to a changed line)

The "Extended test configurations" overview comment still says only the two optimization jobs follow `test_families` and that check-type / hip-board stay on `gfx94X`. `test-gpu-generic` is now a third matrix-fanning job. A one-line mention there would keep that overview accurate.

.github/workflows/therock-ci.yml
lumachad: Done.

.github/workflows/therock-ci.yml
lumachad: Done.

.github/workflows/therock-ci.yml
lumachad: Ack.

.github/workflows/therock-ci.yml
lumachad: Done.

.github/workflows/therock-ci.yml
lumachad: Done.

lumachad:
@spatrang Thanks for the review. Addressed the comments now.

1. Renamed label test:generic-gfx-board → ci:generic-gfx-board for consistency.
2. Collapsed the test-gpu-generic job's folded multi-line if: into a single line.
3. Added a generic_test_families output (excludes gfx90a, which has no generic_target_for mapping) and pointed the generic-board job's matrix at it instead of test_families.
4. Switched test_options for the generic-board job to the folded scalar (>-) style used by hip-board.

PR #341, opened 2026-09-15
ci: add --gpu-corefile-tests switch to test_rocgdb.py

lumachad:
## Summary

- Adds a new `--gpu-corefile-tests` flag to the test selection group in `.github/scripts/test_rocgdb.py`
- The flag globs `gdb.rocm/runtime-core*.exp` against the testsuite directory at runtime, so new tests matching that pattern are picked up automatically without further changes to the script
- Mutually exclusive with `--tests`, `--gpu-tests`, and `--cpu-tests`

## Test plan

- [ ] Verify `--gpu-corefile-tests` runs only `gdb.rocm/runtime-core.exp` on a current tree
- [ ] Verify combining `--gpu-corefile-tests` with `--tests`, `--gpu-tests`, or `--cpu-tests` is rejected by argparse
- [ ] Verify a clear error is emitted if no `runtime-core*.exp` files are found


.github/scripts/test_rocgdb.py
```diff
@@ -519,14 +520,21 @@ def parse_arguments() -> argparse.Namespace:
         "--gpu-tests",
         action="store_true",
         help="Run GPU-specific tests (gdb.rocm directory only). "
-        "Mutually exclusive with --tests and --cpu-tests.",
+        "Mutually exclusive with --tests, --cpu-tests, and --gpu-corefile-tests.",
```
spatrang, line 523: Severity: Important (non-blocking)

looks like the `--tests` help string up on line 516 didn't get the same update. It still says "Mutually exclusive with --gpu-tests and --cpu-tests." Mind adding `--gpu-corefile-tests` to that one too so it matches the `--gpu-tests` and `--cpu-tests` strings? Just a help-text fix, argparse already enforces the exclusion.

lumachad: Fixed in d5009e5d630. Added `--gpu-corefile-tests` to the `--tests` help string.

spatrang: Fixed in the latest push, thanks. The `--tests` help string now lists `--gpu-corefile-tests` too, so all four are consistent.

spatrang (approved):
Thanks for the update. The help-text nit is addressed: `--tests` now lists `--gpu-corefile-tests` alongside `--gpu-tests` and `--cpu-tests`, so the four selection flags read consistently.

Everything else still looks good to me: the new flag is wired into the mutually-exclusive test group, the main() dispatch handles it correctly (a user-supplied --tests still works), and the glob is anchored to the testsuite dir with a clean empty-match error. LGTM.

lumachad:
Thanks! Merging.

PR #342, opened 2026-09-15
testsuite: rename ROCm gcore test files and rocm_generate_core_file

lumachad:
Only `gdb.rocm/runtime-core*.exp` deals with system corefiles generated
by the OS. The remaining corefile tests all use `gcore` to generate core
files. Rename them to reflect that:

- `corefile.{exp,cpp}` -> `gcore.{exp,cpp}`
- `core-no-read-special-files.{exp,cpp}` -> `gcore-no-read-special-files.{exp,cpp}`
- `load-core-remote-system.{exp,cpp}` -> `load-gcore-remote-system.{exp,cpp}`

Also rename `rocm_generate_core_file` to `rocm_generate_gcore_file` in
`lib/rocm.exp` and update all call sites.


lancesix (commented):
shouldn't this also rename the *cpp alonside the *.exp?

lumachad:
I expect this will run into corefile test issues due to the now-unknown names of the new tests. But we'll see.

lumachad:
> shouldn't this also rename the *cpp alonside the *.exp?

Ouch. Obviously.

lumachad:
Updated so we also rename the cpp files.

lumachad:
> I expect this will run into corefile test issues due to the now-unknown names of the new tests. But we'll see.

Now fixed from TheRock's side.

lancesix (approved):
LGTM, thanks.

PR #345, opened 2026-09-16
gdb: add MSVC C++ ABI support PART 1

czidev-amd:
## Motivation

GDB has a single C++ ABI implementation for Windows, gnu-v3, which implements the Itanium ABI used by GCC/MinGW. Binaries produced by MSVC, or by clang with --target=x86_64-pc-windows-msvc, follow a different ABI. Debugging them with the Itanium rules gives wrong answers for object layout, virtual dispatch and inferior calls.

This patch adds a "msvc" cp_abi_ops implementation covering the parts of the MSVC C++ ABI that GDB needs:

- name recognition (??0/??1/??_7/??H forms);
- MSVC x64 struct-passing rules (delegating triviality analysis to gnuv3_pass_by_reference, made non-static for this purpose);
- vbtable-based virtual base offsets;
-  and method pointer ops.

## Prerequisite

- The sret (return-value) fix for non-trivial C++ types  is handled by a separate patch already under review as ROCgdb PR #253. That patch must land before this PR because calls gnuv3_pass_by_reference to determine whether a type has a non-trivial copy constructor or destructor; the correctness of the pass-by-reference path for small structs depends on the sret path in amd64_windows_return_value being fixed for the same types.

- The MSVC symbol demangler ROCgdb  PR #92 . This series deliberately excludes the MSVC symbol demangler (libdemangle-msvc) and the C++ name-matching work that depends on it -- linespec fallbacks, demangled-name normalisation, and the dwarf2/c-lang name handling.

## Test Plan and Results

Full differential run on x86_64-w64-mingw32 (Windows 11, MSYS2 UCRT64), 868 .exp files per leg, gdb.base + gdb.cp, two legs compiled from identical sources (with PR #253 applied):

  oracle  gcc-mingw   (Itanium ABI, x86_64-w64-mingw32-g++)
  target  clang-msvc  (MSVC ABI, clang --target=x86_64-pc-windows-msvc)

| Leg                 | PASS     |FAIL   |Stalled  |Tcl aborts|
|--------------------|------------|---------|-----------|---------------|
| gcc-mingw oracle|   51,656|    1,456|         0 |          6|
|clang-msvc target|  45,480|    3,956|         0|           8|

```   
 gcc-PASS and clang-FAIL:  2,495  (101 files)
```

For comparison, the same measurement on unpatched amd-staging gives a differential of 2,611. The series reduces it by roughly 116 entries on its own.

Oracle leg, compared by unique failing test name against unpatched amd-staging:
| | |
|-------------------------------------|----------------------------|
| unpatched amd-staging |  1,463 unique FAIL|
| this series             | 1,456 unique FAIL|
|                       | net -7  (3 new, 10 resolved)

Four of the ten resolved are this series working (sret fix from ROCgdb PR #253 also contributes to the non-trivial-retval.exp results):

    gdb.cp/non-trivial-retval.exp: finish from f1
    gdb.cp/non-trivial-retval.exp: finish from f2
    gdb.cp/non-trivial-retval.exp: finish from f22
    gdb.cp/non-trivial-retval.exp: finish from f5

The other six are environment noise (prelink needs /usr/sbin/prelink, argv0-symlink, one gcore-nonstop timeout).

The three new entries are not regressions:
- gdb.base/solib-display.exp is a string artifact. The test name embeds the build directory, so one identical failure appears as one removed and one added:
        ```
        unpatched:  can't rebuild .../build/rocgdb/.../solib-display-lib.so
        series:     can't rebuild .../build/abi-series/.../solib-display-lib.so
        ```
- two gdb.base/commands.exp "enter commands (timeout)" entries are pre-existing expect/pty flakiness. Run serially, without -j, the failure set varies between runs of the same binary (5, 5, then 3 failures), and an unrelated build containing none of these four commits fails identically. commands.exp exercises user-defined CLI  commands and touches no C++ ABI code.

## Review Notes
- `msvc_pass_by_reference`: the MSVC x64 rule is that a struct or union is passed by hidden pointer unless its size is exactly 1, 2, 4 or 8 bytes, regardless of triviality. Triviality analysis itself is delegated to gnuv3_pass_by_reference, which already reads DW_AT_defaulted / DW_AT_deleted correctly; the MSVC size rule is applied on top of its result.
-  `msvc_abi_new_objfile`: The location of this function may not be well suited. Also, it is using a heuristic to select the proper ABI.
- `msvc_baseclass_offset` / `msvc_match_vbase_expr`: the vbtable location expression shape was derived from clang-generated DWARF. MSVC-generated DWARF has not been tested and may differ.


gdb/msvc-abi.c
```diff
@@ -0,0 +1,394 @@
...
+    return complete_object_ctor;
+  return (enum ctor_kinds) 0;
+}
+
+/* MSVC destructor mangled names begin with "??1".  */
+
+static enum dtor_kinds
+msvc_is_destructor_name (const char *name)
+{
+  if (name[0] == '?' && name[1] == '?' && name[2] == '1')
```
vuzelac-amd, line 52: Please extend with these: 

??$?0 - Constructor-template instantiation
??_G - Scalar deleting destructor
??_E - Vector deleting destructor
??_D - Virtual-base destructor helper

Also note that ideally, we could call the same functions GNU v3 calls - `is_gnu_v3_mangled_ctor/dtor`. But that would make the PR dependent on libdemangle-msvc PR and also, libdemangle-msvc doesn't have support for `is_gnu_v3_mangled_ctor/dtor` yet. Probably an overkill as testing for ctor/dtor on MSVC is easy.

lumachad:

## Review summary

Solid implementation of a well-scoped MSVC ABI module. The core logic (vbtable pattern-matching, pass-by-reference override, method pointer layout) is correct and clearly documented. A few style and robustness issues need attention before landing.

## Cosmetic / style

- `gdb/Makefile.in:1121` -- `msvc-abi.c` is inserted between `gnu-v3-abi.c` and `go-lang.c`, but alphabetically `m` > `g`, so it belongs after `go-valprint.c \`.
- `gdb/msvc-abi.c:1-23` -- Two back-to-back comment blocks both describe the same purpose. The FSF license header (lines 1-19) is followed immediately by a second descriptive block (lines 21-23) that largely repeats it. Fold the detail into the license header opening line as `gnu-v3-abi.c` does, or drop the second block.
- `gdb/msvc-abi.c` (`init_msvc_abi_ops`) -- Assignments use alignment padding (`msvc_abi_ops.longname  = ...`, `msvc_abi_ops.doc       = ...`) but `gnu-v3-abi.c` uses a single space after `=` throughout with no padding. Remove the extra spaces to match the local convention.

## GNU coding standards

- `gdb/msvc-abi.c:100-103` -- Single-statement `if` body wrapped in braces. GDB omits braces for single-statement bodies without an `else`. Should be `if (...) info.trivially_copyable = false;`.
- `gdb/cp-abi.h:258-260` -- The comment "Both are made non-static in gnu-v3-abi.c specifically for this purpose." is an implementation note about `gnu-v3-abi.c` internals. It belongs next to the now-non-static definitions in that file (explaining why `static` was removed), not in the header. The header comment should describe what the declarations are.

## Typos / comments

- `gdb/msvc-abi.c:65-68` and `gdb/msvc-abi.c:138-141` -- The vbtable "minus direction is a 32-bit artifact, the real offset is vbptr + N" explanation appears twice: once in the comment above `msvc_match_vbase_expr` and again inside `msvc_baseclass_offset`. One occurrence can be trimmed to a back-reference.

## Bugs

- `gdb/msvc-abi.c:198-205` -- The fallback for a missing DWARF location on a virtual base uses `bitpos == 0 && TYPE_N_BASECLASSES (type) > 0` as a sentinel for "no info available" and errors. This is fragile: a legitimate virtual base at offset 0 (e.g., the sole direct virtual base) would trigger the error incorrectly. Consider erroring unconditionally whenever the DWARF block is absent for a virtual base, rather than guessing from bitpos.
- `gdb/msvc-abi.c:73-74` -- `msvc_is_operator_name` checks `name[2]` and `name[3]` after confirming only that `name[0] == '?'` and `name[1] == '?'`. A two-character input `"??"` has `name[2] == '\0'`, which passes all three exclusion tests and causes the function to return 1 -- a false positive. Compare `gnuv3_is_vtable_name` which uses `startswith` and is safe by construction. Add a length guard or use the same approach.

## Tests

- No new `.exp` files are added. The differential run described in the PR is thorough but was done against an external board configuration that is not included. For upstreaming, documenting the board file or reproduction steps for the `clang-msvc` leg would allow reviewers to verify the results independently.

## Nits / optional

- `gdb/msvc-abi.c` (`msvc_print_method_ptr`) -- For virtual targets, `gnuv3_print_method_ptr` resolves and demangles the method name. The MSVC version prints the raw byte offset instead. A brief comment noting this is intentionally deferred pending the MSVC demangler (ROCgdb PR #92) would help future readers understand it is incomplete by design.
- `gdb/msvc-abi.c` (`msvc_method_ptr_to_value`) -- `struct gdbarch *gdbarch;` is declared then assigned several lines later. Collapse to `struct gdbarch *gdbarch = self_type->arch ();` at point of first use.

lumachad:
Should this be aimed at upstream? I don't think our CI is geared towards validating non-gdb.rocm and non-gdb.dwarf2 tests. Would we, at a minimum, benefit from the build test on Linux? Otherwise I'd mark this ci:skip.

gdb/msvc-abi.c
```diff
@@ -0,0 +1,394 @@
...
+  else if (has_itanium_sym)
+    set_cp_abi_as_auto_default ("gnu-v3");
+  /* else: no C++ symbols -- leave ABI unchanged.  */
+}
+
+INIT_GDB_FILE (msvc_abi)
+{
+  init_msvc_abi_ops ();
+  register_cp_abi (&msvc_abi_ops);
+  gdb::observers::new_objfile.attach (msvc_abi_new_objfile, "msvc-abi");
```
vuzelac-amd, line 393: Please check if this ABI selection runs after symbols are expanded, looks to me it's after the expansion in case --readnow is used I think the testsuite must run w/ and w/o readnow before confirming a patch.

gdb/msvc-abi.c
```diff
@@ -0,0 +1,394 @@
...
+    error (_("MSVC virtual base offset not available "
+	     "(no DWARF location expression for base class %d)"), index);
+
+  return bitpos / 8;
+}
+
+/* MSVC x64 single-inheritance method pointer: one pointer-sized slot
+   holding either the function address (non-virtual) or the vtable byte
+   offset with lsb set (virtual).  Multi-inheritance variants (12 or
+   16 bytes) are not implemented.  */
```
vuzelac-amd, line 223: Is it ok to silently ignore ? Is it maybe necessary to at least recognize single inheritance, and for the rest throw an error ?

czidev-amd:
> Should this be aimed at upstream? I don't think our CI is geared towards validating non-gdb.rocm and non-gdb.dwarf2 tests. Would we, at a minimum, benefit from the build test on Linux? Otherwise I'd mark this ci:skip.

I marked it as a `draft PR`. We need to land hole bunch of other patches until this PR will come to its turn. Most probably after we integrated the linux build in ROC.

lumachad:
> > Should this be aimed at upstream? I don't think our CI is geared towards validating non-gdb.rocm and non-gdb.dwarf2 tests. Would we, at a minimum, benefit from the build test on Linux? Otherwise I'd mark this ci:skip.
> 
> I marked it as a `draft PR`. We need to land hole bunch of other patches until this PR will come to its turn. Most probably after we integrated the linux build in ROC.

So I don't think we need CI yet. Marking ci:skip then.

gdb/msvc-abi.c
```diff
@@ -0,0 +1,394 @@
...
+#include "osabi.h"
+#include "valprint.h"
+#include "gdbcore.h"
+
+/* MSVC constructor mangled names begin with "??0".  */
+
+static enum ctor_kinds
+msvc_is_constructor_name (const char *name)
+{
+  if (name[0] == '?' && name[1] == '?' && name[2] == '0')
```
aktemur, line 42: We may want to check the length of str before reading the characters. Or maybe just use `startswith`.

gdb/msvc-abi.c
```diff
@@ -0,0 +1,394 @@
...
+
+/* Return pass-by-reference info for TYPE under MSVC x64 rules.
+
+   The MSVC x64 ABI passes structs whose size is not 1, 2, 4, or 8 bytes
+   via a hidden pointer regardless of triviality.  For register-sized
+   structs, triviality is determined by gnuv3_pass_by_reference.  */
+
+/* gnuv3_pass_by_reference and gnuv3_get_virtual_fn are declared in
+   cp-abi.h; made non-static in gnu-v3-abi.c for reuse here.  */
+
```
aktemur, line 86: This comment won't mean much once the patch goes in. It can be removed.

gdb/msvc-abi.c
```diff
@@ -0,0 +1,394 @@
...
+     non-register-sized structs: MSVC requires a hidden pointer for any
+     struct whose size is not 1, 2, 4, or 8 bytes.  trivially_copyable
+     is cleared to trigger the sret path; trivially_copy_constructible
+     is left alone so infcall uses bitwise copy, not a copy constructor.  */
+  struct language_pass_by_ref_info info = gnuv3_pass_by_reference (type);
+  ULONGEST len = type->length ();
+  if (len != 1 && len != 2 && len != 4 && len != 8)
+    {
+      info.trivially_copyable = false;
+    }
```
aktemur, line 109: No braces here.

gdb/msvc-abi.c
```diff
@@ -0,0 +1,394 @@
...
+
+  return info;
+}
+
+/* Return true if BATON matches the canonical clang-MSVC vbase DWARF
+   expression (dup/deref/constu N/minus/deref/plus) and set *SLOT_OFFSET
+   to N.  The expression encodes a vbtable lookup; the minus direction is
+   a 32-bit artifact, the real offset is vbptr + N.  */
```
aktemur: How likely is it that the DWARF expression will remain the same? A compiler change could break this code easily. I don't have a suggestion but am asking if you think there are alternatives.

gdb/msvc-abi.c
```diff
@@ -0,0 +1,394 @@
...
+  if (ptr_value == 0)
+    {
+      gdb_printf (stream, "NULL");
+      return;
+    }
+
+  int vbit = ptr_value & 1;
+  CORE_ADDR func_addr = ptr_value & ~(CORE_ADDR) 1;
+
+  if (vbit)
```
aktemur, line 269: Nit: please compare against 0 explicitly.

gdb/msvc-abi.c
czidev-amd: I am wondering if we should add all the variants here, or make a function in msvc demangler which will return the right answer for each mangled constructor/destructor like it is done for gnu_v3_abi variant.

gdb/msvc-abi.c
czidev-amd: I was considering using `startswith`, but there are a bit more than this. Please check my question/thought https://github.com/ROCm/ROCgdb/pull/345#discussion_r4027389388

PR #351, opened 2026-09-18
gdb.rocm: fall back to concrete arch when no generic target exists

lumachad:
## Summary
- Under the rocm-generic board, `generic_target_for` returns `""` for architectures with no compatible generic AMDGPU target (e.g. gfx90a, which falls between the gfx9-generic and gfx9-4-generic buckets).
- `hcc_amdgpu_targets` silently dropped such architectures, so no `--offload-arch` flag was passed at all.
- Now falls back to using the concrete arch as the offload target and emits a warning so it's clear the generic target isn't actually being exercised for that architecture.

## Test plan
- [x] Added unit tests in `gdb.rocm/hcc-amdgpu-targets.exp` covering mapped, unmapped (gfx90a), and mixed arch lists under `rocm_use_generic_targets`.
- [x] Run `gdb.rocm/hcc-amdgpu-targets.exp` locally.


gdb/testsuite/lib/rocm.exp
```diff
@@ -197,13 +197,21 @@ proc hcc_amdgpu_targets {} {
     }
 
     # When the rocm-generic board is active, map each concrete arch to
-    # its generic equivalent and drop any that have no mapping.
+    # its generic equivalent.  If a concrete arch has no generic
+    # equivalent, fall back to using the concrete arch itself, so we
```
lancesix: I find "equivalent" to be the wrong word, the concrete and generic arch are not equivalent.

I would probably go with "has no compatible generic target", or something like that.

lumachad: Makes sense. Let me patch that up.

lumachad: Updated the wording to "no compatible generic target" and reworded the surrounding comment/warning to match.

lancesix (approved):
LGTM, thanks.

